PR #38, opened 2022-09-10
Testing by updating scheme json

AbdulMutaal:
Testing by updating scheme json. Adding description in the properties object of TextInput payload.

### DIFF
--- a/assets/schema/ensemble_schema.json
+++ b/assets/schema/ensemble_schema.json
@@ -532,6 +532,7 @@
         {
           "type": "object",
           "properties": {
+            "description": "It also works in arrays",
             "onKeyPress": { "$ref": "#/$defs/Action-payload" },
             "value": {
               "type": "string",


assets/schema/ensemble_schema.json
```diff
@@ -532,6 +532,7 @@
         {
           "type": "object",
           "properties": {
+            "description": "It also works in arrays",
```
vusters, line 535: so "properties" is the special keyword in JSON Schema. Any direct children inside it will be made a property, so your description will ends up be a property named "property". Probably not what you want to do.

If you want the description on the TExtInput, put it at the root before "allOf"

assets/schema/ensemble_schema.json
```diff
@@ -2059,4 +2060,4 @@
 
 
   }
-}
\ No newline at end of file
+}
```
vusters, line 2063: Screen Shot 2022-09-10 at 8 46 52 AM

IF you just want to test the schema locally, just reference a local schema file like this

PR #135, opened 2022-12-12
Dropdown icon

hamxa152:


hamxa152:
# 8 items with icon
      - Dropdown:
          label: Example 8
          hintText: Select something
          items:
            - {'iconname' : star , value : Red}
            - {'iconname' : star , value : Green}
            - {'iconname' : star , value : Blue}

kmahmood74:
@hamxa152 can you please fix your VCS so it doesn't automatically format the code. Now your code appears to have changes everywhere just because the formatting changed. Please fix that and checkin again so we only see the real changes.

lib/widget/dropdown.dart
```diff
@@ -197,6 +207,24 @@ class SelectOneState extends FormFieldWidgetState<SelectOne> {
...
       results = [];
       for (SelectOneItem item in items) {
+        item.iconname != null
+            ? results.add(
+          DropdownMenuItem(
+            child: Row(
+              children: [
+                iconframework.Icon(item.iconname),
+                const SizedBox(
+                  width: 10.0,
```
kmahmood74, line 217: size should not be hardcoded. It can default to 10 but user should be able to override in yaml. We talked about this. @hamxa152

vusters:
@hamxa152 can you delete this PR please

PR #137, opened 2022-12-14
Dropdown icon

hamxa152:


lib/widget/dropdown.dart
```diff
@@ -212,10 +244,11 @@ class SelectOneState extends FormFieldWidgetState<SelectOne> {
...
 class SelectOneItem {
   SelectOneItem({
-    required this.value,
-    this.label
+    required this.value, this.label, this.iconname, this.iconsize
   });
 
   final dynamic value;
   final String? label;
+  final String? iconname;
```
vusters, line 252: we should just use IconModel (icon name, library, color, size) as it's already defined. Please see Utils.getIcon() and example widget like tab_bar.

lib/widget/dropdown.dart
```diff
@@ -96,11 +96,21 @@ abstract class SelectOne extends StatefulWidget with Invokable, HasController<Se
       for (var element in values) {
         // must be of value/label pair. Maybe let user overrides later
         if (element is Map) {
-          if (element['value'] != null) {
-            entries.add(SelectOneItem(
+          if (element['iconname'] != null) {
```
vusters, line 99: no need for if/else here if we have an IconModel. Simply set it even if null.

lib/widget/dropdown.dart
```diff
@@ -199,9 +209,31 @@ class SelectOneState extends FormFieldWidgetState<SelectOne> {
     if (items != null) {
       results = [];
       for (SelectOneItem item in items) {
-        results.add(DropdownMenuItem(
-          value: item.value,
-          child: Text(Utils.optionalString(item.label) ?? item.value)));
+        item.iconname != null
```
vusters, line 212: if a row has icon and another doesn't, they won't align properly. Can we keep the spacing if at least one item has icon?

vusters:
can you also update ensemble_schema.json so icon attributes will show up for dropdown? See TabBar example.

lib/widget/dropdown.dart
hamxa152: okay got it

lib/widget/dropdown.dart
hamxa152: yes we can

lib/widget/dropdown.dart
hamxa152: okay thank you

vusters:
for the ensemble_schema.json change, you can create a new PR to address that after this.

hamxa152:
okay @vusters let me fix this changing than i can update json schema

kmahmood74:
@hamxa152 yes please fix the code as Vu suggested. Changing the json schema can be in a different checkin. Let's get this done and the camera first.

hamxa152:
@kmahmood74 i fixed icon and push the code @vusters  please review the code again

lib/widget/dropdown.dart
```diff
@@ -96,13 +98,12 @@ abstract class SelectOne extends StatefulWidget with Invokable, HasController<Se
       for (var element in values) {
         // must be of value/label pair. Maybe let user overrides later
         if (element is Map) {
-          if (element['iconname'] != null) {
+          if (element['icon'] != null) {
```
vusters, line 101: no need for If/Else nor checking if element['icon'] is there. You can simply replace it with `SelectOneItem(value, label, icon)`

lib/widget/dropdown.dart
```diff
@@ -209,15 +210,12 @@ class SelectOneState extends FormFieldWidgetState<SelectOne> {
     if (items != null) {
       results = [];
       for (SelectOneItem item in items) {
-        item.iconname != null
+        item.icon != null
             ? results.add(
```
vusters, line 214: So if first item has an icon and second item doesn't, the alignment will be off right? What I meant is:

First see if there is at least one item with icon specified, if so always add the gap for all items.

lib/widget/dropdown.dart
hamxa152: okay

lib/widget/dropdown.dart
hamxa152: okay

Issue #150, opened 2022-12-22
Complex calculation is not returning value


amin-nas:
```
View:
   Column:
     children:
       - Text:
           text: ${Math.floor(device.width / 2) - ((48 * 2 + 12)/2) }
```

kmahmood74:
fixed

PR #152, opened 2022-12-22
https://github.com/EnsembleUI/ensemble/issues/150 simplified the rege…

kmahmood74:
…x so we don't have to keep modifying it. Now it should catch everything within the curly braces


kmahmood74:
unit tests are failing with my change. Ignore the pull request for now, let me fix and re-checkin for review

PR #154, opened 2022-12-24
fixed space issue

hamxa152:


lib/widget/dropdown.dart
```diff
@@ -132,6 +126,39 @@ abstract class SelectOne extends StatefulWidget with Invokable, HasController<Se
 
   }
 
+  bool isIconExist(List v)
+  {
+    bool isIcon = false;
+    for (var e in v) {
+      if (e is Map) {
+        if (e['icon'] != null) {
+          isIcon = true;
```
vusters, line 135: you'd want to return immediately to avoid more looping.

lib/widget/dropdown.dart
```diff
@@ -132,6 +126,39 @@ abstract class SelectOne extends StatefulWidget with Invokable, HasController<Se
...
+  }
+
+  void calculateIconSize(List v) {
+    List iconSize = [];
+    _controller.size = 0;
+    for (var e in v) {
+      if (e is Map) {
+        if (e['icon'] != null) {
+          if (e['icon']['size'] != null) {
+            iconSize.add(e['icon']['size']);
```
vusters, line 149: nothing big here but why not just do Math.max(currentSize, newSize) so you don't mess with adding/removing array items?

lib/widget/dropdown.dart
```diff
@@ -210,14 +238,25 @@ class SelectOneState extends FormFieldWidgetState<SelectOne> {
     if (items != null) {
       results = [];
       for (SelectOneItem item in items) {
-        item.icon != null
-            ? results.add(
+        item.isIcon != null || item.isIcon == true
```
vusters, line 241: "item.isIcon == true" requires no need for null check

lib/widget/dropdown.dart
```diff
@@ -210,14 +238,25 @@ class SelectOneState extends FormFieldWidgetState<SelectOne> {
...
               children: [
-                iconframework.Icon.fromModel(item.icon!),
-                const SizedBox(
-                  width: 10.0,
+                item.icon != null
+                    ? iconframework.Icon.fromModel(item.icon!)
+                    : SizedBox(
+                  width: widget._controller.size.toDouble(),
+                ),
+                SizedBox(
```
vusters, line 252: This logic is long and hard to understand. Can you assign it to a variable called 'gap' and add the variable instead?

lib/widget/dropdown.dart
```diff
@@ -210,14 +238,25 @@ class SelectOneState extends FormFieldWidgetState<SelectOne> {
...
+                  width: item.icon != null
+                      ? item.icon!.size != null
+                      ? item.icon!.size == widget._controller.size
+                      ? 10
+                      : checkDifference(item.icon!.size!)
+                      : widget._controller.size.toDouble() + 10.0
```
vusters, line 258: Is this correct? The icon still have a size even if the size is not set right?

hamxa152: yes this logic is correct

lib/widget/dropdown.dart
hamxa152: okay

lib/widget/dropdown.dart
hamxa152: basically 1st is check icon exist or not in all array than is exist than is check how many time icon size are exist in array that 's why i added all icon sizes in are and find max value

lib/widget/dropdown.dart
hamxa152: done

lib/widget/dropdown.dart
hamxa152: I can added sizedbox here and sized box width is 10 because is icon added than add space between icon and text that's why i added 10.0 space between us

lib/widget/dropdown.dart
```diff
@@ -141,7 +141,7 @@ abstract class SelectOne extends StatefulWidget with Invokable, HasController<Se
 
   void calculateIconSize(List v) {
     List iconSize = [];
-    _controller.size = 0;
```
vusters: This is not what I meant. This variable is the 'iconWidth' so it should be named similar to that.

lib/widget/dropdown.dart
vusters: I meant the 2nd Sizebox (starting line 252) is the gap.

lib/widget/dropdown.dart
```diff
@@ -247,15 +247,15 @@ class SelectOneState extends FormFieldWidgetState<SelectOne> {
                 item.icon != null
                     ? iconframework.Icon.fromModel(item.icon!)
                     : SizedBox(
-                  width: widget._controller.size.toDouble(),
+                  width: widget._controller.gap.toDouble(),
```
vusters, line 250: please see above (this is not the gap)

lib/widget/dropdown.dart
```diff
@@ -247,15 +247,15 @@ class SelectOneState extends FormFieldWidgetState<SelectOne> {
...
                 ),
                 SizedBox(
                   width: item.icon != null
                       ? item.icon!.size != null
-                      ? item.icon!.size == widget._controller.size
+                      ? item.icon!.size == widget._controller.gap
                       ? 10
                       : checkDifference(item.icon!.size!)
-                      : widget._controller.size.toDouble() + 10.0
+                      : widget._controller.gap.toDouble() + 10.0
```
vusters, line 258: this logic runs when the icon size is not specified, so shouldn't it just be 10 pixels only? since the 1st SizeBox above already account for the empty icon space.

lib/widget/dropdown.dart
hamxa152: I used that sizedbox because icon is existing in list but there's no icon in that object so it can give gap between them and the next size box is giving space between icon and text that’s why it’s necessary to use both to keep it aligned

lib/widget/dropdown.dart
hamxa152: @vusters Basically I’m using first SizedBox with that icon because if icon exist in that list then give it specific gap and the second SizedBox is for if icon is available in list and it’s size is available or not available then it can give gap between text and icon

lib/widget/dropdown.dart
hamxa152: List created for icon size is for calculating maximum size and if i do this directly then function will render

PR #159, opened 2023-01-10
autocomplete and dropdown updated code

Moeez1999:


lib/widget/form_helper.dart
```diff
@@ -13,17 +12,30 @@ import 'package:flutter/material.dart';
...
+  double? enabledBoderRadius;
+  TextStyle? hintStyle = EnsembleTheme.body2;
+  TextStyle? labelStyle = EnsembleTheme.body2;
+  double? focusedBoderRadius;
 
   @override
   Map<String, Function> getBaseGetters() {
     Map<String, Function> getters = super.getBaseGetters();
     getters.addAll({
+      'autoComplete': () => autoComplete,
```
vusters, line 38: This is the wrong place. AutoComplete only make sense on dropdown. Why put it into the FormFieldController, which is the base controller for all form widgets?

lib/widget/dropdown.dart
```diff
@@ -182,70 +212,194 @@ class SelectOneState extends FormFieldWidgetState<SelectOne> {
     }
   }
 
-
-
   @override
   Widget buildWidget(BuildContext context) {
-    if (widget.getType() == SelectOneType.dropdown) {
+    if (widget._controller.autoComplete == false &&
+        widget.getType() == SelectOneType.dropdown) {
```
vusters, line 218: need consistency checking if the widget is of type SelectOneType.dropdown. You should put the if check, then inside branch whether it's an AutoComplete or simle Dropdown.

Either that or just remove the SelectOneType check. Whoever implements another SelectOne widget will have to refactor this code anyway

vusters (commented):
Please squash the commits and re-upload "all" files without reformatting.

Issue #176, opened 2023-01-18
expose navigation events from webview


kmahmood74:
related to https://github.com/EnsembleUI/ensemble/discussions/166

kmahmood74:
checked in

kmahmood74:
@amin-nas can you please do a build for studio. This only works on iOS and Android so we need to update the Ensemble Preview app.

PR #180, opened 2023-01-23
https://github.com/EnsembleUI/ensemble/issues/176 nav events for webview

kmahmood74:
Added - 
EnsembleEvent object
All cases where a widget has an action as an eventHandler, I pass an event with it. 

Code example - 

      - WebView:
          id: webview
          uri: https://ensembleui.com/
          styles:
            height: 400
          onPageStart:
            action: executeCode
            body: |
              //@code
              ensemble.debug('onPageStart:'+event.data.url);

          onPageFinished:
            action: executeCode
            body: |
              //@code
              ensemble.debug('onPageFinished:'+event.data.url);
          onNavigationRequest:
            action: executeCode
            body: | 
              //@code
              if ( event.data.url.includes('studio.ensembleui.com') ) {
                event.allowNavigation = false;
                ensemble.debug('prevented navigation');
              }
              ensemble.debug('onNavigationRequest:'+event.data.url);

### DIFF
--- a/lib/screen_controller.dart
+++ b/lib/screen_controller.dart
@@ -8,6 +8,7 @@ import 'package:ensemble/framework/device.dart';
 import 'package:ensemble/framework/error_handling.dart';
 import 'package:ensemble/framework/action.dart';
 import 'package:ensemble/framework/data_context.dart';
+import 'package:ensemble/framework/event.dart';
 import 'package:ensemble/framework/scope.dart';
 import 'package:ensemble/framework/widget/screen.dart';
 import 'package:ensemble/framework/widget/toast.dart';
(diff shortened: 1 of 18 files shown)


lib/screen_controller.dart
```diff
@@ -88,6 +89,9 @@ class ScreenController {
     if (action.initiator != null) {
       dataContext.addInvokableContext('this', action.initiator!);
     }
+    if ( event != null ) {
+      dataContext.addInvokableContext('event', event);
```
vusters, line 93: can you remove null check and always do `.addInvokableContext('event', event ?? {})` instead? It's easier to not having to check for null event in javascript e.g. if (event != null && event.source !=null )

lib/widget/button.dart
```diff
@@ -168,7 +170,7 @@ class ButtonState extends WidgetState<Button> {
 
     // execute the onTap action
     if (widget._controller.onTap != null) {
-      ScreenController().executeAction(context, widget._controller.onTap!);
+      ScreenController().executeAction(context, widget._controller.onTap!,event: EnsembleEvent(widget));
```
vusters, line 173: not a fan of having to explicitly set the event - I'd forget. I guess we can refactor later to dispatch the Action as an action instead of calling a Singleton. Widget will then automatically inject itself in

vusters (approved):

lib/screen_controller.dart
kmahmood74: But that can actually cause a bunch of issues in javascript. If as an app developer, you think that event is not null and start using its properties, you will get errors and won't know what to do. event.source is a required property so when event is not-null, it has to be there and not-null as well. By making the event an empty object, we are making life harder and causing more issues for the app developer because now they won't just have to check for event != null but every single property of the event - event.source != null, event.data != null.

It's a bad practice and highly error prone to put an empty object there when there actually should be null.

lib/widget/button.dart
kmahmood74: @vusters note that event is NOT required. It must not be required because an action can be triggered without an event as well. You should be able to trigger an action from the code or trigger an event from the code. Event should be there only when the action is being triggered as a result of an event. So we can't just always add it.

lib/screen_controller.dart
kmahmood74: @vusters please see my comment above.

vusters: executeAction can be called from outside a widget, in which case event.source would be null, but you can still pass the data. e.g. page's onLoad.

kmahmood74: no, the event.source would the page in that case. Event always MUST have a source.

vusters: why enforcing that event always need a source? a Timer can repetitively executeActions and pass new event data each time, but source is irrelevant. Why force the user to pass a source, especially when they don't need or can't do anything to it?

PR #187, opened 2023-01-26
Camera updated

hamxa152:


lib/screen_controller.dart
```diff
@@ -147,7 +147,11 @@ class ScreenController {
         final res = await Navigator.push(
           context,
           MaterialPageRoute(builder: (context) => CameraScreen(
-            initialCamera: action.options!['initialCamera'] ?? 'DEFAULT',
+            mode: action.options!['mode'] != null ? Utils.getString(action.options!['mode'], fallback: '') : 'DEFAULT',
```
vusters, line 150: Please never force !. This will throw an exception if options are not included. Options are comletely optional

lib/screen_controller.dart
```diff
@@ -147,11 +147,11 @@ class ScreenController {
         final res = await Navigator.push(
           context,
           MaterialPageRoute(builder: (context) => CameraScreen(
-            mode: action.options!['mode'] != null ? Utils.getString(action.options!['mode'], fallback: '') : 'DEFAULT',
-            initialCamera: action.options!['initialCamera'] != null ? Utils.getString(action.options!['initialCamera'], fallback: '') : 'DEFAULT',
-            useGallery: action.options!['useGallery'] != null ? Utils.getBool(action.options!['useGallery'], fallback: false) : true,
-            preview: action.options!['preview'] != null ? Utils.getBool(action.options!['preview'], fallback: false) : false,
-            maxCount: action.options!['maxCount'] != null ? Utils.getInt(action.options!['maxCount'], fallback: 0) : 1,
+            mode: action.options == null? 'DEFAULT' : action.options!['mode'] != null ? Utils.getString(action.options!['mode'], fallback: '') : 'DEFAULT',
```
vusters, line 150: Don't use 'Default' as it is non-descriptive (doesn't tell you anything) and is another state to keep track of. Either use null (I prefer this for simple attribute) or a name (video_and_photo).

This applies for all the attributes also. Don't use the name `default`

lib/screen_controller.dart
hamxa152: i fixed

lib/framework/widget/camera_manager.dart
```diff
@@ -56,24 +22,37 @@ class CameraManager {
...
     }
+    // if camera action is not null 
     else {
-      print('Check Else Condition');
+    // camera instance is created for set properties to setter
+    // if any other camera instance is created already automatically remove beacuse camera instance is dispose
+    // when screen is close
+     
       CameraScreen camera = CameraScreen();
       cameraAction.options!['mode'] == null
```
vusters, line 32: This is wrong as you expect the user to enters the correct mode (they can enter mode=whatever and you will pick up the invalid value). Better to use an enum, something liek this:
if (options['mode'] == 'photo') 
  mode = CameraMode.PHOTO;
else if (options['model] == 'video')
   mode = CameraMode.VIDEO  
else
   mode = CameraMode.BOTH.
   
This way you always get a valid value, and default to BOTH (that's the spec) if they enter incorrect value

lib/framework/widget/camera_manager.dart
```diff
@@ -56,24 +22,37 @@ class CameraManager {
...
-          ? camera.setProperty('mode', 'back')
+          ? camera.setProperty('mode', 'photo')
           : camera.setProperty('mode', cameraAction.options!['mode']);
-      // cameraAction.options!['mode'] != null ? camera.setProperty('mode', Utils.getString(cameraAction.options!['mode'], fallback: '')) : camera.setProperty('mode', 'photo');
-      // cameraAction.options!['initialCamera'] != null ? camera.setProperty('initialCamera', Utils.getString(cameraAction.options!['initialCamera'], fallback: '')) : camera.setProperty('initialCamera', 'back');
-      // cameraAction.options!['useGallery'] != null ? camera.setProperty('useGallery', Utils.getBool(cameraAction.options!['useGallery'], fallback: false)) : camera.setProperty('useGallery', true);
-      // cameraAction.options!['preview'] != null ? camera.setProperty('preview', Utils.getBool(cameraAction.options!['preview'], fallback: false)) : camera.setProperty('preview', false);
-      // cameraAction.options!['maxCount'] != null ? camera.setProperty('maxCount', Utils.getInt(cameraAction.options!['maxCount'], fallback: 0)) : camera.setProperty('maxCount', 1);
+      cameraAction.options!['initialCamera'] == null
```
vusters, line 35: Same for the rest of the attributes

lib/widget/camera.dart
```diff
@@ -1,34 +1,75 @@
...
 }
 
-class _CameraScreenState extends State<CameraScreen> {
-  List<CameraDescription> cameras = [];
+class MyCameraController extends WidgetController{
 
   CameraController? cameracontroller;
+
+  String mode = 'photo';
+  String initialCamera = 'photo';
```
vusters, line 54: wrong value

lib/widget/camera.dart
```diff
@@ -1,34 +1,75 @@
...
+  }
 }
 
-class _CameraScreenState extends State<CameraScreen> {
-  List<CameraDescription> cameras = [];
+class MyCameraController extends WidgetController{
 
   CameraController? cameracontroller;
+
+  String mode = 'photo';
```
vusters, line 53: mode = both according to the specs

lib/widget/camera.dart
```diff
@@ -1,34 +1,75 @@
...
+  @override
+  Map<String, Function> methods() {
+    return {};
+  }
 
   @override
-  State<CameraScreen> createState() => _CameraScreenState();
+  Map<String, Function> setters() {
+    return {
+      'mode':(value) => _controller.mode = Utils.optionalString(value) ?? _controller.mode,
```
vusters, line 40: don't do this. Make it explicit like optionalString(value) ?? CameraMode.BOTH

lib/widget/camera.dart
```diff
@@ -54,18 +96,33 @@ class _CameraScreenState extends State<CameraScreen> {
...
+    WidgetsBinding.instance.addObserver(this);
+    errorString = 'You just pick at least ${widget._controller.maxCount} image';
     initCamera().then((_) {
       ///initialize camera and choose the back camera as the initial camera in use.
       if (cameras.length == 2) {
-        if (widget.initialCamera == 'back') {
-          setCamera(0);
-        } else {
+        if (widget._controller.initialCamera == 'back') {
           setCamera(1);
```
vusters, line 94: Does the library not having an enum like FRONT / BACK? How do we know camera1 is the back and camera0 is the front?

lib/widget/camera.dart
```diff
@@ -236,7 +306,7 @@ class _CameraScreenState extends State<CameraScreen> {
           textbutton(
             title: 'Pick from gallery',
             onPressed: () {
```
vusters, line 295: Never show error when you don't have to. If useGallery is false, simply don't display the button.

lib/widget/camera.dart
hamxa152: There’s no enum defined in library. I have checked, whenever front camera opens at 0 index and back at 1 index, camera name’s got written in camera description but there’s no description about front or back camera

vusters (commented):
You use two variables (isFrontCamera and index) to track whether it's the front camera or back. Please use only 1. Even better use none if you can get it from the library which camera it is on. That way there is less logic to go wrong.

lib/widget/camera.dart
```diff
@@ -89,17 +95,24 @@ class CameraScreenState extends WidgetState<CameraScreen> with WidgetsBindingObs
...
-          setCamera(1);
+      if (cameras.length >= 2) {
+        if (widget._controller.initialCamera == InitialCamera.back) {
+          final back = cameras.firstWhere(
+              (camera) => camera.lensDirection == CameraLensDirection.back);
+          setCamera(cameraDescription: back);
           isFrontCamera = false;
-        } else {
-          setCamera(0);
+          setState(() {});
```
vusters, line 104: these lines are repeated three times in this block. Can you simplify it.

hamxa152: When initial camera gets back then we can find description of back camera through list and pass to set camera then that camera runs. Same goes for front camera and in case of no back camera and no front camera then default. If user doesn’t add initial camera option then it will go back by default. But this process will happen when camera length is 2 or greater than 2. Otherwise it will open default camera which will run isdefine that can be front or back

hamxa152:
@vusters I’m not using infront and index for front and back camera. I’m using index for photo and video. I’m using infront because when i shuffle camera then i can check which condition i want to run

lib/widget/camera.dart
hamxa152: i fixed according to your suggestion

lib/widget/camera.dart
hamxa152: yes i done my changes and using enum and remove 0 or 1

lib/widget/camera.dart
hamxa152: yes i change string to enum and define according to your requirements

vusters (approved):

PR #190, opened 2023-01-27
Listview added intially

Moeez1999:


lib/layout/box_layout.dart
```diff
@@ -154,7 +164,9 @@ class BoxLayoutState extends WidgetState<BoxLayout> with TemplatedWidgetState {
       children = updatedChildren;
     }
 
-    Widget boxWidget = _buildBoxWidget(children);
+    Widget boxWidget = templatedChildren != null && widget is ListView
+        ? _buildListView(children)
```
vusters, line 168: if else is not ideal as ListView is very different than BoxWidget. Can you extract the barebone (didChangeDependencies(), dispose()) into a base class say "WidgetListState", then have both BoxLayoutState and a new ListViewState extend it?

vusters: that way ListViewState is clean, and knows nothing about the Column/Row/...

lib/layout/box_layout.dart
```diff
@@ -200,6 +212,16 @@ class BoxLayoutState extends WidgetState<BoxLayout> with TemplatedWidgetState {
 
   }
 
+  Widget _buildListView(List<Widget> listViewChildren) {
+    return flutter.ListView.builder(
+        scrollDirection: Axis.vertical,
+        itemCount: listViewChildren.length,
+        shrinkWrap: true,
+        itemBuilder: (BuildContext context, int index) {
```
vusters, line 220: Can you run a test of say 100 widgets. All 100 widgets will be constructed and called their initState(), but only the visible one in the ListView will call their buildWidget() method?

vusters: I think it should work like that but if it doesn't we have to do proper paging.

lib/page_model.dart
```diff
@@ -165,7 +165,7 @@ class PageModel {
   WidgetModel buildRootModel(YamlMap viewMap, Map<String, dynamic>? customViewDefinitions) {
     WidgetModel? rootModel = getRootModel(viewMap, customViewDefinitions!);
     if (rootModel != null) {
-      if (![Column.type, Row.type, Flex.type, EnsembleStack.type, AppScroller.type].contains(rootModel.type)) {
+      if (![Column.type, Row.type, Flex.type, EnsembleStack.type, AppScroller.type, ListView.type].contains(rootModel.type)) {
```
vusters, line 168: remove this. ListView doesn't need to be the root widget. The rootwidget is meant as a wrapper to include other widgets.

lib/layout/list_view.dart
```diff
@@ -0,0 +1,147 @@
...
+  static const type = 'ListView';
+  ListView({Key? key}) : super(key: key);
+
+  @override
+  bool isVertical() {
+    return true;
+  }
+}
+
+abstract class BoxLayout extends StatefulWidget
```
vusters, line 21: I don't understand. Is this a duplicate of the original BoxLayout? Did you mean to remove the original one?

lib/layout/list_view.dart
```diff
@@ -8,22 +9,17 @@ import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart' as flutter;
 
 // ignore: must_be_immutable
-class ListView extends BoxLayout {
+class ListView extends ListViewWidget {
```
vusters, line 12: I didn't catch this before. Why have both ListView and ListViewWidget when they are not used by anything else? Please combine them

vusters (commented):
- Rename onTap to onItemTap. We want to know when each item is tapped, not the whole ListView.

vusters:
Go ahead and pull from main and merge your existing code. It is out of date. Once done, you can implement my changes suggested here

vusters:
Hey @Moeez1999 Please read over the specs carefully https://github.com/orgs/EnsembleUI/projects/2/views/1?pane=issue&itemId=15643685. You don't have most of those implemented. For clarity I removed the boxLayout styling and list them all out for you. You may merge the code then add all of them in.

lib/layout/list_view.dart
```diff
@@ -106,22 +105,43 @@ class ListViewWidgetState extends WidgetState<ListViewWidget> with TemplatedWidg
...
+            shrinkWrap: true,
+            itemBuilder: (BuildContext context, int index) {
+              return GestureDetector(
+                  onTap: widget._controller.onItemTap == null ? null : () =>
+                      ScreenController().executeAction(context, widget._controller.onItemTap!),
+                  child: children[index]);
+            }),
+      ),);
+  }
+  BoxDecoration _buildBoxDecoration() {
```
vusters, line 131: did you copy this from some other widgets? can you re-use it as function instead?

vusters (commented):
Looks good. What about backgroundColor and separator*?

vusters (approved):

kmahmood74:
@Moeez1999 every commit MUST HAVE THE URL TO THE TICKET as the first comment. This way github automatically links it.

PR #197, opened 2023-02-02
Essemble live error

Moeez1999:


lib/ensemble_app.dart
```diff
@@ -72,12 +72,38 @@ class EnsembleAppState extends State<EnsembleApp> {
...
+                      },
+                      child: const Text(
+                        "OK",
+                        style: TextStyle(
+                          color: Color(0xff059ACD),
+                        ),
+                      ))
+                ],
+                title: const Text(
+                  "Invalid QR code",
```
kmahmood74, line 92: ensemble is a library, it doesn't know anything about QR Codes. It gets an appid and tries to load it. You need to check the error and see if it is due to definition not found and then show a message. Also every text within Ensemble must be translatable and must not be hardcoded.

lib/ensemble_app.dart
```diff
@@ -72,12 +72,38 @@ class EnsembleAppState extends State<EnsembleApp> {
...
+                  style: TextStyle(
+                    color: Color(0xffFFFFFF),
+                    fontWeight: FontWeight.w600,
+                  ),
+                ),
+                content: const Text(
+                    "Make sure you are scanning a QR code displayed on Ensemble studio.",
```
kmahmood74, line 99: see comment above

vusters:
Please start putting description on what the review is about, plus link to the Github issue.

vusters (commented):
None of these logic belong here. This only makes sense on Ensemble Live as you are trying to go back to the root page. This will give you unexpected result 99% of the time while you are in the app.

lib/ensemble_app.dart
Moeez1999: Mostly when we try to scan invalid qr code then config will not be loaded and it gives us error loading configuration So i have made an invalid qr code error message there but when we write invalid app id it gives us invalid page defination error so i appear invalid app id error there but if you want i will change this invalid qr code to invalid app id error message
@kmahmood74

PR #225, opened 2023-02-15
next preview function added

hamxa152:


vusters:
carousel is a separate issue than the video. You need to put it in a different PR.

kmahmood74:
@hamxa152 your PR has video changes as well. Please do not mix changes from two different tickets into one PR. Each PR must only address one ticket. 

Also as I have requested a dozen times - please put the link to the ticket in the comments of your commit. I don't see any links in your commit and don't know which PR they match to

PR #226, opened 2023-02-15
listview ontap index fix done https://github.com/EnsembleUI/ensemble/…

Moeez1999:
…issues/220

### DIFF
--- a/lib/layout/list_view.dart
+++ b/lib/layout/list_view.dart
@@ -1,5 +1,6 @@
 
 
+import 'package:ensemble/framework/event.dart';
 import 'package:ensemble/framework/widget/widget.dart';
 import 'package:ensemble/layout/box_layout.dart';
 import 'package:ensemble/layout/layout_helper.dart';
@@ -28,7 +29,9 @@ class ListView extends StatefulWidget
 
   @override
   Map<String, Function> getters() {
-    return {};
+    return {
+      'selectedItemIndex': () => _controller.selectedItemIndex,
+    };
   }
 
   @override
@@ -137,10 +140,9 @@ class ListViewState extends WidgetState<ListView> with TemplatedWidgetState {
           shrinkWrap: true,
           itemBuilder: (BuildContext context, int index) {
             return GestureDetector(
-                onTap: widget._controller.onItemTap == null
+               onTap: widget._controller.onItemTap == null
                     ? null
-                    : () => ScreenController()
-                    .executeAction(context, widget._controller.onItemTap!),
+                    : () => _onItemTapped(index),
                 child: children[index]);
           }),
     );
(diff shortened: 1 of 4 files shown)


lib/layout/list_view.dart
```diff
@@ -158,6 +160,18 @@ class ListViewState extends WidgetState<ListView> with TemplatedWidgetState {
...
   }
 
+  // ----------------- To GET the current [index] of the item in data array -------------------
+
+  _onItemTapped(int index) {
+    if (index != widget._controller.selectedItemIndex &&
+        widget._controller.onItemTap != null) {
+      widget._controller.selectedItemIndex = index;
+      //log("Changed to index $index");
+      ScreenController().executeAction(context, widget._controller.onItemTap!,event: EnsembleEvent(widget));
```
kmahmood74, line 170: @Moeez1999 I am not sure if you are understanding the code. By just sending the widget as part of EnsembleEvent, you achieve nothing at all. This code change is of no use. You have to trace how events are used.. Then create your own event that extends EnsembleEvent and pass the selectedIndex and selectedItem as data on that event. Please fix this.

lib/widget/image.dart
```diff
@@ -85,19 +86,14 @@ class ImageState extends WidgetState<EnsembleImage> {
         // image binding is tricky. When the URL has not been resolved
         // the image will throw exception. We have to use a permanent placeholder
         // until the binding engages
-        return Image.network(
-            widget._controller.source,
-            width: widget._controller.width?.toDouble(),
+        return
```
kmahmood74, line 89: @Moeez1999 we need to use CachedNetworkImage if the cache property is set to true. Default is true. Please read the instructions. You are using the CachedImage always without checking the property

kmahmood74: also what about svgImage?

kmahmood74 (changes requested):

lib/widget/image.dart
Moeez1999: I have done the cache one for network images and for the SVG i am working on it when it will be completed i will push that

lib/widget/image.dart
```diff
@@ -86,27 +88,41 @@ class ImageState extends WidgetState<EnsembleImage> {
         // image binding is tricky. When the URL has not been resolved
         // the image will throw exception. We have to use a permanent placeholder
         // until the binding engages
-        return 
-        CachedNetworkImage(
-                  width: widget._controller.width?.toDouble(),
+        if (widget._controller.cache == true) {
```
kmahmood74, line 91: @Moeez1999 you can just do if (widget._controller.cache)
no need for == true

Moeez1999: sure we can do that

kmahmood74 (approved):

PR #234, opened 2023-02-16
Issue#200 js line numbers in errors

kmahmood74:


lib/screen_controller.dart
```diff
@@ -413,16 +413,6 @@ class ScreenController {
 
     // silently fail if error handle is not defined? or should we alert user?
   }
-
```
kmahmood74, line 446: I removed this as this was dead code

Issue #245, opened 2023-02-17
Strings within camera code must be translatable, settable in yaml and use proper english


kmahmood74:

kmahmood74:
closing as this was completed

PR #252, opened 2023-02-18
added properties

hamxa152:
**Ticket No** #245 
**Sample YAML configuration:**

      - Button:
         label: Open Camera
         onTap:
           action: openCamera
           options:
              mode: both
              preview: true
              maxCount: 3
              errormessage: testing
              imgPickerIcon: {name: star}
              initialCamera: front
              useGallery: true


lib/framework/widget/camera_manager.dart
```diff
@@ -15,32 +15,44 @@ class CameraManager {
...
-          'useGallery', cameraAction.options!['useGallery']);
+              'useGallery', cameraAction.options!['useGallery']);
       cameraAction.options!['preview'] == null
           ? camera.setProperty('preview', false)
           : camera.setProperty('preview', cameraAction.options!['preview']);
       cameraAction.options!['maxCount'] == null
           ? camera.setProperty('maxCount', 1)
-          : camera.setProperty('maxCount', cameraAction.options!['maxCount']);      
+          : camera.setProperty('maxCount', cameraAction.options!['maxCount']);
+      cameraAction.options!['errormessage'] == null
```
kmahmood74, line 42: @hamxa152 call it maxCountMessage so it is clear. It is not an error.

lib/framework/widget/camera_manager.dart
```diff
@@ -15,32 +15,44 @@ class CameraManager {
...
+          ? ''
+          : camera.setProperty(
+              'errormessage', cameraAction.options!['errormessage']);
+      cameraAction.options!['imgPickerIcon'] == null
```
kmahmood74, line 46: @hamxa152 we are using image instead of img everywhere. Please be consistent and call is imagePickerIcon

lib/widget/camera.dart
```diff
@@ -135,6 +158,23 @@ class CameraScreenState extends WidgetState<CameraScreen> with WidgetsBindingObs
     setState(() {});
   }
 
+  void defineErrorText() {
```
kmahmood74, line 161: @hamxa152 I thought we were going to let the user override this in YAML right?

kmahmood74: also this should be called getMaxSelectedMessage as it is not an error

lib/widget/camera.dart
```diff
@@ -135,6 +158,23 @@ class CameraScreenState extends WidgetState<CameraScreen> with WidgetsBindingObs
...
+    if (widget._controller.mode == CameraMode.photo) {
+      errorString =
+          'Maximum ${widget._controller.maxCount} images may be selected';
+      cameraoptionsList = ['PHOTO'];
```
kmahmood74, line 165: why are you setting cameraoptionsList in a function that is just constructing the error message? this function should not change any state right? it's just a getter for the message

lib/widget/camera.dart
```diff
@@ -135,6 +158,23 @@ class CameraScreenState extends WidgetState<CameraScreen> with WidgetsBindingObs
...
+    } else if (widget._controller.mode == CameraMode.video) {
+      errorString =
+          'Maximum ${widget._controller.maxCount} videos may be selected';
+      cameraoptionsList = ['VIDEO'];
```
kmahmood74, line 169: again, this should not be set in a function that's getting error message

lib/widget/camera.dart
```diff
@@ -135,6 +158,23 @@ class CameraScreenState extends WidgetState<CameraScreen> with WidgetsBindingObs
...
+    } else {
+      errorString =
+          'Maximum ${widget._controller.maxCount} images and videos may be selected';
+      cameraoptionsList = ['PHOTO', 'VIDEO'];
```
kmahmood74, line 173: same comment - don't set it here

lib/widget/camera.dart
```diff
@@ -135,6 +158,23 @@ class CameraScreenState extends WidgetState<CameraScreen> with WidgetsBindingObs
...
+    }
+    setState(() {});
```
kmahmood74, line 175: should not be calling setState in a function that is just constructing a message. fix your code please.

lib/widget/camera.dart
```diff
@@ -135,6 +158,23 @@ class CameraScreenState extends WidgetState<CameraScreen> with WidgetsBindingObs
...
+  }
+
   /// chooses the camera to use, where the front camera has index = 1, and the rear camera has index = 0
   void setCamera(
       {bool isNotDefine = false, CameraDescription? cameraDescription}) {
```
kmahmood74, line 180: proper english please. isNotDefine is incorrect English. Call it - isUndefined

lib/widget/camera.dart
```diff
@@ -38,32 +38,59 @@ class CameraScreen extends StatefulWidget with Invokable, HasController<MyCamera
...
-          InitialCamera.values.from(type) ?? InitialCamera.back,
+      'mode': (value) => _controller.initCameraMode(value),
+      'initialCamera': (value) => _controller.initCameraOption(value),
       'useGallery': (value) => _controller.useGallery =
           Utils.optionalBool(value) ?? _controller.useGallery,
       'maxCount': (value) => _controller.maxCount =
           Utils.optionalInt(value) ?? _controller.maxCount,
       'preview': (value) => _controller.preview =
           Utils.optionalBool(value) ?? _controller.preview,
+      'errormessage': (value) =>
```
kmahmood74, line 49: @hamxa152 allow for expressions such as 'Max count must not be ${this.maxCount}' as well in the messages. @vusters please comment on which framework method to call to handle that

kmahmood74 (changes requested):
see my comments

lib/widget/camera.dart
hamxa152: I’m setting camera option list value in this function because i called this function in initstate and it sets message as well as camera option list value will be equal to mode. I think we should change the function name. That’s your call. I’ll do accordingly

lib/widget/camera.dart
```diff
@@ -358,12 +358,7 @@ class CameraScreenState extends WidgetState<CameraScreen>
           textbutton(
             title: 'Pick from gallery',
```
kmahmood74, line 359: @hamxa152 please don't hardcode these strings. This needs to come from resource file. I can't go over all your code looking for hardcoded strings. Please remove all hardcoded strings and make sure they come from resource files.

kmahmood74 (changes requested):

lib/widget/camera.dart
```diff
@@ -158,7 +158,7 @@ class CameraScreenState extends WidgetState<CameraScreen>
     setState(() {});
   }
 
-  void defineErrorText() {
+  void getMaxSelectedMessage() {
```
kmahmood74, line 161: you should never change state in a "getxxxx' function. Please change the name of the function to configureOptions

lib/widget/camera.dart
```diff
@@ -48,9 +48,18 @@ class CameraScreen extends StatefulWidget
           Utils.optionalBool(value) ?? _controller.preview,
       'maxCountMessage': (value) =>
           _controller.maxCountMessage = Utils.getString(value, fallback: ''),
+      'permissionDeniedMessage': (value) => _controller
```
kmahmood74, line 51: @hamxa152 all fallback strings should be the default strings and not empty strings. Please do that for all fallback strings

lib/widget/camera.dart
```diff
@@ -346,17 +359,24 @@ class CameraScreenState extends WidgetState<CameraScreen>
         children: [
           imagePreviewButton(),
           const Spacer(),
-          const Text(
-              'To capture photos and videos, allow access to your camera.'),
+          Text(
+            widget._controller.permissionDeniedMessage ??
+                Utils.translateWithFallback('ensemble.input.maxCountMessage',
+                    'To capture photos and videos, allow access to your camera.'),
```
kmahmood74, line 365: put these messages as const static so you can use them as fallbacks for Utils.getString and also here

lib/widget/camera.dart
```diff
@@ -823,6 +851,13 @@ class CameraScreenState extends WidgetState<CameraScreen>
       }
     }
   }
+
+  String buttonLable(String label) {
```
kmahmood74, line 855: @hamxa152 please pay attention to your spellings. It should be buttonLabel

kmahmood74 (changes requested):
minor changes required. Please make the changes quickly so we may move forward

kmahmood74 (approved):

Issue #239, opened 2023-02-17
Some widgets should support Box styling


sharjeelyunus:
Text, Image, Lottie, QRCode, Carousel, Icon should support Box styling, which includes:

margin / padding
width / height
backgroundColor, backgroundImage, backgroundGradient
borderColor, borderWidth, borderRadius
shadowColor, shadowOffset, shadowRadius, shadowStyle

kmahmood74:
@amin-nas based on our discussion, are doing padding and margin both or just padding? Please update the description.

amin-nas:
Margin only applies to containers. We will add padding to all widgets.

vusters:
All widgets should support Box styling, which includes:
- margin / padding
- width / height
- backgroundColor, backgroundImage, backgroundGradient
- borderColor, borderWidth, borderRadius
- shadowColor, shadowOffset, shadowRadius, shadowStyle

PR #263, opened 2023-02-21
Text, Image, Lottie, QRCode, and Carousel now support box stylings

vusters:
#239 All the widgets above now support box styles, which include:
- margin / padding
- width / height (where applicable)
- backgroundColor, backgroundImage, backgroundGradient
- borderColor, borderWidth, borderRadius
- shadowColor, shadowOffset, shadowRadius, shadowStyle


vusters:
Some more important points:
- We should use ThemeManager going forward for all fallback stylings. The idea is to consolidate styles such that we have a reliable way to fallback to: `Ensemble theme` then `Material colorScheme` then `hardcoded value`
- Utility wrapper widget should be a StatelessWidget for rendering optimization. I only have BoxWrapper here but we should eventually move to this model.

Issue #260, opened 2023-02-20
backgroundColor for form elements


amin-nas:
* ability to set `backgroundColor` on form widgets such as `TextInput`
* ability to set `backgroundColor` of `Inputs` in the global theme

vusters:
![Screen Shot 2023-02-20 at 11 27 30 AM](https://user-images.githubusercontent.com/622269/220185560-ca2c851c-5db4-4787-b505-b60dc40bc66c.png)
In theme today we can define a variety of styles for Input widget (applicable to the entire App) - see screenshot. BackgroundColor is missing so please add that.

kmahmood74:
closing this as @hamxa152 is done

PR #267, opened 2023-02-21
https://github.com/EnsembleUI/ensemble/issues/260

hamxa152:
Ticket : https://github.com/EnsembleUI/ensemble/issues/260

Sample code:
            - TextInput:
                label: Input
                labelHint: Hello
                backgroundColor: green
                required: true

### DIFF
--- a/assets/schema/theme_schema.json
+++ b/assets/schema/theme_schema.json
@@ -125,6 +125,10 @@
         "borderDisabledColor": {
           "$ref": "#/$defs/Color",
           "description": "The disabled border color of applicable input fields"
+        },
+        "backgroundColor": {
+          "$ref": "#/$defs/Color",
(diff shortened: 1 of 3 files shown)


kmahmood74:
@hamxa152 you have to add this property to the theme as well. Please review the ticket

kmahmood74:
@hamxa152 you also have to add this to the theme schema as well otherwise the intelli-sense won't work - https://github.com/EnsembleUI/ensemble/blob/main/assets/schema/theme_schema.json

vusters:
@hamxa152 Good work. I merged this so the customer can use it, but you'll need to add:
1. backgroundColor to FormInput-payload in ensemble-schema. This is how people know that the property exists with intelli-sense. 
2. add "TextInput" theming (has all attribute of Input) so you can set the style once for all TextInputs.

Issue #258, opened 2023-02-20
Add maxFileSize property to uploadFileAction


kmahmood74:
I think we should have another property for uploadFiles action called maxFileSize. That would be important. 

maxFileSize - a double.
maxFileSizeMessage - message to show when user selects a file larger than the max allowed size. Default is below 

- size of a file in KB
- default is 100MB (100000 KB)
- for our preview sample, the default should be based on the server we are hitting

If user picks a file that is larger than the maxFileSize, show an error - 'The size of <filename> is <size> which is larger than the maximum allowed <maxFileSize>

This error message should be translatable. see Utils.translateWithFallback

kmahmood74:
cc @amin-nas

PR #275, opened 2023-02-22
File enhancements

snehmehta:
TIcket: #258 

Added two new parameters:
- maxFileSize 
- overMaxFileSizeMessage

Updated one parameter 
- UploadUrl => apiName

Users can expect the same functionality in `invokeAPI`, like setting url, method, headers, inputs.


vusters (approved):
Looks good.

Issue #231, opened 2023-02-16
ListView: add to the schema


kmahmood74:
Add ListView as a widget in Ensemble schema here with its properties - 
https://github.com/EnsembleUI/ensemble/blob/main/assets/schema/ensemble_schema.json

kmahmood74:
@Moeez1999 please don't forget this one. Feature is not complete unless it is available to be used in schema.

kmahmood74:
done

PR #285, opened 2023-02-24
listview schema added https://github.com/EnsembleUI/ensemble/issues/231

Moeez1999:


Moeez1999:
I have added the schema as i got the idea from the existing one so if there change tell me i will do that

assets/schema/ensemble_schema.json
```diff
@@ -956,6 +956,52 @@
...
+                }
+              ]
+            },
+            "selectedItemIndex": {
+              "type": "integer",
+              "description": "Selecting a ListView item gives the index of selected item"
+            }
+          }
+        },
+        { "$ref": "#/$defs/Templated-payload" }
```
vusters, line 1002: put this as the 1st child to be consistent with other rules, and easier to spot.

assets/schema/ensemble_schema.json
```diff
@@ -956,6 +956,52 @@
...
+                    },
+                    "separatorPadding": {
+                      "type": ["integer", "string"],
+                      "description": "Padding with CSS-style value e.g. padding: 5 20 5 Default 0 0 0"
+                    }
+                  }
+                }
+              ]
+            },
+            "selectedItemIndex": {
```
vusters, line 996: move this to before styles. The either is to list properties together, then styles afterward. It's a huge file and easier to debug/review.

assets/schema/ensemble_schema.json
```diff
@@ -1716,6 +1762,11 @@
             "Column": { "$ref": "#/$defs/Column-payload" }
           }
         },
+        {
+          "properties": {
+            "Column": { "$ref": "#/$defs/ListView-payload" }
```
vusters, line 1767: typo

Issue #219, opened 2023-02-13
TextArea widget


amin-nas:
Extend the current TextInput widget to allow multi-line

```yaml
TextInput:
  multiline: true | false //default is false. When true, this textinput allows user to type in multiple lines
  maxLines: null //its a number, default is null which means that user can keep typing and the textinput grows and wraps text
  minLines: 1 //default is 1. 
```

kmahmood74:
updated description. I looked at iOS/Android/Flutter, they all do TextInput/TextField with multiline property. HTML is old.

kmahmood74:
already works, we don't have minLines, rest is there. closing

PR #286, opened 2023-02-24
max line added

hamxa152:
Ticket : https://github.com/EnsembleUI/ensemble/issues/219
Ticket Url : [https://github.com/EnsembleUI/ensemble/issues/219](url)
Sample YAML configuration

- TextInput:
    label: Example 1
    hintText: Placeholder value
    required: true
    maxLines: 5

### DIFF
--- a/lib/widget/form_textfield.dart
+++ b/lib/widget/form_textfield.dart
@@ -298,6 +298,7 @@ class TextInputState extends FormFieldWidgetState<BaseTextInput> {
         return null;
       },
       textInputAction: widget._controller.keyboardAction,
+      keyboardType: widget.keyboardType,
       maxLines: widget._controller.maxLines,
       obscureText: isObscureOrPlainText(),
       enableSuggestions: !widget.isPassword(),
--- a/lib/widget/form_textfield.dart
+++ b/lib/widget/form_textfield.dart
@@ -157,8 +157,7 @@ class TextInputController extends FormFieldController {
   {
     if(data != null)
     {
-      maxLines = data;
-      notifyListeners();
+      maxLines =  Utils.optionalInt(data)!;
     }
   }
 }
--- a/lib/widget/form_textfield.dart
+++ b/lib/widget/form_textfield.dart
@@ -100,7 +100,8 @@ abstract class BaseTextInput extends StatefulWidget with Invokable, HasControlle
       'borderRadius': (value) => _controller.borderRadius = Utils.optionalInt(value),
       'validator': (value) => _controller.validator = Utils.getValidator(value),
       'obscureToggle': (value) => _controller.obscureToggle = Utils.optionalBool(value),
-      'keyboardAction': (value) => _controller.keyboardAction = _getKeyboardAction(value)
+      'keyboardAction': (value) => _controller.keyboardAction = _getKeyboardAction(value),
+      'maxLines' : (value) => _controller.updateMaxline(value),
     };
   }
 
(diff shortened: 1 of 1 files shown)


lib/widget/form_textfield.dart
```diff
@@ -287,7 +298,7 @@ class TextInputState extends FormFieldWidgetState<BaseTextInput> {
         return null;
       },
       textInputAction: widget._controller.keyboardAction,
-      keyboardType: widget.keyboardType,
```
vusters, line 290: why are you removing keyboardType? This is important to show the correct keyboard depending on the input.

lib/widget/form_textfield.dart
```diff
@@ -100,7 +100,8 @@ abstract class BaseTextInput extends StatefulWidget with Invokable, HasControlle
       'borderRadius': (value) => _controller.borderRadius = Utils.optionalInt(value),
       'validator': (value) => _controller.validator = Utils.getValidator(value),
       'obscureToggle': (value) => _controller.obscureToggle = Utils.optionalBool(value),
-      'keyboardAction': (value) => _controller.keyboardAction = _getKeyboardAction(value)
+      'keyboardAction': (value) => _controller.keyboardAction = _getKeyboardAction(value),
+      'maxLines' : (value) => _controller.updateMaxline(value),
```
vusters, line 104: 1. user may enter a string and this will throw an error. Use Utils.optionalInt(value)
2. You shouldn't need to do any of this nor call notifyListeners. Simply setting the maxLine will automatically call notifyListeners().

lib/widget/form_textfield.dart
hamxa152: oh sorry line got deleted by mistake

lib/widget/form_textfield.dart
hamxa152: fixed

lib/widget/form_textfield.dart
```diff
@@ -157,8 +157,7 @@ class TextInputController extends FormFieldController {
   {
```
vusters, line 157: why even have this function? like all the setters you can do 
'maxLines' : (value) => _controller.maxLines = Utils.getInt(value, min: 1, fallback: 1)
Then remove the default value of 1 when you declare maxLines.

Please test your changes though.

hamxa152: its not working its always null when maxLines is empty or null

lib/widget/form_textfield.dart
```diff
@@ -101,7 +101,7 @@ abstract class BaseTextInput extends StatefulWidget with Invokable, HasControlle
       'validator': (value) => _controller.validator = Utils.getValidator(value),
       'obscureToggle': (value) => _controller.obscureToggle = Utils.optionalBool(value),
       'keyboardAction': (value) => _controller.keyboardAction = _getKeyboardAction(value),
-      'maxLines' : (value) => _controller.updateMaxline(value),
+      'maxLines' : (value) =>  _controller.maxLines = Utils.getInt(value, min: 1, fallback: _controller.maxLines),
```
kmahmood74, line 104: @hamxa152 I don't understand. First you said that Utils.getInt(value, min: 1, fallback: _controller.maxLines) is not always working and then you just used it here?

hamxa152: when we set int? maxLines then its always set null value when maxlines is not define in ymal file 
but when  int maxLines = 1 then its work fine.

kmahmood74: so the problem was always with your controller.maxLines as it was an optional int with the default value of null?

lib/widget/form_textfield.dart
vusters: the line I have has a fallback, which forces maxLines to be 1 if the user doesn't enter anything. That will work yes?

vusters (approved):

Issue #238, opened 2023-02-17
Button styles: width and height


sharjeelyunus:
As of this moment, we can only style buttons with padding and margin, button should have width and height

kmahmood74:
already done - https://docs.ensembleui.com/#/widget-reference/button

PR #296, opened 2023-02-28
height/width added https://github.com/EnsembleUI/ensemble/issues/238

Moeez1999:


lib/widget/button.dart
```diff
@@ -95,9 +99,12 @@ class ButtonState extends WidgetState<Button> {
     }
 
     // add margin if specified
-    return widget._controller.margin != null ?
-      Padding(padding: widget._controller.margin!, child: rtn) :
-      rtn;
+    return Container(
```
vusters, line 102: Hmm, Button should configure its own width/height and not rely on the Container outside. Same goes with border and such. The only thing outside the button is probably the margin, and for that use a Margin widget.
If you want to change the width/height, see ThemeUtils.getButtonStyle(). Inside you you can set width/height, minWidth/minHeight, maxwidth/maxHeight.

lib/util/theme_utils.dart
```diff
@@ -34,13 +34,15 @@ class ThemeUtils {
     if (isOutline) {
       return TextButton.styleFrom(
         padding: padding,
+        fixedSize: Size(buttonWidth??Size.infinite.width, buttonHeight??Size.infinite.height),
```
vusters, line 37: This looks dangerous - setting infinite size. Did you try on all different permutations?

lib/util/theme_utils.dart
```diff
@@ -22,7 +22,7 @@ class ThemeUtils {
 
   /// this function is also called while building the button, so make sure we don't use any fallback
   /// to ensure the style reverts to the button theming
-  static ButtonStyle getButtonStyle({required bool isOutline, Color? backgroundColor, Color? color, RoundedRectangleBorder? border, EdgeInsets? padding, FontWeight? fontWeight, int? fontSize}) {
+  static ButtonStyle getButtonStyle({required bool isOutline, Color? backgroundColor, Color? color, RoundedRectangleBorder? border, EdgeInsets? padding, FontWeight? fontWeight, int? fontSize,double? buttonWidth,double? buttonHeight}) {
```
vusters, line 25: use integer for width/height as that is what the user enters. You can convert to double when you need to `buttonWidth?.toDouble() ?? ...`

lib/util/theme_utils.dart
vusters: put this Button (without explicit size) inside a Row or Column and it still work?

Moeez1999: I have tested it without giving height and width in column and row and works fine for me i am testing it with other use cases.

Issue #256, opened 2023-02-19
add Camera action to the schema


kmahmood74:
Schema is here - https://github.com/EnsembleUI/ensemble/blob/main/assets/schema/ensemble_schema.json

To test your schema changes - 

1. Use Visual Studio Code
3. use these instructions for adding custom schemas - https://dev.to/brpaz/how-to-create-your-own-auto-completion-for-json-and-yaml-files-on-vs-code-with-the-help-of-json-schema-k1i
4. point to the ensemble schema on your disk to test

vusters:
@snehmehta please close this if no longer applies.

PR #298, opened 2023-02-28
camera scheme added https://github.com/EnsembleUI/ensemble/issues/256

hamxa152:


snehmehta:
#369 has new schema changes.

PR #303, opened 2023-03-01
Ability to override field name

snehmehta:
Added new property to uploadFiles fieldName, so users can override default field name in multipart/form-data from 'files' to whatever their servers are expecting.

### DIFF
--- a/lib/util/utils.dart
+++ b/lib/util/utils.dart
@@ -316,6 +316,7 @@ class Utils {
             onError: Utils.getAction(payload['onError']),
             uploadApi: payload['uploadApi'],
             inputs: inputs,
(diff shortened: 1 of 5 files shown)


lib/util/utils.dart
```diff
@@ -316,6 +316,7 @@ class Utils {
             onError: Utils.getAction(payload['onError']),
             uploadApi: payload['uploadApi'],
             inputs: inputs,
+            fieldName: Utils.optionalString(payload['fieldName']),
```
kmahmood74, line 319: @snehmehta other option is to have fieldName as non-nullable and default to 'files'. Then use Utils.getString with the fallback as the default value. That way it will never be null so no checks needed. Your call

PR #310, opened 2023-03-01
https://github.com/EnsembleUI/ensemble/issues/307 adding id as an opt…

kmahmood74:
…ional attribute for all widgets. It will be automatically read but can still be overriden by an individual widget

### DIFF
--- a/lib/widget/helpers/controllers.dart
+++ b/lib/widget/helpers/controllers.dart
@@ -27,6 +27,7 @@ abstract class WidgetController extends Controller {
   @override
   Map<String, Function> getBaseSetters() {
     return {
+      'id': (value) => id = Utils.optionalString(value),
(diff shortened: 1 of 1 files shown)


lib/widget/helpers/controllers.dart
```diff
@@ -27,6 +27,7 @@ abstract class WidgetController extends Controller {
   @override
   Map<String, Function> getBaseSetters() {
     return {
+      'id': (value) => id = Utils.optionalString(value),
```
vusters, line 30: you don't need this. ID is added to the Scope's data context. It is handled outside the widget. We do need the schema to reflect the ID though

kmahmood74: good point, will close it

PR #323, opened 2023-03-07
changes

Moeez1999:


lib/widget/button.dart
```diff
@@ -121,7 +122,10 @@ class ButtonState extends WidgetState<Button> {
             color: widget._controller.borderColor ?? defaultShape.side.color,
             width: widget._controller.borderWidth?.toDouble() ?? defaultShape.side.width);
       }
-
+      widget._controller.borderGradient!=null?
+      border=GradientRoundedRectangleBorder(gradient: widget._controller.backgroundGradient??LinearGradient(colors: [Colors.red,Colors.black]), width: widget._controller.borderWidth?.toDouble()??defaultShape.side.width, radius:  widget._controller.borderRadius == null ?
```
vusters, line 126: 1. you want borderGradient instead of backgroundGradient
2. Why are you using some random gradient on behalf of the users? This is the only time when you can use !
```
controller.borderGraident != null ? border=controller.borderGraident!
```

lib/widget/button.dart
```diff
@@ -121,7 +122,10 @@ class ButtonState extends WidgetState<Button> {
             color: widget._controller.borderColor ?? defaultShape.side.color,
             width: widget._controller.borderWidth?.toDouble() ?? defaultShape.side.width);
       }
-
+      widget._controller.borderGradient!=null?
```
vusters, line 125: why use the shorthand notation on its own? It's harder to read. Either use proper if/else or do short-hand on the assignment

vusters (commented):
So does this gradient border works on both BoxBorder and Burtton with this code?

Moeez1999:
Its working fine for boxborder but for gradient around button it only draws border with default color not the gradient border

vusters:
then let's cut our loss for this is taking too long. Please remove both PRs, create a new one that works for both in a new review please.

Issue #213, opened 2023-02-13
Copy to clipboard


amin-nas:
As an Ensemble app developer, I want an action for copying a value to clipboard:

```yaml
Button:
  onTap:
    action: copyToClipboard
    value: copy this value

# also available as a JS function
Button:
  onTap:
    action: executeCode
    body: |
      //@code
      ensemble.copyToClipboard('copy this value');
```

amin-nas:
@hamxa152 - please also provide an option for a user to include an action once the copy process is done

```yaml
Button:
  onTap:
    action: copyToClipboard
    value: copy this value
    onComplete:
       action: # any action could go here, like showToast
```

PR #325, opened 2023-03-08
copy to clipboard done https://github.com/EnsembleUI/ensemble/issues/213

hamxa152:
**Ticket No:** https://github.com/EnsembleUI/ensemble/issues/213

**Ymal Code :** 

      - Button:
          label: copyToClipboard Js Function
          onTap:
            action: executeCode
            body: |
              ensemble.copyToClipboard('copy this value');

      - Text:
          id: testcopytoclipboard
          text: 6. Copy to Clipboard
      - Button:
         label: Copy to Clipboard
         onTap:
           action: copyToClipboard
           id: copy
           value: copy to clipboard test
           onComplete: |
            ensemble.debug('test complete')

### DIFF
--- a/.ios/Podfile
+++ b/.ios/Podfile
@@ -31,7 +31,6 @@ target 'Runner' do
   flutter_install_all_ios_pods File.dirname(File.realpath(__FILE__))
 
   pod 'FlutterPluginRegistrant', :path => File.join('Flutter', 'FlutterPluginRegistrant'), :inhibit_warnings => true
-  pod 'FirebaseFirestore', :git => 'https://github.com/invertase/firestore-ios-sdk-frameworks.git', :tag => '10.3.0'
 end
 
 post_install do |installer|
--- a/lib/framework/action.dart
+++ b/lib/framework/action.dart
@@ -222,7 +222,16 @@ class FileUploadAction extends EnsembleAction {
   String? overMaxFileSizeMessage;
 }
 
+class CopyToClipboardAction extends EnsembleAction {
+  CopyToClipboardAction({
+    this.value,
+    this.onComplete,
+  });
+  String? value;
+  EnsembleAction? onComplete;
+}
 
-enum ActionType { invokeAPI, navigateScreen, navigateModalScreen, showDialog, startTimer, stopTimer, closeAllDialogs, executeCode, showToast, getLocation, openUrl, openCamera, uploadFiles, navigateBack }
+
+enum ActionType { invokeAPI, navigateScreen, navigateModalScreen, showDialog, startTimer, stopTimer, closeAllDialogs, executeCode, showToast, getLocation, openUrl, openCamera, uploadFiles, navigateBack , copyToClipboard }
 
 enum ToastType { success, error, warning, info, custom }
--- a/lib/framework/data_context.dart
+++ b/lib/framework/data_context.dart
@@ -3,6 +3,7 @@ import 'package:ensemble/framework/config.dart';
 import 'package:ensemble/framework/device.dart';
 import 'package:ensemble/framework/error_handling.dart';
 import 'package:ensemble/framework/widget/view_util.dart';
+import 'package:ensemble/util/copyToClipboard_utils.dart';
 import 'package:ensemble/util/extensions.dart';
 import 'package:ensemble_ts_interpreter/invokables/invokablecontroller.dart';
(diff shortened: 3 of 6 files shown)


.ios/Podfile
```diff
@@ -31,7 +31,6 @@ target 'Runner' do
   flutter_install_all_ios_pods File.dirname(File.realpath(__FILE__))
 
   pod 'FlutterPluginRegistrant', :path => File.join('Flutter', 'FlutterPluginRegistrant'), :inhibit_warnings => true
-  pod 'FirebaseFirestore', :git => 'https://github.com/invertase/firestore-ios-sdk-frameworks.git', :tag => '10.3.0'
```
vusters, line 34: revert this change. We need it. Try not to check in anything that is auto-updated by the system

lib/framework/data_context.dart
```diff
@@ -311,7 +312,8 @@ class NativeInvokable with Invokable {
       ActionType.stopTimer.name: stopTimer,
       ActionType.openCamera.name: showCamera,
       ActionType.navigateBack.name: navigateBack,
-      'debug': (value) => log('Debug: $value')
+      'debug': (value) => log('Debug: $value'),
```
vusters, line 315: can you check if `log('...')` works in release mode? As in does the browser in Web release mode even print it out? If not use print() or something else.

lib/util/copyToClipboard_utils.dart
```diff
@@ -0,0 +1,17 @@
...
+
+class CopyToClipboard{
+
+
+  static Future<void> copyText(String text) async {
+    if (text.isNotEmpty) {
+      Clipboard.setData(ClipboardData(text: text));
+      return;
+    } else {
+      throw ('Please enter a string');
```
vusters, line 13: You can't just throw an exception based on a likely user error like this. The user will see a blank screen and the app will be hosed. Why crash the app?

lib/util/copyToClipboard_utils.dart
```diff
@@ -0,0 +1,17 @@
+
+
+import 'package:flutter/services.dart';
+
+class CopyToClipboard{
```
vusters, line 5: No need for create a class for really a 1-liner. Delete and use it inline please.

lib/framework/action.dart
```diff
@@ -222,7 +222,16 @@ class FileUploadAction extends EnsembleAction {
   String? overMaxFileSizeMessage;
 }
 
+class CopyToClipboardAction extends EnsembleAction {
+  CopyToClipboardAction({
+    this.value,
+    this.onComplete,
+  });
+  String? value;
+  EnsembleAction? onComplete;
```
vusters, line 231: use `onSuccess` instead. Add `onFailure` calllback in case the value is empty or there's an error (don't throw the error like you did below). The user can show the toast error if they want.

vusters:
You'll also need to update schema too.

lib/util/copyToClipboard_utils.dart
hamxa152: okay i fixed

lib/framework/data_context.dart
hamxa152: its working in release mode

lib/util/copyToClipboard_utils.dart
hamxa152: i fixed this on and set onFailure now app is not crashed

lib/framework/action.dart
hamxa152: i added these callback

lib/util/copyToClipboard_utils.dart
```diff
@@ -2,16 +2,7 @@
...
-  static Future<void> copyText(String text) async {
-    if (text.isNotEmpty) {
-      Clipboard.setData(ClipboardData(text: text));
-      return;
-    } else {
-      throw ('Please enter a string');
-    }
-  }
-
+Future<void> copyText(String text) async {
```
vusters, line 5: my point was why even have this class at all? It's a whole class that makes exactly 1 call. Remove this class and use Clipboard.setData directly.

hamxa152: its done

vusters:
This is good. Please add to ensemble-schema.

vusters (approved):
You can merge this now.

PR #333, opened 2023-03-11
Update ensemble_schema.json

amin-nas:
The schema is breaking both the auto-complete and the widget properties panel

image


amin-nas:
Fixing this issue:
https://github.com/EnsembleUI/ensemble/issues/334

Issue #347, opened 2023-03-14
Button can have any widget replacing its label


vusters:
Today Button can only have a label. We have use cases where people wants to add an icon/image next to the label. To make it extensible, we should allow user to swap in a widget in place of the label. All existing button stylings will still work.

PR #358, opened 2023-03-15
#347 Allow custom layouts in Button with labelWidget property

evshi:


lib/widget/button.dart
```diff
@@ -29,6 +31,7 @@ class Button extends StatefulWidget with Invokable, HasController<ButtonControll
   Map<String, Function> setters() {
     return {
       'label': (value) => _controller.label = Utils.getString(value, fallback: ''),
+      'labelWidget': (value) => _controller.labelWidget = value,
```
vusters, line 34: we do this everywhere for widget but something to think about: we have to assume what the user enters here will resolve to a valid widget later (or else we show 'invalid widget' text). Probably need a `Utils.getWidget(value)` here where we check against the list of valid widgets, plus any user-defined custom widget.

vusters (approved):

vusters:
You can add the schema and add to Kitchen sink example now please https://studio.ensembleui.com/app/e24402cb-75e2-404c-866c-29e6c3dd7992/screen/09c1087b-f9ee-4a8c-9286-e0e881184c07

lib/widget/button.dart
evshi: The check makes sense but how could we handle it differently if it doesn't resolve to a valid widget? For example, fail silently or return a default widget?

evshi: Got it. Shall I create a new issue for the widget util validation or is there one already?

vusters: https://github.com/orgs/EnsembleUI/projects/2/views/1?pane=issue&itemId=23052103

assets/schema/ensemble_schema.json
```diff
@@ -814,12 +814,15 @@
...
       "type": "object",
-      "required": ["label"],
       "properties": {
         "label": {
           "type": "string",
           "description": "The button label"
         },
+        "labelWidget": {
+          "$ref": "#/$defs/Row-payload",
+          "description": "A row widget containing the button contents. For example, a Text and an Icon"
```
vusters, line 824: Use this: "the custom widget to display in place of the label."

vusters (approved):
Change the widget reference and you should be good to merge

lib/widget/button.dart
```diff
@@ -82,20 +90,39 @@ class ButtonState extends WidgetState<Button> {
...
+      endingIcon = scopeManager?.buildWidgetFromDefinition(widget._controller.endingIcon);
+    }
+    List<Widget> labelParts = [Text(Utils.translate(widget._controller.label ?? '', context))];
+    if (startingIcon != null) {
+      labelParts.insertAll(0, [startingIcon]);
+    }
+    if (endingIcon != null) {
+      labelParts.addAll([endingIcon]);
+    }
+    Widget labelLayout = Row(
```
evshi, line 111: @vusters it seems like Row is required to render icons after the button and have the ability to have both icons. Another benefit is Row automatically responds to the device text direction, i.e. RTL. Let me know if this is concerning to you or you think we should move in a different direction.

lib/widget/button.dart
```diff
@@ -89,25 +92,37 @@ class ButtonState extends WidgetState<Button> {
...
-    Widget label;
-    if (scopeManager != null && widget._controller.labelWidget != null) {
-      label = scopeManager.buildWidgetWithScopeFromDefinition(widget._controller.labelWidget);
-    } else {
-      label = Text(
-          Utils.translate(widget._controller.label ?? '', context));
+    Widget? startingIcon;
+    Widget? endingIcon;
+    if (widget._controller.startingIcon != null) {
+      startingIcon = scopeManager?.buildWidgetFromDefinition(
```
vusters, line 98: use Icon.fromModel(IconModel)

lib/widget/button.dart
vusters: This is fine.

lib/widget/button.dart
```diff
@@ -89,25 +92,37 @@ class ButtonState extends WidgetState<Button> {
...
+          widget._controller.startingIcon);
+    }
+    if (widget._controller.endingIcon != null) {
+      endingIcon = scopeManager?.buildWidgetFromDefinition(widget._controller.endingIcon);
+    }
+    List<Widget> labelParts = [Text(Utils.translate(widget._controller.label ?? '', context))];
```
vusters, line 104: User may want to control the gap between the icons and label. Add an `iconGap` attribute? by default it should be 0.

lib/widget/button.dart
```diff
@@ -89,25 +92,37 @@ class ButtonState extends WidgetState<Button> {
...
+    if (startingIcon != null) {
+      labelParts.insertAll(0, [startingIcon]);
```
vusters, line 106: why not just use `add(endingIcon)` and `insert(0, startingIcon)`?

evshi: This is a holdover from when I was going to add gaps in between the icon and label.

lib/widget/button.dart
evshi: This is easy enough to add, but I think this is a slippery slope where we might consider going back to a custom label widget Row so they can control things like spacing without needing to pixel push. Do you feel strongly we should add this now or can we defer this until we get some more user feedback?

vusters (approved):

Issue #181, opened 2023-01-23
Invisible widgets take space


amin-nas:
Widgets that are invisible take space instead of being completely invisible.

kmahmood74:
please provide a test case in the form of a EDL definition so this may be assigned to someone who can just get it from the description.

amin-nas:
Test case:
```yaml
View:
  Column:
    styles: { gap: 8, backgroundColor: yellow }
    children:
      - Column:
          visible: false
          children:
            - Text:
                text: I am invisible
      - Column:
          visible: false
          children:
            - Text:
                text: I am invisible
      - Column:
          visible: false
          children:
            - Text:
                text: I am invisible
      - Column:
          visible: false
          children:
            - Text:
                text: I am invisible
      - Column:
          visible: false
          children:
            - Text:
                text: I am invisible
      - Column:
          visible: true
          children:
            - Text:
                text: I am visible
```
<img width="969" alt="Screenshot 2023-01-25 at 11 30 33 AM" src="https://user-images.githubusercontent.com/491202/214505375-7663dd27-c2ac-4d80-ba1a-f16510810f67.png">

vinothzyngl:
It took space because of adding a `gap: 8` in the styles of column @amin-nas

vinothvino42:
This is fixed in this PR #370

PR #370, opened 2023-03-21
Fixed issue of gap in the BoxLayout widgets

vinothvino42:
Ticket #181

In this PR, Fixed the BoxLayout widgets that add spaces for their children even if the child's visible property is false


lib/layout/box_layout.dart
```diff
@@ -202,7 +202,8 @@ class BoxLayoutState extends WidgetState<BoxLayout> with TemplatedWidgetState {
       List<Widget> updatedChildren = [];
       for (var i=0; i<children.length; i++) {
         updatedChildren.add(children[i]);
-        if (i != children.length-1) {
+        final child = children[i] as BoxLayout;
```
vinothvino42, line 205: Is it safe to downcasting the child to BoxLayout type right? @vusters I hope the gap property is only for BoxLayout and its subclasses like Row, Column and Flex widgets

lib/layout/box_layout.dart
```diff
@@ -202,7 +202,8 @@ class BoxLayoutState extends WidgetState<BoxLayout> with TemplatedWidgetState {
...
+        if (i != children.length - 1 && child.controller.visible) {
```
vusters, line 206: bool visibleChild = child is HasController && (child as HasController).controller.visible

vinothvino42: Thanks, @vusters I tried your solution but the controller is of type ```Controller```. So I just downcasted to ```WidgetController``` to access the visible bool property

```
final visibleChild = children[i] is HasController && ((children[i] as HasController).controller as WidgetController).visible;
```

I just pushed the code, can you review it?

vusters (approved):
Please change as commented and you are good to go

lib/layout/box_layout.dart
```diff
@@ -202,8 +202,8 @@ class BoxLayoutState extends WidgetState<BoxLayout> with TemplatedWidgetState {
       List<Widget> updatedChildren = [];
       for (var i=0; i<children.length; i++) {
         updatedChildren.add(children[i]);
-        final child = children[i] as BoxLayout;
-        if (i != children.length - 1 && child.controller.visible) {
+        final visibleChild = children[i] is HasController && ((children[i] as HasController).controller as WidgetController).visible;
```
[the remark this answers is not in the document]
vinothvino42: `children[i] is HasController && ((children[i] as HasController).controller is WidgetController)`

It returns a `bool` instead of a `WidgetController` to access visible bool property

vusters (approved):

Issue #408, opened 2023-04-03
setting dropdown's value to an invalid value does not clear the selected value


kmahmood74:
this was reported by @gaurikalp 

Setting a dropdown's value to an invalid value (e.g. setting it to null or empty string) should clear out the selected value but that doesn't work when done in code. 

This is because the maybeValue is not a setter and is being directly set. As a result, there is no setState() being called. 

Inside dropdown.dart - 
```
class SelectOneController extends FormFieldController {
  List<SelectOneItem>? items;

  // this is our value but it can be in an invalid state.
  // Since user can set items/value in any order and at anytime, the value may
  // not be one of the items, hence it could be in an incorrect state
  dynamic maybeValue;
  int gap = 0;
  bool autoComplete = false;

  framework.EnsembleAction? onChange;
}
```
and it's set directly in the setters - 
```
  @override
  Map<String, Function> setters() {
    return {
      'value': (value) => _controller.maybeValue = value,
      'items': (values) => updateItems(values),
      'onChange': (definition) => _controller.onChange =
          framework.EnsembleAction.fromYaml(definition, initiator: this),
      'itemsFromString': (dynamic strValues) => setItemsFromString(strValues),
      'itemsFromArray': (dynamic arrValues) => setItemsFromArray(arrValues),
      'autoComplete': (value) =>
          _controller.autoComplete = Utils.getBool(value, fallback: false),
    };
  }
```
so there is no setState being called after it is set and thus no update

vinothvino42:
@gaurikalp @kmahmood74 
Add an id to the Dropdown widget and change it from any widget by calling `dropdownId.value = '`' value to reset the dropdown value and the hint text will be shown.


https://user-images.githubusercontent.com/12869588/230012667-2069ce41-9bec-4591-9bbc-6281f7ed60b3.mp4

<details>
  <summary>YAML Code</summary>

  ### YAML Code
  ```yaml
View:
  header:
    title: Reset Dropdown
  Column:
    styles:
      gap: 24
      padding: 24

    children:
      - Markdown:
          text: |
            ## Reset Dropdown value example

      - Dropdown:
          id: dropDown
          label: 1 - array of strings
          hintText: Select something
          items:
            - Red
            - Blue
            - Green
          onChange: |
            //@code
            onChangeDemo.text = 'You selected: ' +this.value;

      - Text:
          id: onChangeDemo

      - Button:
          label: Reset
          onTap: |
            //@code
            dropDown.value = '';
[...]

PR #417, opened 2023-04-06
Fixed raw autocomplete field reset value issue

vinothvino42:
In this PR, I fixed the raw autocomplete dropdown issue #408


lib/widget/input/dropdown.dart
```diff
@@ -179,14 +185,28 @@ class SelectOneController extends FormFieldController {
...
 
 class SelectOneState extends FormFieldWidgetState<SelectOne> {
   final focusNode = FocusNode();
+  late TextEditingController rawTextEditingController;
+  late StreamSubscription<dynamic> streamSubscription;
+
   @override
   void initState() {
+    rawTextEditingController = TextEditingController();
+    streamSubscription =
```
vusters, line 202: This pattern is OK but in Controller we already have a way to invoke the widgetState's setState(). Just use `notifyListeners()`.

lib/widget/input/dropdown.dart
```diff
@@ -179,14 +185,28 @@ class SelectOneController extends FormFieldController {
...
   int gap = 0;
   bool autoComplete = false;
+  final StreamController<dynamic> fieldController = StreamController<dynamic>();
 
   framework.EnsembleAction? onChange;
 }
 
 class SelectOneState extends FormFieldWidgetState<SelectOne> {
   final focusNode = FocusNode();
+  late TextEditingController rawTextEditingController;
```
vusters, line 196: just create the instance here and remove late.

lib/widget/input/dropdown.dart
```diff
@@ -99,6 +98,13 @@ abstract class SelectOne extends StatefulWidget
     return null;
   }
 
+  void _setValue(dynamic value) {
+    _controller.maybeValue = value;
+    if (value is String && value == '') {
```
vusters: See my comment on the Github issue. <Empty string> maybe an actual value users want to use, so we reset only on `null`, plus expose a clear() method for the user to clear out the value.

lib/widget/input/dropdown.dart
```diff
@@ -99,6 +98,13 @@ abstract class SelectOne extends StatefulWidget
     return null;
   }
 
+  void _setValue(dynamic value) {
+    _controller.maybeValue = value;
```
vusters, line 102: I know this is the existing code but it maybe better to check the validity here and either ignore to set the value (vs waiting for later). No more maybeValue.

lib/widget/input/dropdown.dart
```diff
@@ -61,7 +60,7 @@ abstract class SelectOne extends StatefulWidget
   @override
   Map<String, Function> setters() {
     return {
-      'value': (value) => _controller.maybeValue = value,
+      'value': (value) => _setValue(value),
```
vusters, line 63: if you put this method on the controller, you can force the widgetState to call setState() using `notifyListeners()` inside Controller

vusters (approved):

vinothvino42:
@vusters @kmahmood74 
Added clear method and you can set null to a value property of dropdown to reset the value.

**Sample YAML**:
```yaml
View:
  header:
    title: Reset Dropdown
  Column:
    styles:
      gap: 24
      padding: 24

    children:
      - Markdown:
          text: |
            ## Reset Dropdown value example

      - Dropdown:
          id: dropDown
          label: 1 - array of strings
          hintText: Select something
          items:
            - Red
            - Blue
            - Green

      - Text:
          id: onChangeDemo
          text: ${dropDown.value}

      - Button:
          label: Reset
          onTap: |
            //@code
            dropDown.clear();
            //dropDown.value = null;

```

Issue #419, opened 2023-04-07
navigateBack now working


sharjeelyunus:
navigateBack should send you to the last screen you were on without of saving it in back history, navigate screen saves the screen as history.

navigateBack isn't even working on Kitchen Sink App

https://studio.ensembleui.com/app/e24402cb-75e2-404c-866c-29e6c3dd7992/screen/XjvL2XseLnRvYO4FS82e#

evshi:
I can take a look at this since I last touched it. It should work as a YAML string but also a YAML object.

PR #421, opened 2023-04-08
#419 Parse actions with null payloads

evshi:


lib/framework/action.dart
```diff
@@ -482,7 +482,7 @@ abstract class EnsembleAction {
     if (action is YamlMap) {
       ActionType? actionType = ActionType.values.from(action.keys.first);
       dynamic payload = action[action.keys.first];
-      if (actionType != null && payload is YamlMap) {
+      if (actionType != null && payload is YamlMap?) {
```
vusters, line 485: what's the question mark for?

vusters: GPT taught me the ?. Learning new stuff every day.

vusters (approved):

lib/framework/action.dart
evshi: So it also matches null, which was the regression

Issue #330, opened 2023-03-10
Conditional UI


snehmehta:
Ability to render widget based on condition.

Sample EDL for conditional UI. 
```yaml
if: 
  condition: ${a == b} # hypothetical condition that returns bool 
  body: 
    Text:
      text: I'll render based on `condition`
```
_Side note : Syntax might change. Open for suggestions._

kmahmood74:
how about - 

```
Conditional:
  - condition:
     if: ${isImage == true}
     then:
       Image:
         src: ${filePath}
  - condition:
     if: ${isVideo == true}
     then:
       Video:
        src: ....
  - default:
      Text:
        text: no match
```

vusters:
latest:
```
Conditional:
  - if: ${isImage == true}
     then:
       Image:
         src: ${filePath}
  - if: ${isVideo == true}
     then:
       Video:
        src: ....
  - default:
      Text:
        text: no match
```

snehmehta:
> If a condition matches, execute `then` and exit. This is modeled after Switch's break. If no condition was matched, execute the `default` block. This means default is required.

Default won't be needed in all cases.

kmahmood74:
hi @snehmehta right after you are done with the background downloading, please work on this one as the highest priority

snehmehta:
https://studio.ensembleui.com/app/e24402cb-75e2-404c-866c-29e6c3dd7992/screen/XljPegGrdieyKjy16URC

PR #437, opened 2023-04-13
Conditional UI

snehmehta:
Ticket: #330

```yaml
- Conditional:
    conditions:
      - if: ${1 != 1}
        then:
          Text:
            text: If condition works
      - elif: ${1 != 1}
        then:
          Text:
            text: First elif condition works
      - elif: ${1 == 1}
        then:
          Text:
            text: Second elif condition works
      - else:
          Text:
            text: oops no match

```


kmahmood74:
hi @snehmehta I think we should stick to what everyone knows - elseif and not invent elif. Already people have to learn Ensemble, we must not add more cognitive burden when this works exactly like if/then/else. elseif is a well know thing that people understand.

kmahmood74:
@snehmehta sorry I may take that back. I think you are inspired by python's elif. that is quite relevant for us as python uses indentation and doesn't have braces. 

PHP uses elseif

@vusters @evshi what do you think? I am fine with elif as well.

lib/widget/conditional.dart
```diff
@@ -0,0 +1,118 @@
...
+    try {
+      return scopeManager != null &&
+          conditions != null &&
+          (conditions is YamlList) &&
+          conditions.isNotEmpty &&
+          conditions.first.containsKey('if') &&
+          conditions.first.containsKey('then');
+    } on Exception catch (_) {
+      throw LanguageError(
+          'In Proper structure, make sure atleast if then condition ');
```
evshi, line 83: In proper -> Improper

evshi (commented):
`elif` looks good to me. Is there any value in having the `then`?

lib/widget/conditional.dart
snehmehta: Thanks, fixed.

snehmehta:
@evshi `then` mainly to pick widget definition, also readability if true then build this widget.

vusters:
elseif, elsif, elif - personally like the first option - read best for non developer. Plus in the Property Panel we probably will spell out 'else if', so without space in the Code editor is the closest to it.

snehmehta:
> You need to handle binding changes (watch for all changes to all of the expressions, then re-execute the entire if/elseif/then). See Scope's listen() method. It essentially want a setter on this Conditional widget that it can call when bindings have changed.

thanks! We'll do the changes.

snehmehta:
Updated example with binding changes.

```yaml
- Button:
    label: Capture
    onTap:
      openCamera:
        id: cameraId
        options:
          mode: photo

- Conditional:
    conditions:
      - if: ${cameraId.files.length == 1}
        Text:
          text: Single Image
      - elseif: ${cameraId.files.length == 2}
        Text:
          text: Double Image
      - elseif: ${cameraId.files.length >= 3}
        Text:
          text: multiple image
      - else:
        Text:
          text: No images

```

Also, agreed with @vusters elseif is better in readability than elif.

kmahmood74 (approved):

kmahmood74:
@snehmehta please merge

Issue #440, opened 2023-04-14
Allow navigating to Flutter screen from within Ensemble


kmahmood74:
copied from [discussion](https://github.com/EnsembleUI/ensemble/discussions/391#discussioncomment-5615695) - 

Use case:

I have an existing Flutter app that I want to add Ensemble screens to
From an Ensemble screen I want to navigate to a Flutter screen
Example:
```
- Button:
          label: Navigate and pass inputs to a Flutter screen
          onTap:
            executeCode:
              body: |
                //@code
                ensemble.navigateFlutterScreen(
                  'MyFlutterWidget', # this is my screen's name
                  { # and here are the constructor parameters
                    'input': someInput 
                  }
                );
```

- You should be able to do this directly in the navigateScreen action as well without having to write code. 
- Support the same for showModal action as well
- Support the other way around too i.e. navigating to an Ensemble screen from Flutter
- As the first step, propose the yaml and the corresponding javascript syntax before proceeding with the implementation

kmahmood74:
assigning to @vusters for assessment and re-assignment

kmahmood74:
@snehmehta please update, this was completed a while ago I thought?

PR #442, opened 2023-04-18
Navigate to flutter screen

snehmehta:
Ticket: #440 

EDL remains the same i.e pass the screen name.

Only configuration needs to be done on flutter side that is pass custom widgets along with name.
Example:
```dart
void main() async {
  WidgetsFlutterBinding.ensureInitialized();

  // (optional) pre-initialize asynchronously so Ensemble is ready
  // when your app switches to Ensemble screens. Add `await` before
  // this statement if your first page is an Ensemble's page.
  Ensemble().initialize();

  /// (optional) set map of custom flutter widget that can be called from ensemble
  Ensemble().setCustomFlutterWidgets(
    {
      'testScreen': (context, args) =>
          TestScreen(name: args?['name'] ?? 'Hello'),
    },
  );
  runApp(const MyApp());
}
```

EDL to navigate to `TestScreen` will be no different, 

```yaml
- Button:
    label: Test Screen
    onTap:
      navigateModalScreen:
        name: testScreen
        inputs:
          name: PageName

```

### DIFF
--- a/lib/ensemble.dart
+++ b/lib/ensemble.dart
@@ -26,6 +26,9 @@ import 'html_shim.dart' if (dart.library.html) 'dart:html' show window;
 import 'framework/theme/theme_loader.dart';
 import 'layout/ensemble_page_route.dart';
 
+typedef CustomBuilder = Widget Function(
+    BuildContext context, Map<String, dynamic>? args);
+
 /// Singleton Controller
 class Ensemble {
   static final Ensemble _instance = Ensemble._internal();
@@ -36,9 +39,14 @@ class Ensemble {
 
   late FirebaseApp ensembleFirebaseApp;
   static final Map<String, dynamic> externalDataContext = {};
+  Map<String, CustomBuilder> externalScreenWidgets = {};
 
   static final RouteObserver<PageRoute> routeObserver =
       RouteObserver<PageRoute>();
+
+  void setExternalScreenWidgets(Map<String, CustomBuilder> widgets) {
+    externalScreenWidgets = widgets;
+  }
 
   /// initialize all the singleton/managers. Note that this function can be
   /// called multiple times since it's being called inside a widget.
@@ -243,6 +251,7 @@ class Ensemble {
     String? screenName,
     bool? asModal,
     Map<String, dynamic>? pageArgs,
+    bool isExternal = false,
   }) {
     PageType pageType = asModal == true ? PageType.modal : PageType.regular;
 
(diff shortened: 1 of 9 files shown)


kmahmood74:
Thanks @snehmehta how is the opposite interaction achieved i.e. from Flutter to a particular Ensemble screen? Let's get that done as well if it is not a major effort

kmahmood74:
@snehmehta the only issue I see is if there is a name conflict between a screen in my app vs the custom widget I am passing in. Secondly, how would the preview distinguish between 'testScreen' missing or 'testScreen' being a custom widget as the preview doesn't know about flutter custom widgets. It should degrade gracefully. 

Suggest adding a parm such as - 

```
- Button:
    label: Test Screen
    onTap:
      navigateModalScreen:
        name: testScreen
        type: externalWidget
        inputs:
          name: PageName
```
I am intentionally calling it externalWidget so as not to tie to Flutter only

lib/ensemble.dart
```diff
@@ -30,11 +33,17 @@ class Ensemble {
   /// the configuration required to run an App
   EnsembleConfig? _config;
 
+  Map<String, WidgetBuilder>? customFlutterWidgets;
```
vusters, line 36: call it externalScreenWidgets?

lib/screen_controller.dart
```diff
@@ -720,8 +720,12 @@ class ScreenController {
     Map<String, dynamic>? pageArgs,
   }) {
     PageType pageType = asModal == true ? PageType.modal : PageType.regular;
-
-    Widget screenWidget =
+    Widget? screenWidget;
+    if (Ensemble().customFlutterWidgets?.containsKey(screenName) ?? false) {
```
vusters, line 724: This will just take the user to Page Not Found error when running in Preview. Instead we want to say something like "External Screen Placeholder", and include a link to an Ensemble Doc's page where we explain how to use it with ensemble_starter to get your page to appear.

This means we probably want the user to explicitly say `external: true` when using navigateScreen*. This way you just check for the flag presence if not found, then show that message

snehmehta: make sense, we'll do the changes

kmahmood74:
why is this PR still open? I thought this was completed a while ago? @snehmehta ?

kmahmood74:
@snehmehta please close this if it is done

kmahmood74 (approved):

Issue #432, opened 2023-04-12
Named styles


amin-nas:
I want to be able to define a group of styling, globally across the app, and apply them to widgets in my screens.

```yaml

# in theme
Colors: 
  ...
Widgets: 
  ...
Styles:
  myPersonCard:
    shadowColor: 0xFFCECECE
    shadowOffset: [ 0, 0 ]
    shadowRadius: 2
    backgroundColor: white
    borderRadius: 12


# in screen
View:
  body:
    Column:
      styleName: myPersonCard
      styles:
         # styles here will take precedent
```

vinothvino42:
@amin-nas @vusters Is the theme needs to be in a separate file? or can it be placed in any files? It's supposed to be a single global theme right? All other widgets can use it. So how we're going to do this in the ensemble studio?

amin-nas:
> @amin-nas @vusters Is the theme needs to be in a separate file? or can it be placed in any files? It's supposed to be a single global theme right? All other widgets can use it. So how we're going to do this in the ensemble studio?

In Studio, there's a theme section for each app. Styles defined in the theme would be available globally.
Styles defined as part of a screen would be available inside that screen only. If a style is defined in both the theme and the screen, the screen style would apply.

vinothvino42:
```
# define your app theme here
Colors:
  primary: 0xFFF19D39
  
Widgets:
  Button:
    borderRadius: 100
    color: black
    padding: 20 24
    backgroundColor: 0xFFF1D05C
  Input:
    borderColor: 0xFFEDEDF0
  
Styles:
  myButtonTheme:
    type: Button
    color: white
    backgroundColor: green
    borderRadius: 12
```

@amin-nas Here, In this case which one will take precedent? Styles or Widgets?

amin-nas:
Order of precedent:
1. inline styles
2. styles
3. widgets

vinothvino42:
@vusters Can you review this PR - https://github.com/EnsembleUI/ensemble/pull/466/files

kmahmood74:
@vinothvino42 is this done? Please update

kmahmood74:
@vusters given how important this is and it has been hanging around for 6 months! this is crazy. Assigning to you as it's not going anywhere

kmahmood74:
@vusters get it done yourself please. We can't clean up the utter mess that our styles are today without this in place.

kmahmood74:
[...]

PR #466, opened 2023-05-03
WIP: Added named styles

vinothvino42:


lib/widget/button.dart
```diff
@@ -33,6 +37,15 @@ class Button extends StatefulWidget
 
   @override
   Map<String, Function> setters() {
+    final styleModel = GetIt.I<StyleProvider>();
+    final StyleTheme? style = styleModel.getStyleName(_controller.styleName);
+
+    if (style != null) {
+      _controller.color = _controller.color ?? style.color;
```
vusters, line 44: This doesn't scale. None of the widgets should have to know about styleName.

vinothvino42: This kind of style will be applied to a widget only if it has the `styleName` right? So each widget is supposed to know its styleName right? If it has the styleName, then the style with that name is applied to that widget.

vusters: @amin-nas i thought we said styleName has no type, and can be applied to any widget.

amin-nas: Correct. I believe the ticket here was updated: https://github.com/EnsembleUI/ensemble/issues/432

vinothvino42: @vusters Okay no need to add type for the widget in styleName but we need to store the styleName for each widget. If it has the styleName, then the style with that name is applied to that widget.

kmahmood74:
@vinothvino42 is this complete? if yes, always close after merging please.

vinothvino42:
> @vinothvino42 is this complete? if yes, always close after merging please.

This is not done. We kept this one as a low priority later and worked on other high-priority issues. @vusters Can we do this again?

kmahmood74:
@vusters what's your take, see Vinoth's comment?

Issue #465, opened 2023-05-02
Remove margin in Carousel


sharjeelyunus:
There is an automatic margin on images on in Carousel

```
Carousel:
            styles: 
              height: 220
              layout: auto
              indicatorType: circle
              indicatorPosition: bottom
              indicatorWidth: 6
              indicatorHeight: 6
              singleItemWidthRatio: 1
            item-template:
              data: ${record.fields.photo}
              name: image
              template:
                Image:
                  source: ${image.url}
```

![Image](https://user-images.githubusercontent.com/61178058/235729592-a971793f-16ab-4d08-83b9-82883221250e.png)

kmahmood74:
@vinothvino42 please take a look at this. This is urgently needed for a customer

vinothvino42:
Okay, I'll fix it @kmahmood74

PR #472, opened 2023-05-04
Fixed carousel padding issue

vinothvino42:
Ticket: #465
In this PR, I fixed the padding issue and also removed the default padding for the leading and trailing items.

`gap: 10` -> It adds the gap between the children. Default value is 10
`leadingGap: 10` -> It adds the gap at the start of the first item. Default value is 0
`trailingGap: 10` -> It adds the gap at the end of the last item. Default value is 0

### DIFF
--- a/lib/widget/carousel.dart
+++ b/lib/widget/carousel.dart
@@ -211,6 +211,7 @@ class CarouselState extends WidgetState<Carousel> with TemplatedWidgetState {
     return BoxWrapper(
       widget: carousel,
       boxController: widget._controller,
+      ignoresPadding: true,
       ignoresDimension:
           true, // width/height shouldn't be apply in the container
     );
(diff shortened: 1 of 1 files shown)


lib/widget/carousel.dart
```diff
@@ -241,8 +242,8 @@ class CarouselState extends WidgetState<Carousel> with TemplatedWidgetState {
     // wrap each child inside Container to add padding and gap
     double gap =
         widget._controller.gap?.toDouble() ?? MyController.defaultItemGap;
-    double leadingGap = widget._controller.leadingGap?.toDouble() ?? gap / 2;
-    double trailingGap = widget._controller.trailingGap?.toDouble() ?? gap / 2;
+    double leadingGap = widget._controller.leadingGap?.toDouble() ?? 0;
```
vusters, line 245: So the reason leadingGap = gap/2 is because we want to scroll items from the screen edge, so margin is often time not effective. When margin is not used, user have to set both the leadingGap and trailingGap to be exactly the same for it to look right.

I think we are OK to set this to 0, but it just won't look good out of the box (unless you really want zero margin). Go ahead and merge this in.

vinothvino42: Okay I’ll merge this PR

Issue #469, opened 2023-05-03
Theming bugs


amin-nas:
1. Disabled input text is too light. Set default  to #878199

![image](https://user-images.githubusercontent.com/491202/235993838-f8376e69-4de7-4e7a-8f9f-2c69956a184f.png)

2. Divider's in DataGrid are too dark. Set default to #DAD7E0

3. Button labels are too small. Set default to 16 medium.

4. Setting button border color, fontSize, and fontWeight from Theme does not work.
```yaml
Widgets:
  Button:
    borderColor: 0xFF0077CC
    fontSize: 18
    fontWeight: bold
```

PR #474, opened 2023-05-04
fix default style

snehmehta:
Theming bugs #469 
 
 *New: Inline google fonts support

### DIFF
--- a/lib/framework/theme/theme_loader.dart
+++ b/lib/framework/theme/theme_loader.dart
@@ -44,6 +44,7 @@ mixin ThemeLoader {
       textTheme: _buildTextTheme(),
       outlinedButtonTheme: OutlinedButtonThemeData(
         style: OutlinedButton.styleFrom(
+          textStyle: const TextStyle(fontSize: 16),
           shape: RoundedRectangleBorder(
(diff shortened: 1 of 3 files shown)


lib/framework/theme/theme_loader.dart
```diff
@@ -139,16 +141,14 @@ mixin ThemeLoader {
...
-      if (fontFamily == null) throw Exception();
-      defaultStyle = GoogleFonts.getFont(fontFamily, color: Colors.black);
-    } catch (e) {
-      defaultStyle = const TextStyle(
-          fontFamily: 'Inter',
-          fontWeight: FontWeight.w400,
-          color: Colors.black);
-    }
+      if (fontFamily != null) {
+        defaultStyle =
```
vusters, line 148: This can return null if fontFamily=giberish

kmahmood74: GoogleFonts.getFont throws an exception when the fontFamily does not exist, see here - https://pub.dev/documentation/google_fonts/latest/google_fonts/GoogleFonts/getFont.html

So this is not an issue as the code is already catching exceptions and ignoring them.

kmahmood74 (approved):

Issue #423, opened 2023-04-10
Need a onItemTap for all itemTemplate widgets


kmahmood74:
```
        - Column:
            item-template:
              data: ${getMockHomepage.body.data.homepage.mainImages}
              name: mainImages
              template:
                ImageWidget:
                  inputs:
                    source: ${mainImages}
            onItemTap:
              navigateScreen:
                name: Image
                inputs:
                  data: >-
                    ${getMockHomepage.body.data.homepage.mainImages[this.selectedItemIndex]}
```
doesn't work

vinothvino42:
PR - https://github.com/EnsembleUI/ensemble/pull/490

kmahmood74:
this was fixed by Vinoth, closing it

PR #490, opened 2023-05-09
Added onItemTap for all item-template widgets

vinothvino42:
Ticket: #423 
In this PR, I added `onItemTap` setter and `selectedItemIndex` getter property for the following item template widgets

1. Carousel
2. GridView
3. Flow
4. DataGrid

Added `onItemTap` setter for the flex widgets, there is **no** `selectedItemIndex` getter property. To access the index using event's data like this `${event.data.selectedItemIndex}`

1. Column
2. Row
3. Flex


lib/widget/gesture_detector.dart
```diff
@@ -0,0 +1,39 @@
+import 'package:ensemble/framework/widget/widget.dart';
```
vusters, line 1: move this to Util package

vusters (approved):

Issue #513, opened 2023-05-17
WalletConnect, connect with one custom wallet, i.e. MetaMask, Trust, using the enum.


Ammar-Ishfaq:
need additional functionality for the meta mask specific login

kmahmood74:
@snehmehta should this be closed as complete or not to be done?

kmahmood74:
closing as not needed

PR #517, opened 2023-05-18
Specific wallet in connectWallet

snehmehta:
Ticket: #513

Ability to specify, specific wallet to connect. 

```yaml
Button:
  label: Connect to Metamask
  onTap:
    connectWallet:
      id: wallet
      wcProjectId: 77740b7e86cfcba224464923b8e115e3
      wallet: metamask #here specify wallet, like metamask, trust, rainbow..
      appMetaData:
        name: Ensemeble app
        description: Ensemble forever
        url: https://ensembleui.com/
        iconUrl: https://ensembleui.com/assets/images/logo.svg
      onComplete: |
        //@code

        address.text = wallet.addresses;
```

It's optional for android and web, but mandatory for IOS. Reason, android is designed to handle multiple applications subscribing to the same deep linking schema.

### DIFF
--- a/assets/schema/ensemble_schema.json
+++ b/assets/schema/ensemble_schema.json
@@ -2190,6 +2190,10 @@
                   "type": "string",
                   "description": "Wallet connect project Id, get it from wallet connect dashboard"
                 },
+                "wallet": {
+                  "type": "string",
+                  "description": "Optional for Android, Required in IOS, when specified walletconnect will open and connect to specified wallet. e.g metamask, trust, rainbow..."
+                },
                 "onComplete": {
                   "$ref": "#/$defs/Action-payload",
                   "description": "Execute another Action upon successful upload of files"
--- a/lib/screen_controller.dart
+++ b/lib/screen_controller.dart
@@ -473,7 +473,15 @@ class ScreenController {
             );
             return;
           }
-          launchUrlString(uri, mode: LaunchMode.externalApplication);
+
+          uri = action.wallet == null
+              ? uri
+              : '${action.wallet?.toLowerCase().trim()}://wc?uri=$uri';
+
+          launchUrlString(
+            uri,
+            mode: LaunchMode.externalApplication,
+          );
         });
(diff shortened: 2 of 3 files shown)


assets/schema/ensemble_schema.json
```diff
@@ -2190,6 +2190,10 @@
                   "type": "string",
                   "description": "Wallet connect project Id, get it from wallet connect dashboard"
                 },
+                "wallet": {
+                  "type": "string",
+                  "description": "Optional for Android, Required in IOS, when specified walletconnect will open and connect to specified wallet. e.g metamask, trust, rainbow..."
```
kmahmood74, line 2195: perhaps we should make it required for both iOS and Android otherwise people will test their app on Android and it works and doesn't work on iOS. What do you think?

snehmehta: Can do, for android we can add new type `any` to open android tray of all wallet connection and for IOS any won't work and we'll show error for specific wallet

lib/screen_controller.dart
```diff
@@ -473,7 +473,15 @@ class ScreenController {
             );
             return;
           }
-          launchUrlString(uri, mode: LaunchMode.externalApplication);
+
+          uri = action.wallet == null
+              ? uri
+              : '${action.wallet?.toLowerCase().trim()}://wc?uri=$uri';
```
vusters, line 479: do you have a complete list? if so force the action.wallet to be an enum and throw LanguageError if not in the list. Otherwise this is ok

vusters (approved):

kmahmood74:
@snehmehta this has been approved a while ago. If it is still valid, merge and close

kmahmood74:
this is still open, @snehmehta either close or discard

snehmehta:
This capability is added using wallet connect V1 which is down, so we have to rewrite it using wallet connect v2.
Since, v2 bring breaking changes we can't use this code hence closing this PR without merging.

Issue #563, opened 2023-06-07
Support for Local and Push Notifications in iOS and Android


kmahmood74:
Two requirements - 

1. Allow Ensemble apps to trigger local notifications. Example - calendar app sends a reminder for an event. 
2. Provide the ability for Ensemble apps to handle push notifications from the server. For example, an app developer should be able to write a handler in js

This is a very much needed feature that we have talked about for a while and would be a great addition to the platform. 

cc @vusters

PR #571, opened 2023-06-13
Remote & local Notifications

snehmehta:
Ticket:  #563

sample EDL:
```dart

View:
  header:
    title: Home

  onLoad:
    notification:
      onReceive:
        showNotification:
          title: ${title} //  extraDataContext, simply specify key of a given map
          body: ${body}
      onTap:
        navigateScreen:
          name: Form
```
For specific notification configuration, e.g Firebase messaging the configuration need to be done.

```dart
  // inside main.dart=> main function
  FirebaseMessaging.onMessage.listen((RemoteMessage message) {
    Ensemble.externalDataContext.addAll({
      'title': message.notification?.title,
      'body': message.notification?.body,
    });
    notificationUtils.handleRemoteNotification();
  });

  FirebaseMessaging.onMessageOpenedApp.listen((RemoteMessage message) {
    Ensemble.externalDataContext.addAll({
      'title': message.notification?.title,
      'body': message.notification?.body,
    });
    notificationUtils.handleRemoteNotificationOpened();
  });
```


kmahmood74:
@snehmehta @vusters I think notifications should be handled at a more global level and not within an individual screen. You need to be able to provide a global notifications handler function that can then load any screen or perform any action. It doesn't make sense to have that notification handler under the View node. Thoughts?

snehmehta:
I agree, it shouldn't be under View node. 
The current implementation is such, it is getting set on home screen, but is it handle in global fashion. So whether the user is on home screen or not the code will get executed on notification received. 

But again mention it under View node doesn't seem good pattern, I was thinking we should have some global EDL yaml for such setters.

kmahmood74:
@snehmehta this has been pending for a long time. Could you please make the changes and wrap it up

lib/screen_controller.dart
```diff
@@ -496,6 +498,18 @@ class ScreenController {
...
+
+      if (isEnabled && action.onAccept != null) {
+        executeAction(context, action.onAccept!);
+      }
+
+      if (!isEnabled && action.onReject != null) {
+        executeAction(context, action.onReject!);
+      }
+
+      if (action.onComplete != null) executeAction(context, action.onComplete!);
```
vusters, line 501: what is the onComplete for if you already have onAccept or on Reject?

lib/util/notification_utils.dart
```diff
@@ -24,6 +26,33 @@ class _NotificationUtils {
       iOS: initializationSettingsIOS,
     );
     await localNotificationsPlugin.initialize(initializationSettings);
+
+    return _requestPermissions();
+  }
+
+  Future<bool> _requestPermissions() async {
+    if (kIsWeb) {
+      return true;
```
vusters, line 35: what does this mean?

lib/screen_controller.dart
```diff
@@ -492,15 +492,24 @@ class ScreenController {
...
 
-  String evaluateString(String input, Map<String, String> contextMap) {
-    return input.replaceAllMapped(RegExp(r'\$(\w+)'), (match) {
-      final key = match.group(1);
-      return contextMap[key] ?? match.group(0) ?? input;
+  String evaluateExtraContext(String input) {
+    final contextMap = Ensemble.externalDataContext;
+    RegExp placeholderRegex = RegExp(r'\${(.*?)}');
+
+    String evaluatedString = input.replaceAllMapped(placeholderRegex, (match) {
```
vusters, line 511: don't manage this yourself. Use DataContext.eval() just like everywhere else in the system. The strings need to be evalulate with all the other context (e.g. on the current page)

vusters:
I get the local notification part, that gets dispatch manually with EnsembleAction. But how do you handle incoming Remote notification? where does the handler get set and how does it get triggered?

snehmehta:
> I get the local notification part, that gets dispatch manually with EnsembleAction. But how do you handle incoming Remote notification? where does the handler get set and how does it get triggered?

That user has to provide inside main.dart. For e.g if they want to use firebase messaging
```dart
  FirebaseMessaging.onMessage.listen((RemoteMessage message) {
    Ensemble.externalDataContext.addAll({
      'title': message.notification?.title,
      'body': message.notification?.body,
    });
    notificationUtils.handleRemoteNotification();
  });
```

lib/screen_controller.dart
snehmehta: yes, removed

lib/util/notification_utils.dart
snehmehta: image
I picked it from flutter local notification source code..
Basically flutter web has the local notification permission.

vusters (approved):
You are good to check in after the ENsembleServices injection.

PR #605, opened 2023-06-27
Add styles and item properties to bottom nav bar schema

vinothvino42:
In this PR, I updated the following for the `BottomNavBar` widget

1. Added styles like `color, floatingBackgroundColor, floatingIconColor, selectedColor`
2. Added items properties like `activeIcon, floating, floatingAlignment, floatingMargin`


assets/schema/ensemble_schema.json
```diff
@@ -3336,7 +3336,93 @@
           "properties": {
             "BottomNavBar": {
               "description": "Use the bottom navigation bar",
-              "$ref": "#/$defs/MenuBase"
+              "allOf": [
```
evshi, line 3339: You don't need allOf here, you can combine the two objects into one schema

vinothvino42: I just updated it. Can you review it again?

PR #619, opened 2023-07-05
Add autoplay to carousel and update schema

vinothvino42:
In this PR, I added autoplay and duration to the carousel widget

### DIFF
--- a/assets/schema/ensemble_schema.json
+++ b/assets/schema/ensemble_schema.json
@@ -1768,7 +1768,7 @@
                     },
                     "autoplay": {
                       "type": "boolean",
-                      "description": "If true, Enables autoplay, sliding one page at a time. Default (true)"
+                      "description": "If true, Enables autoplay, sliding one page at a time. Default (false)"
                     },
                     "autoplayInterval": {
                       "type": "integer",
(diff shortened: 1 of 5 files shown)


lib/widget/carousel.dart
```diff
@@ -40,6 +40,9 @@ class Carousel extends StatefulWidget
           _controller.layout = CarouselLayout.values.from(input),
       'autoLayoutBreakpoint': (value) =>
           _controller.autoLayoutBreakpoint = Utils.optionalInt(value, min: 0),
+      'autoplay': (value) => _controller.autoplay = Utils.optionalBool(value),
+      'autoplayInterval': (value) =>
```
vusters, line 44: add min: 1

assets/schema/ensemble_schema.json
```diff
@@ -1766,6 +1766,14 @@
                     "indicatorHeight": {
                       "type": "integer"
                     },
+                    "autoplay": {
```
vusters, line 1769: default is false according to your code. Default=false is what we want, just just change the schema here.

vinothvino42: @vusters Can you check the new changes? I updated the two things (min value and default false changes) which you mentioned.

lib/widget/carousel.dart
vusters: yes, since you use seconds, you want the user to enter at least 1 sec.

vusters (approved):
Ship it

PR #653, opened 2023-07-17
allow expressions for boolean values

sharjeelyunus:


vusters:
Almost all properties (except id, children, item-template) can be expressions, so we shouldn't need to add them to the schema (otherwise it'll balloon up and become unmanageable. Instead of modifying the schema, can we make the changes on the Editor itself to account for expressions? @sharjeelyunus @evshi

kmahmood74:
this is needed for the yaml editor as well right? how would the dev know that s/he can write an expression instead of a boolean? 

@evshi please see Vu's comment from a month ago

evshi:
@kmahmood74 yes it should ideally be represented in the schema so it's reflected in both the visual editor and yaml editor. IIRC we discussed this in standup and we were going to move forward with just the visual editor changes for now and address this change later when we do schema validation and automation

kmahmood74:
@evshi @vusters this is stuck, please close if not needed?

PR #663, opened 2023-07-21
Access secrets from data context

evshi:


lib/framework/secrets.dart
```diff
@@ -0,0 +1,39 @@
...
+
+class SecretsStore with Invokable {
+  // ignore since we override getProperty
+  @override
+  Map<String, Function> getters() {
+    throw UnimplementedError();
+  }
+
+  @override
+  getProperty(prop) {
```
evshi, line 15: @vusters I'm not sure how I should handle Future values when reading back from secure storage. Any tips?

lib/ensemble_provider.dart
```diff
@@ -166,6 +172,12 @@ class AppModel {
             baseUrl: doc.data()?['appBaseUrl'],
             useBrowserUrl: Utils.optionalBool(doc.data()?['appUseBrowserUrl']),
             envVariables: envVariables);
+      } else if (doc.data()?['type'] == ArtifactType.secrets.name) {
+        dynamic rawSecrets = doc.data()!['secrets'];
+        if (rawSecrets is Map) {
+          secrets = {};
+          rawSecrets.forEach((key, value) => secrets![key] = value);
```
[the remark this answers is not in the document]
vusters: Yea, can we store secrets encrypted on the DB, then decode it with a key on the client?

lib/framework/secrets.dart
vusters: All getters are synchronous - exposing async is tricky for the user to use. Can we simply read all the secrets into memory and getProperty(key) returns synchronously? If the secrets are modified from the server (local overrides are static) we can simply update the map.

evshi: So secrets are resolved in the following way?

1. Fetch from local, remote, secure storage
2. Encrypt and store in secure storage (if necessary)
3. Keep copy in memory
4. Secrets read from memory

What value is secure storage adding here?

lib/framework/secrets.dart
```diff
@@ -36,4 +54,9 @@ class SecretsStore with Invokable {
   Map<String, Function> setters() {
     return {};
   }
+
+  _cacheSecret(String key, dynamic value) {
+    secretCache[key] = value;
+    StorageManager().writeSecurely(key: key, value: value.toString());
```
vusters, line 60: why store to storage for?

vusters (approved):
I'm not sure why you need to store to StorageManager. Otherwise this looks good.

lib/framework/secrets.dart
evshi: For the ensemble framework code to be able to access app secrets, e.g. auth

PR #667, opened 2023-07-24
update label and description in GLOBAL code schema AND update API sch…

saddamrashid:
…ema body type to onOf structure


assets/schema/ensemble_schema.json
```diff
@@ -193,10 +195,12 @@
...
           },
           "body": {
-            "type": [
-              "string",
-              "object"
-            ],
+            "type": {
+              "oneOf": [
+                { "type": "string" },
+                { "type": "object" }
```
evshi, line 201: Please add `additionalProperties` of type: string so user can specify custom keys for their body https://rjsf-team.github.io/react-jsonschema-form/docs/usage/objects/#additional-properties. There is an example in the invokeAPI body.

saddamrashid: done

evshi (approved):

Issue #669, opened 2023-07-24
Need number as inputType in TextField widget


anserwaseem:
As seen from the screenshot attached, there's no way to restrict user for only numeric values. Please add number/numeric inputType for TextField.

![image](https://github.com/EnsembleUI/ensemble/assets/32386713/2d5b4357-e901-4b6a-9a0d-3751f9d74495)

vusters:
Remember to add the validator for each inputType we support. It is not just the visual keyboard.

vinothvino42:
It's completed. PR - #676

PR #676, opened 2023-07-25
Add few input types, input formatter and update validator

vinothvino42:
Ticket: #669

In this PR, I updated the following things
1. Added input types like `number`, `text`, `url`, `datetime`
2. Added `InputFormatter` for `digits` (Now user can't copy and paste any characters or symbols if we set the input formatters, currently I added it only for digits)
3. Changed the validation logic. The user passed validator will take the first precedence instead of our default validation.

We have the following TextInputType by default in Flutter
1. `number` - It doesn’t have any symbols but only numbers
2. `phone` - It has numbers and symbols (For ex: +, *, #)
4. `emailAddress` - Characters with @ symbol
5. `text, name, streetAddress, visiblePassword` - Only characters (Both types are the same keyboard layout)
6. `url` - It has dot, forward slash, .com with characters keyboard layout
7. `datetime` - It has number with more symbols

Here 4th point, it looks the same keyboard layout for those text input types. So I added only the text instead of others like name, streetAddress, visiblePassword

### DIFF
--- a/lib/util/input_formatter.dart
+++ b/lib/util/input_formatter.dart
@@ -1,0 +1,14 @@
+import 'package:flutter/services.dart';
+
+class InputFormatter {
+  static List<TextInputFormatter> getFormatter(String? inputType) {
+    if (inputType == null) return [];
+
+    switch (inputType) {
+      case 'number':
+        return [FilteringTextInputFormatter.digitsOnly];
+      default:
+        return [];
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


snehmehta (approved):

amin-nas:
@vinothvino42 please reflect the new options in schema.

vinothvino42:
> @vinothvino42 please reflect the new options in schema.

Added now

Issue #626, opened 2023-07-06
Ability to access device contacts


amin-nas:
I want to be able to prompt user to provide access to device contacts, and get contact list upon user granting access.

vinothvino42:
@vusters @amin-nas 
Sample **YAML**

I am using https://pub.dev/packages/fluttercontactpicker/ flutter package, which shows a native contact picker.

```dart
Button:
  label: Pick Phone Contact
  onTap:
    pickPhoneContact:
      id: phoneContactId
      options:
        linkToken: ${getLinkTokenFromServer.body.linkToken}
      onSuccess: |
        //@code
        console.log("Contact fetched successfully");
        console.log(event.data);
        console.log("Phone Number" + event.data['phoneNumber']);

      onExit: |
        //@code
        console.log("Something went wrong!");
```

amin-nas:
@vinothvino42 - I want to be able to get the full contact list so that I can display them with Ensemble widget of my choice. Here's an example:

![Invite Friend - post permission](https://github.com/EnsembleUI/ensemble/assets/491202/77be56af-a806-44bb-b42f-690676f5e59b)

vinothvino42:
Okay @amin-nas Then I'll use this package instead - https://pub.dev/packages/flutter_contacts
Here, you need to use your own view to populate all the contacts

amin-nas:
@vinothvino42 we use `onAccept` and `onReject` for notification access request. Let's change the above to be consistent.

cc @vusters

vinothvino42:
@amin-nas I think we've used `onSuccess` in other widgets too.

PR #691, opened 2023-08-02
Add ensemble contacts action

vinothvino42:
Ticket: #626 

```yaml
Button:
  label: Get Phone Contacts
  onTap:
    getPhoneContacts:
      onSuccess: |
        //@code
        console.log("Phone Number Fetched Successfully");
        ensemble.storage.set('contacts', event.data['contacts']);

      onError: |
        //@code
        console.log("Something went wrong!");
```


lib/framework/stub/contacts_manager.dart
```diff
@@ -0,0 +1,19 @@
+import 'package:ensemble/framework/error_handling.dart';
+
+typedef ContactSuccessCallback = void Function(
+    List<Map<String, dynamic>> contacts);
```
evshi, line 4: Can we have stronger typing on contacts?

lib/screen_controller.dart
```diff
@@ -235,6 +236,31 @@ class ScreenController {
...
+          );
+        }
+      }, (error) {
+        if (action.getOnError(dataContext) != null) {
+          executeActionWithScope(
+            context,
+            scopeManager!,
+            action.getOnError(dataContext)!,
+            event: EnsembleEvent(
+              action.initiator!,
```
evshi, line 259: Include error details in event?

lib/framework/stub/contacts_manager.dart
vinothvino42: @evshi I need to have an exact clone of https://pub.dev/documentation/flutter_contacts/latest/contact/Contact-class.html then, we no need to use the Contact class (because it's from flutter_contacts). So can I do an exact clone of it with all nested classes with properties or just create a class which exposes only phone numbers?

lib/screen_controller.dart
vinothvino42: Added errors in the data param of EnsembleEvent

evshi (approved):
Thanks for adding the complete class! Was there an issue with just adding interfaces?

PR #692, opened 2023-08-02
Ensemble Chat Widget

snehmehta:
Schema:
```yaml
Chat:
  ws: ws://localhost:8000/ws
  initialMessageApi: http://localhost:8000/messages/
  user:
    id: 1
    name: Spark
```


evshi (commented):
Seems like you have a few changes unrelated to chat, is that intentional?

lib/ensemble_app.dart
```diff
@@ -197,3 +197,64 @@ class EnsembleAppState extends State<EnsembleApp> {
         home: Scaffold(backgroundColor: loadingBackgroundColor, body: widget));
   }
 }
+
+class Message {
```
vusters: Wrong place as these classes are unrelated. Please move them elsewhere to its own class. EnsembleApp is the root widget that folks will integrate with and look at. It should be minimal and make sense.

lib/widget/chat.dart
```diff
@@ -0,0 +1,354 @@
...
+              ),
+            ),
+          ),
+        ),
+      ],
+    );
+  }
+}
+
+class Message {
```
vusters: These classes are repeated.

vusters (commented):
don't use the same names we already have (rename models & utils to chat_models and chat_utils).

lib/widget/chat/models.dart
```diff
@@ -0,0 +1,78 @@
...
+
+  Map<String, dynamic> toJson() {
+    return {
+      'id': id,
+      'name': name,
+    };
+  }
+}
+
+String generateRandomString({int length = 8}) {
```
vusters: move this to the root Utils class so it can be re-used

lib/widget/chat.dart
vusters, line 1: move the chat.dart inside the chat folder. This is the common pattern we have.

lib/widget/input/dropdown.dart
```diff
@@ -73,6 +73,7 @@ abstract class SelectOne extends StatefulWidget
   @override
   Map<String, Function> methods() {
     return {
+      'getValue': () => getValue(),
```
vusters, line 76: why not just use getters value?

lib/widget/visualization/chart_js.dart
```diff
@@ -136,6 +137,10 @@ class ChartJsState extends WidgetState<ChartJs> {
 
   @override
   Widget buildWidget(BuildContext context) {
+    if (Platform.isWindows) {
```
vusters, line 140: does it even work outside Web? if it doesn't just do this for everything except kIsWeb

lib/widget/chat/utils.dart
```diff
@@ -0,0 +1,15 @@
+import 'dart:math' as math;
+
+String generateRandomString({int length = 8}) {
```
vusters: duplicate

lib/widget/input/slider.dart
```diff
@@ -23,14 +23,14 @@ class EnsembleSlider extends StatefulWidget
...
     return {
-      'value': () => _controller.value.round(),
+      'value': () => _controller.value,
     };
   }
 
   @override
   Map<String, Function> methods() {
     return {
-      'getValue': () => _controller.value,
```
vusters: don't need this - just use value getter

lib/widget/chat.dart
```diff
@@ -0,0 +1,354 @@
...
+
+  @override
+  Map<String, Function> methods() {
+    return {};
+  }
+
+  @override
+  Map<String, Function> setters() {
+    return {
+      'url': (value) => controller.url =
```
vusters: Make Url required and don't fill in any default value.

lib/widget/chat.dart
```diff
@@ -0,0 +1,354 @@
...
+  WebSocketChannel? channel;
+
+  @override
+  void initState() {
+    super.initState();
+
+    _user = const User(
+      id: 'abc123',
+    );
+    final uri = Uri.tryParse('ws://127.0.0.1:8000/ws');
```
vusters: how is this different than the URL above? Plus don't hardcode these in.

lib/widget/chat.dart
```diff
@@ -73,7 +71,7 @@ class EnsembleChatState extends framework.WidgetState<EnsembleChat> {
 
     _assistant = const User(id: '123');
 
-    final uri = Uri.tryParse('ws://35.202.36.118:8000/ws');
+    final uri = Uri.tryParse('wss://sparksavings.ddns.net/python/ws');
```
vusters: what does this mean? Why hardcode?

vusters:
Feedback: there are a few code duplications per commit. While you fixed them it's harder to follow, comment, and then having to delete the comments. I usually try to not make a commit unless it's a semi-solid and self-explanatory.

lib/widget/chat.dart
```diff
@@ -0,0 +1,354 @@
...
+  Map<String, Function> methods() {
+    return {};
+  }
+
+  @override
+  Map<String, Function> setters() {
+    return {
+      'url': (value) => controller.url =
+          Utils.getString(value, fallback: 'wc://localhost:8000/wc'),
+      'userId': (newValue) =>
```
vusters: not sure I understand why the userId is specified here

vusters:
It seems the chat widget is hardcoded and not something people can just use just yet. That is OK but don't add the schema - we'll add schema once it is generally available. 

Once you fixed some of the points here, go ahead and merge in.

vusters (approved):

lib/widget/input/dropdown.dart
snehmehta: image
The thought behind this was when we render any dynamic widget (e.g in chat) we can call a single method `getValue` for all ensemble input widget.

lib/widget/visualization/chart_js.dart
snehmehta: Yes, it work on web, android and IOS. I was under the impression it works as well on macOS but it doesn't so I'll update the condition.

snehmehta:
> It seems the chat widget is hardcoded and not something people can just use just yet. That is OK but don't add the schema - we'll add schema once it is generally available.
> 
> Once you fixed some of the points here, go ahead and merge in.

Yes, right now it's more specific toward our sparksavings requirement, I will have to do another pass to make it more generic.


Some of the duplication / outdated comment mentioning are about the code which is moved to [ensemble_chat](https://github.com/EnsembleUI/ensemble_chat) so I go through comments and update the ensemble_chat, while merging this PR.

PR #698, opened 2023-08-04
Added a maskedValue and unmaskedValue getter properties to TextInput

vinothvino42:


vusters:
always have the schema along with your changes

lib/widget/input/form_textfield.dart
```diff
@@ -38,6 +40,14 @@ class TextInput extends BaseTextInput {
     return setters;
   }
 
+  @override
+  Map<String, Function> getters() {
+    return {
+      'maskedValue': () => _controller.inputFieldAction?.getMaskedValue(),
+      'unmaskedValue': () => _controller.inputFieldAction?.getUnmaskedValue(),
```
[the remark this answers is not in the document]
vinothvino42: Here, you want both masked and unmasked values right?

Maybe I can remove the `maskedValue` getter property (because the value getter property will return what we entered in the textfield, with masked values). Still, you want the `unmaskedValue` getter property right?

vusters (approved):
Reminder that we only need `value` and `maskedValue`. Once you've removed unmasked, you can check this in.

kmahmood74:
@vinothvino42 value getter should return unmasked value. So please remove the getter for unmaskedValue. You still need the getter for maskedValue.

You don't need approval, just make the change and check it in

kmahmood74:
if this is done, close the PR please @vinothvino42

vusters:
@vinothvino42 I scanned through your code. I think you have to fix your pattern. Please look into your code why the action is null out.

vinothvino42:
> @vinothvino42 I scanned through your code. I think you have to fix your pattern. Please look into your code why the action is null out.

Now I have done it with a simple approach. Can you please take a look at my new changes?

vusters (commented):
so now the text's value is stored in both TextEditingController and our controller, and you run some processing to sync them. It's not ideal and may cause potential bugs. Can the value be only in textEditingController?

lib/widget/input/form_textfield.dart
```diff
@@ -400,6 +413,24 @@ class TextInputState extends FormFieldWidgetState<BaseTextInput>
         ));
   }
 
+  void _updateTextInputController() {
+    widget.controller.maskedValueText = getTextInputValue(isMasked: true);
+    widget.controller.value = getTextInputValue(isMasked: false);
```
evshi, line 418: Just get both unmasked and masked values here once instead of calling a function twice with both true and false?

evshi:
@vusters having the value in two places is fine as long as we can enforce that one is always the source of truth.

lib/widget/input/form_textfield.dart
vinothvino42: I updated it. Can you check again?

lib/widget/input/form_textfield.dart
```diff
@@ -218,6 +222,8 @@ class TextInputState extends FormFieldWidgetState<BaseTextInput>
 
   @override
   void initState() {
+    widget.textController =
+        TextEditingController(text: widget.controller.value);
```
evshi, line 226: Why initialize a new controller here instead of setting the value? Based on comment from line 92, textController should be source of truth for value?

vinothvino42: It initializes the text editing controller with the setter value (controller.value). 

It's because previously we had `textController.text` for setters and getters, so the value is added to the TextField directly. 

Now I'm using `controller.value `for setters and getters, so I need to reinitialize the text editing controller with this `controller.value` in the init method (226 line)

See line number 33 and 103 in the files changed tab

evshi: Hmmm @vusters what do you think about this? It would be inverting the flow of data.

Issue #631, opened 2023-07-09
Header's backgroundColor and margin/padding


sibteali786:
Header enhancements:

- Enable setting padding: 0 - there is built in padding that user cannot control.
- Enable setting backgroundColor when user scrolls . Currently a color is auto-set depending on seed color and user cannot change it.
![image](https://github.com/EnsembleUI/ensemble/assets/62849614/37199261-c998-43ec-a364-a6d67087e3ad)
![image](https://github.com/EnsembleUI/ensemble/assets/62849614/281e6344-a233-4bec-b2b6-6508cc927ec3)

snehmehta:
@sibteali786 can you please add me as collaborator 
sneh@ensembleui.com

sibteali786:
@snehmehta sent you invitation

snehmehta:
Thanks! @sibteali786

PR #702, opened 2023-08-07
set header bg color on scroll

snehmehta:
Ticket: #631 

Adding new property to header widget, set background when scrolled. Below is sample yaml.

```yaml
View:
  styles:
    useSafeArea: true
    scrollableView: true

  header:
    title: Images
    styles:
      backgroundColor: white
      surfaceTintColor: brown #new property

```

### DIFF
--- a/assets/schema/ensemble_schema.json
+++ b/assets/schema/ensemble_schema.json
@@ -42,7 +42,7 @@
                   "$ref": "#/$defs/typeColors",
                   "description": "By default the background color uses the theme's 'primary' color. You can override the header's background color here."
                 },
-                "onScrollBackgroundColor": {
+                "surfaceTintColor": {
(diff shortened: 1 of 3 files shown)


assets/schema/ensemble_schema.json
```diff
@@ -42,6 +42,10 @@
                   "$ref": "#/$defs/typeColors",
                   "description": "By default the background color uses the theme's 'primary' color. You can override the header's background color here."
                 },
+                "onScrollBackgroundColor": {
```
vusters, line 45: The onScroll* name doesn't make sense. Not that surfaceTintColor makes sense either but at least it's the same name Flutter uses.

vusters (approved):
call it by the original name and you can check this in.

PR #726, opened 2023-08-18
Update Keychain service

vinothvino42:
Ticket: #730 

Added clearKeychain method and update saveToKeychain

### DIFF
--- a/lib/framework/data_context.dart
+++ b/lib/framework/data_context.dart
@@ -363,28 +363,29 @@ class NativeInvokable with Invokable {
   }
 
   Future<void> saveToKeychain(String key, String value) async {
-    if (defaultTargetPlatform == TargetPlatform.iOS) {
-      try {
-        const platform = MethodChannel('com.ensembleui.dev/safari-extension');
-        final _ = await platform
-            .invokeMethod(ActionType.saveToKeychain.name, {key: value});
-      } on PlatformException catch (e) {
-        throw LanguageError(
-            'Failed to invoke ensemble.saveToKeychain. Reason: ${e.toString()}');
-      }
+    if (defaultTargetPlatform != TargetPlatform.iOS) {
+      return;
+    }
+    try {
+      const platform = MethodChannel('com.ensembleui.dev/safari-extension');
+      final _ = await platform
+          .invokeMethod(ActionType.saveToKeychain.name, {key: value});
+    } on PlatformException catch (e) {
+      throw LanguageError(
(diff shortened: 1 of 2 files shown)


lib/framework/data_context.dart
```diff
@@ -361,12 +361,16 @@ class NativeInvokable with Invokable {
...
+    if (defaultTargetPlatform == TargetPlatform.iOS) {
+      try {
+        const platform = MethodChannel('com.ensembleui.dev/safari-extension');
+        final _ = await platform.invokeMethod('saveToKeychain', {key: value});
+      } on PlatformException catch (e) {
+        throw LanguageError(
+            'Failed to invoke ensemble.saveToKeychain. Reason: ${e.toString()}');
+      }
+    } else {
+      throw LanguageError('ensemble.saveToKeychain is only supported for iOS');
```
vusters, line 373: What does this mean? This function can now never be used in Ensemble unless you only deploy on iOS?

vinothvino42: Now I've done this, especially for iOS and Tin. The method points to safari extension

vusters:
@vinothvino42 see what I wrote above. In our YAML code we simply use `ensemble.saveToKeyChain()`. It should be a no-op on platforms that don't support it. Currently it throws an exception on other platforms

vinothvino42:
> @vinothvino42 see what I wrote above. In our YAML code we simply use `ensemble.saveToKeyChain()`. It should be a no-op on platforms that don't support it. Currently it throws an exception on other platforms

Okay, I'll remove the throwing an exception on other platforms. So now it'll simply work for iOS and nothing will happen for other platforms when using `ensemble.saveToKeychain` or `ensemble.clearKeychain`

evshi (approved):

lib/framework/data_context.dart
```diff
@@ -361,12 +363,28 @@ class NativeInvokable with Invokable {
   }
 
   Future<void> saveToKeychain(String key, String value) async {
-    try {
-      const platform = MethodChannel('com.ensembleui.dev/safari-extension');
-      final dynamic result =
-          await platform.invokeMethod('saveToKeychain', {key: value});
-    } on PlatformException catch (e) {
-      print('Failed to get value: ${e.message}.');
+    if (defaultTargetPlatform == TargetPlatform.iOS) {
```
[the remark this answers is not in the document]
vinothvino42: Okay I'll update now

vusters (approved):

Issue #701, opened 2023-08-06
Ability to detect if permission was granted by user to use device capabilities


amin-nas:
As an app developer, I want to know if user has granted access to specific device capabilities inside my Ensemble app so I can show appropriate UI.

1. push notification (immediate need)
2. location
3. storage
4. camera
5. microphone
6. contacts

vusters:
These are mostly async, so it's ok to implement as Actions. Please also see if you can expose these to Javascript (sync)

vinothvino42:
@vusters Do we need to check whether the permissions are enabled/disabled? OR the services are enabled/disabled by the EnsembleServices (or EnsembleModules)?

vusters:
This is to check if the user has granted/denied permissions. So the result can be:
1. TRUE - user has granted permission
2. FALSE - user has denied
3. null - service is not enabled, or status is not available.

vusters:
@vinothvino42 I already did #1. Please do the rest on this jira.

vinothvino42:
Today I did it for location like how you have done it for notification but I think we need to move all the above services as a separate flutter package, right? After that can I expose a permission check method to use it from the ensemble based on that I'll set the status?

vinothvino42:
@evshi Can you review this PR again? Added permission check support for the camera also

vusters:
A sure way to make sure we don't inadvertently requires user permission when checking for permission is:
1. move the actual implementation to its own package (which it should be anyway)
     a. If the package is turned on, it can return true/false
     b. if the package is not turned on, the stub function can just return false.

vinothvino42:
This issue is closed with PR #736 

It includes a permission check for the following services. 

1. Notification (Added by vu)
2. Location
3. Contacts
4. Camera

We're not using any storage or microphone direct service which requires permission. We will reopen the issue if we need anything later. Closing it as of now.

PR #736, opened 2023-08-21
WIP: Add permission check for some services

vinothvino42:
Ticket: #701

1. Notification (Added by vu)
2. Location
3. Contacts
4. Camera


lib/framework/permissions_manager.dart
```diff
@@ -11,33 +14,63 @@ class PermissionsManager {
...
+            case AuthorizationStatus.provisional:
+              status = true;
+              break;
+            case AuthorizationStatus.denied:
+              status = false;
+              break;
+            case AuthorizationStatus.notDetermined:
+            default:
+              status = null;
+              break;
```
evshi, line 38: We should move this location specific logic into the utilities or this class will become very large.

evshi (approved):

lib/framework/stub/camera_manager.dart
```diff
@@ -4,11 +4,18 @@ import 'package:ensemble/framework/scope.dart';
...
 abstract class CameraManager {
+  Future<bool?> hasPermission();
   Future<void> openCamera(BuildContext context, ShowCameraAction cameraAction,
       ScopeManager? scopeManager);
 }
 
 class CameraManagerStub extends CameraManager {
+  @override
+  Future<bool?> hasPermission() {
+    throw ConfigError(
```
vusters, line 15: We don't throw an Error here, simply false. This is valid use case. We only throw the error if the user attempts to use the camera.

vinothvino42: I updated it. Can you check now?

lib/framework/stub/contacts_manager.dart
```diff
@@ -20,8 +20,7 @@ class ContactManagerStub extends ContactManager {
 
   @override
   Future<bool> requestPermission() {
-    throw ConfigError(
-        "Phone Contact Service is not enabled. Please review the Ensemble documentation.");
+    return Future.value(false);
```
vusters, line 23: is this checking for permission or request user access? If it's the latter you'd want to throw the Error right?

vinothvino42: According to the doc, there's no dedicated method to check for the permission is enabled or not in the flutter_contacts package.

FYI, See the following
1. https://pub.dev/documentation/flutter_contacts/latest/flutter_contacts/FlutterContacts-class.html
2.https://github.com/QuisApp/flutter_contacts/blob/b1fd688c34968c30e76977479e4ed392bcb3bfda/ios/Classes/SwiftFlutterContactsPlugin.swift#L447C15-L447C32

There's a way to do that natively, But this package didn't expose the method `authorizationStatus(for:)` to check the permission which returns `CNAuthorizationStatus` - https://developer.apple.com/documentation/contacts/cncontactstore/1403143-authorizationstatus

vusters (approved):

Issue #716, opened 2023-08-13
Add a Loading Container


vusters:
https://discord.com/channels/1031982848485359626/1091875561447641108/1138898040833458268

vinothvino42:
@vusters @evshi What about this syntax for loading async task and showing results based on the state

```dart
AsyncView:
  api: getUser
  onIdle:
    Button:
       label: Get Account Summary
  onLoading:
    Shimmer:
      color: #ABCDEF
  onError:
    Text:
      text: ${getUser.error}
  onSuccess:
    AccountsSummary:
```

vinothvino42:
NEW EDL:

```yaml
LoadingContainer:
  isLoading: ${myAPI.body == null}
  useShimmer: true
  widget: MyWidget
  loadingWidget: MyLoadingWidget
```

vinothvino42:
@vusters @evshi PR - https://github.com/EnsembleUI/ensemble/pull/761

vinothvino42:
It is now closed with #761 and #782 PR

PR #761, opened 2023-08-24
WIP: Add LoadingContainer widget

vinothvino42:
Ticket: #716 

**Sample YAML**:

```yaml
View:
  header:
    title: LoadingContainer
  styles:
    scrollableView: true
  body:
    Column:
      styles:
        gap: 16
        padding: 30
        crossAxis: center
        mainAxis: center
      children:
        - LoadingContainer:
            id: loadingWidget
            isLoading: false
            useShimmer: true
            widget:
              Text:
                id: dataId
                text: Data State
            loadingWidget:
              Progress:
                display: circular
                styles:
                  size: 24

        - Button:
            label: Change State
            onTap: |
              //@code
              loadingWidget.isLoading = !loadingWidget.isLoading;

```

**Experiment**:
Here, I created a two widget.
1. LoadingContainer
2. Box

In the `LoadingContainer`, if the useShimmer is true. I'll simply wrap the loading widget with Shimmer.
`Box` widget is used to create simple box like Container widget in Flutter. Internally, I'm using BoxWrapper with BoxController to style it. 

```yaml
View:
  header:
    title: LoadingContainer
  styles:
    scrollableView: true
  body:
    Column:
      styles:
        gap: 16
        padding: 20
        crossAxis: center
        mainAxis: center
      children:
        - LoadingContainer:
            id: loadingWidget
            isLoading: true
            useShimmer: true
            widget:
              Text:
                id: dataId
                text: Data State
            loadingWidget:
              ListViewLoader:

        - Button:
            label: Change State
            onTap: |
              //@code
              loadingWidget.isLoading = !loadingWidget.isLoading;

ListViewLoader:
  body:
    Column:
      styles:
        gap: 20
      children:
        - ListTile:
        - ListTile:
        - ListTile:
        - ListTile:
        - ListTile:

ListTile:
  body:
    Row:
      styles:
        gap: 10
      children:
        - Box:
            styles:
              {
                height: 70,
                width: 70,
                borderRadius: 8,
                backgroundColor: 0xFFF5F5F7,
              }
        - Column:
            styles:
              gap: 10
            children:
              - Box:
                  styles:
                    {
                      height: 18,
                      width: 260,
                      borderRadius: 7,
                      backgroundColor: 0xFFF5F5F7,
                    }
              - Box:
                  styles:
                    {
                      height: 16,
                      width: 200,
                      borderRadius: 7,
                      backgroundColor: 0xFFF5F5F7,
                    }
```

https://github.com/EnsembleUI/ensemble/assets/12869588/eb68924d-19b1-4f65-967f-3bb611eeba46

### DIFF
--- a/assets/schema/ensemble_schema.json
+++ b/assets/schema/ensemble_schema.json
@@ -808,6 +808,38 @@
         }
       }
     },
+    "LoadingContainer-payload": {
+      "type": "object",
+      "properties": {
+        "id": {
+          "type": "string",
+          "description": "A unique identifier for this widget"
+        },
+        "isLoading": {
+          "type": "boolean",
+          "description": "Whether to display loading widget"
+        },
+        "useShimmer": {
+          "type": "boolean",
+          "description": "Whether to use a shimmer widget in loading state"
+        },
+        "widget": {
+          "$ref": "#/$defs/Widget",
+          "description": "The widget to display when non loading state"
+        },
+        "loadingWidget": {
+          "$ref": "#/$defs/Widget",
+          "description": "The widget to display when loading state"
+        },
+        "styles": {
+          "allOf": [
+            {
+              "$ref": "#/$defs/baseStyles"
+            }
+          ]
+        }
+      }
+    },
     "PopupMenu-payload": {
       "type": "object",
       "properties": {
--- a/lib/widget/box.dart
+++ b/lib/widget/box.dart
@@ -1,0 +1,63 @@
+import 'package:ensemble/framework/scope.dart';
+import 'package:ensemble/framework/widget/widget.dart';
+import 'package:ensemble/widget/helpers/controllers.dart';
+import 'package:ensemble/widget/helpers/widgets.dart';
+import 'package:ensemble_ts_interpreter/invokables/invokable.dart';
+import 'package:flutter/material.dart';
+
+import '../framework/view/page.dart';
+
+class Box extends StatefulWidget
+    with Invokable, HasController<BoxContainerController, BoxState> {
+  static const type = 'Box';
+  Box({Key? key}) : super(key: key);
+
+  final BoxContainerController _controller = BoxContainerController();
+  @override
+  get controller => _controller;
+
+  @override
+  State<StatefulWidget> createState() => BoxState();
+
+  @override
+  Map<String, Function> getters() {
+    return {};
+  }
+
+  @override
+  Map<String, Function> methods() {
+    return {};
+  }
+
+  @override
+  Map<String, Function> setters() {
+    return {
+      'widget': (widget) => _controller.widget = widget,
+    };
+  }
+}
+
+class BoxContainerController extends BoxController {
+  dynamic widget;
+}
+
+class BoxState extends WidgetState<Box> {
+  @override
+  Widget buildWidget(BuildContext context) {
+    ScopeManager? scopeManager = DataScopeWidget.getScope(context);
+
+    return BoxWrapper(
+      widget: _buildWidget(widget._controller.widget, scopeManager),
+      boxController: widget._controller,
+      ignoresPadding: false,
+      ignoresDimension: false,
+    );
+  }
+
+  Widget _buildWidget(dynamic widgetDefinition, ScopeManager? scopeManager) {
+    if (scopeManager != null && widgetDefinition != null) {
+      return scopeManager.buildWidgetFromDefinition(widgetDefinition);
+    }
+    return const SizedBox();
+  }
+}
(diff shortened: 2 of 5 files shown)


assets/schema/ensemble_schema.json
```diff
@@ -808,6 +808,38 @@
...
+    "LoadingContainer-payload": {
+      "type": "object",
+      "properties": {
+        "id": {
+          "type": "string",
+          "description": "A unique identifier for this widget"
+        },
+        "isLoading": {
+          "type": "boolean",
+          "description": "Whether to display loading widget"
```
evshi, line 820: Let's highlight that this should be an expression/string instead of a static boolean.

assets/schema/ensemble_schema.json
```diff
@@ -808,6 +808,38 @@
...
+        },
+        "useShimmer": {
+          "type": "boolean",
+          "description": "Whether to use a shimmer widget in loading state"
```
evshi, line 824: What is the relation of shimmer widget to loading widget?

lib/widget/loading_container.dart
```diff
@@ -0,0 +1,96 @@
...
+  void initState() {
+    super.initState();
+  }
+
+  @override
+  Widget buildWidget(BuildContext context) {
+    ScopeManager? scopeManager = DataScopeWidget.getScope(context);
+
+    dataWidget = _buildWidget(widget._controller.widget, scopeManager);
+    loadingWidget =
```
evshi, line 71: Don't build the widget until it's needed?

lib/widget/loading_container.dart
```diff
@@ -0,0 +1,96 @@
...
+      widget: getWidget(),
+      boxController: widget._controller,
+      ignoresPadding: true,
+      ignoresDimension:
+          true, // width/height shouldn't be apply in the container
+    );
+  }
+
+  Widget getWidget() {
+    if (widget._controller.isLoading != null && widget._controller.isLoading!) {
```
evshi, line 84: Need to evaluate the expression from context

assets/schema/ensemble_schema.json
vinothvino42: It wraps the loading widget with the Shimmer widget (https://pub.dev/packages/shimmer) if `useShimmer = true`

lib/widget/loading_container.dart
vinothvino42: Done. Can you check again?

lib/widget/loading_container.dart
vusters: so the expression should be automatically evaluated for setters on Invokable widgets. By the time it gets here the value has been evaluated and binding executed.

evshi: Gotcha, where does that happen? @vinothvino42 if you've already tested this with an expression then disregard my comment

lib/widget/box.dart
```diff
@@ -0,0 +1,63 @@
+import 'package:ensemble/framework/scope.dart';
+import 'package:ensemble/framework/widget/widget.dart';
+import 'package:ensemble/widget/helpers/controllers.dart';
+import 'package:ensemble/widget/helpers/widgets.dart';
+import 'package:ensemble_ts_interpreter/invokables/invokable.dart';
+import 'package:flutter/material.dart';
+
+import '../framework/view/page.dart';
+
+class Box extends StatefulWidget
```
evshi, line 10: Is it intended to include box in this PR?

evshi (approved):
LGTM if you've verified the expression behavior

lib/widget/box.dart
vinothvino42: Currently, we don't have any widgets to create empty boxes with styles like backgroundColor, borderRadius, etc. So I created one for it. So we can design our own shimmer layout using the Box widget instead of showing images for the loading state.

lib/widget/loading_container.dart
vusters: @evshi when we build any widget (ViewBuilder), the 2nd step is to evalPropertyAndRegisterBinding() on all the widget's properties. This means by the time the widget's setters are called, the values are already evaluated. The exception is any property name that starts with `on*` (Actions) and `children` and `item-template`. These are evaluated at the time they are being used.

lib/widget/box.dart
vusters: So we've been refrained to add a single-use widgets like this. I modifed it to be more useful to draw multiple shapes here https://github.com/EnsembleUI/ensemble/pull/782

vusters:
@vinothvino42 I merged the code because I need it, and also enhance a few things https://github.com/EnsembleUI/ensemble/pull/782.

PR #765, opened 2023-08-25
Add clear, focus and unfocus methods and few more style properties to ConfirmationInput widget

vinothvino42:
In this PR, I added a few properties and methods for the **ConfirmationInput** widget

**Style Properties:**
1. fieldHeight
2. fieldWidth
3. gap
4. borderRadius
5. borderWidth
6. Added custom type for field

NOTE: 4th and 5th properties will work only if you've `fieldType` as `custom`

**Methods:**
1. clear
2. focus
3. unfocus

### DIFF
--- a/lib/widget/confirmation_input.dart
+++ b/lib/widget/confirmation_input.dart
@@ -39,11 +39,12 @@ class ConfirmationInput extends StatefulWidget
           _controller.enableCursor = Utils.getBool(newValue, fallback: true),
       'length': (newValue) =>
           _controller.length = Utils.getInt(newValue, fallback: 4),
-      'width': (value) => _controller.fieldWidth = Utils.optionalDouble(value),
-      'height': (value) =>
+      'fieldWidth': (value) =>
+          _controller.fieldWidth = Utils.optionalDouble(value),
+      'fieldHeight': (value) =>
           _controller.fieldHeight = Utils.optionalDouble(value),
-      'padding': (value) =>
-          _controller.fieldPadding = Utils.getDouble(value, fallback: 10.0),
+      'gap': (value) =>
+          _controller.fieldGap = Utils.getDouble(value, fallback: 10.0),
       'borderRadius': (value) =>
           _controller.fieldBorderRadius = Utils.getDouble(value, fallback: 2.0),
       'borderWidth': (value) =>
@@ -111,7 +112,7 @@ class ConfirmationInputController extends FormInputController {
   String? inputType;
   double? fieldWidth;
   double? fieldHeight;
-  double? fieldPadding;
+  double? fieldGap;
   double? fieldBorderRadius;
   double? fieldBorderWidth;
   Color? defaultFieldBorderColor;
@@ -170,7 +171,7 @@ class ConfirmationInputState extends framework.WidgetState<ConfirmationInput>
             controller.filledFieldBackgroundColor ?? Colors.transparent,
         filledFieldBorderColor:
             controller.filledFieldBorderColor ?? Colors.transparent,
-        fieldPadding: controller.fieldPadding ?? 10.0,
+        fieldPadding: controller.fieldGap ?? 10.0,
         fieldBorderRadius: controller.fieldBorderRadius ?? 2.0,
         fieldBorderWidth: controller.fieldBorderWidth ?? 2.0,
       ),
--- a/lib/widget/confirmation_input.dart
+++ b/lib/widget/confirmation_input.dart
@@ -102,8 +102,7 @@ class ConfirmationInput extends StatefulWidget
   }
 }
 
-class ConfirmationInputController extends BoxController {
-  TextInputFieldAction? inputFieldAction;
+class ConfirmationInputController extends FormInputController {
   String? text;
   late int length;
   bool? autoComplete;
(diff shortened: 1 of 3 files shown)


lib/widget/confirmation_input.dart
```diff
@@ -39,6 +39,15 @@ class ConfirmationInput extends StatefulWidget
           _controller.enableCursor = Utils.getBool(newValue, fallback: true),
       'length': (newValue) =>
           _controller.length = Utils.getInt(newValue, fallback: 4),
+      'width': (value) => _controller.fieldWidth = Utils.optionalDouble(value),
```
vusters, line 42: name the variables what they are, fieldWidth/fieldHeight, since width/height normally apply to the root widget.

lib/widget/confirmation_input.dart
```diff
@@ -39,6 +39,15 @@ class ConfirmationInput extends StatefulWidget
...
+      'height': (value) =>
+          _controller.fieldHeight = Utils.optionalDouble(value),
+      'padding': (value) =>
```
vusters, line 45: not quite the same as fieldPadding, but use `gap`. We already have that concept.

lib/widget/confirmation_input.dart
```diff
@@ -64,7 +73,11 @@ class ConfirmationInput extends StatefulWidget
 
   @override
   Map<String, Function> methods() {
-    return {};
+    return {
+      'clear': () => controller.inputFieldAction?.clear(),
```
vusters, line 77: you'll be repeating these a lot. Can you do this on the controller itself once?

lib/widget/confirmation_input.dart
```diff
@@ -161,13 +207,30 @@ class ConfirmationInputState extends framework.WidgetState<ConfirmationInput> {
...
+  @override
+  void unfocusInputField() {
+    _otpPinFieldController.currentState?.hasFocus = false;
+    _otpPinFieldController.currentState?.focusNode.unfocus();
+  }
 }
 
 extension FieldTypeOtpValue on String {
   OtpPinFieldDecoration get otpPinField {
     switch (this) {
```
vusters, line 231: why not just have 2 options: underline & outline (default). To make it rounded just set the borderRadius.

vusters:
Small changes and you should be good to go

lib/widget/confirmation_input.dart
vinothvino42: I have done this but later I set it as width and height instead.
Okay, I'll change it back to `fieldWidth` and `fieldHeight`.

lib/widget/confirmation_input.dart
vinothvino42: Okay I'll change it `gap` property

lib/widget/confirmation_input.dart
vinothvino42: Currently, In that package, the borderRadius and borderWidth are used for some fieldType only, it's not for all the fieldTypes.

See all the fieldType's of Border - https://github.com/shivbo96/otp_pin_field/blob/4bff49ca5800b1f7a411ea7f8bd1d8b912bdf877/lib/src/otp_pin_field_state.dart#L245

lib/widget/confirmation_input.dart
vinothvino42: Like this - https://github.com/EnsembleUI/ensemble/pull/765/files#diff-817acf83c5ba49db96f33064a06e666c213c175ddf4e2dbe36795b92aab1281d

But I'm getting method not found exception and couldn't see any getBaseMethods usage in our ensemble. Am I doing wrong? I updated the PR with these changes.

`Error Message: cannot compute statement=CallExpression as no method found for property=clearCode: confirmationInputId.clear()`

vusters (approved):

lib/widget/confirmation_input.dart
evshi: @vusters any idea about the error above?

vusters: Not sure, @vinothvino42 you have to look into your logic. Alternatively, have a InputController (if not already exists) that will do these directly without InputFieldAction.

vinothvino42:
```dart
class InputController extends BoxController {
  InputFieldAction? inputFieldAction;

  @override
  Map<String, Function> getBaseMethods() {
    Map<String, Function> methods = super.getBaseMethods();
    methods.addAll({
      'flutter': () => print("Flutter"),
    });
    return methods;
  }

  @override
  Map<String, Function> getBaseSetters() {
    Map<String, Function> setters = super.getBaseSetters();
    setters.addAll({
      'flutterSet': (value) => print("Value: $value"),
    });
    return setters;
  }
}
```

Here, when I assigned the value to the setter and called the methods. setters are working fine but methods are not calling their methods in the controller.
```yaml
Button:
  label: Flutter
  onTap: |
    //@code
    confirmationInputId.flutterSet = 'Apple Developer';
    confirmationInputId.flutter();
```

**Log:**
```
flutter: Value: Apple Developer
flutter: Line: 4 in YAML and Line: 2 within the code block. Error Message: cannot compute statement=CallExpression as no method found for property=flutterCode: confirmationInputId.flutter()
```

**Exception Screenshot:**
Screenshot 2023-08-31 at 3 41 29 PM (2)

vinothvino42: Revert back to mixin approach

PR #799, opened 2023-09-07
Shimmer animation support for Image and Custom View

vinothvino42:
In this PR, I refactored the loading container widget for Shimmer to support images and custom views with shapes.

See the flow video - https://drive.google.com/file/d/1G2nVZ0s0fP1Ot89BDVkLWVgIOI9o8ty0/view?usp=share_link

### DIFF
--- a/assets/schema/ensemble_schema.json
+++ b/assets/schema/ensemble_schema.json
@@ -486,6 +486,135 @@
         }
       }
     },
+    "ImageCropper-payload": {
+      "type": "object",
+      "required": [
+        "source"
+      ],
+      "properties": {
+        "id": {
+          "type": "string",
+          "description": "A unique identifier for this widget"
+        },
+        "source": {
+          "type": "string",
+          "description": "URL to or asset name of the image. If the URL is used, it is highly recommended that the dimensions is set (either with width/height or other means) to prevent the UI jerkiness while loading."
+        },
+        "onTap": {
+          "$ref": "#/$defs/Action-payload",
+          "description": "Call Ensemble's built-in functions or execute code"
+        },
+        "onCropped": {
+          "$ref": "#/$defs/Action-payload",
+          "description": "Call Ensemble's built-in functions or execute code"
+        },
+        "styles": {
+          "allOf": [
+            {
+              "$ref": "#/$defs/baseStyles"
+            },
+            {
+              "$ref": "#/$defs/boxStyles"
+            },
+            {
+              "type": "object",
+              "properties": {
+                "placeholderColor": {
+                  "$ref": "#/$defs/typeColors",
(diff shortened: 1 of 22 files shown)


lib/widget/loading_container.dart
```diff
@@ -167,3 +170,177 @@ class ListDetailShape extends StatelessWidget {
         ],
       );
 }
+
+const _shimmerGradient = LinearGradient(
```
vusters, line 174: Move these shimmer-related stuff below to its own class.

lib/widget/loading_container.dart
```diff
@@ -92,12 +91,16 @@ class LoadingContainerState extends WidgetState<LoadingContainer> {
             opacity: 1,
             duration: const Duration(milliseconds: 300),
             child: widget._controller.useShimmer == true
-                ? Shimmer.fromColors(
-                    baseColor: const Color(0xFFE0E0E0),
-                    highlightColor: const Color(0xFFF5F5F7).withOpacity(0.5),
-                    child: loadingWidget ??
-                        DefaultLoadingShape(
-                            padding: widget._controller.defaultShimmerPadding))
+                ? CustomShimmer(
```
vusters, line 94: we need to expose the shimmer colors to YAML. Previously i would just expose baseColor and highlightColor - very simple. How do we do this with this new update? a list of colors and stops?

lib/widget/loading_container.dart
vusters: We need to document what the below classes do. Is it maintainable? If not I'd prefer to just use a library.

vinothvino42: Okay, I just built it from the official cookbook of Flutter - https://docs.flutter.dev/cookbook/effects/shimmer-loading.
That package is not working with the image.

lib/widget/loading_container.dart
vinothvino42: Even in the package they used the Gradient with these baseColor and highlightColor - https://github.com/hnvn/flutter_shimmer/blob/b97cfd5ea5fe0067ec1cf07ae585d8bfcb7a0fad/lib/shimmer.dart#L91

We'll pass the color from loading container to our custom shimmer widget in the constructor itself.

assets/schema/ensemble_schema.json
```diff
@@ -823,6 +823,14 @@
           "type": "boolean",
           "description": "a shining animation to designate that the content is loading. The animation can be over a defaults shimmer or the loadingWidget if specified."
         },
+        "baseColor": {
```
vusters, line 826: these 2 styles related shoudl be inside "styles". This is sort of how other widgets works.

vusters (approved):
just fix the last comment and you should be good to merge in

PR #800, opened 2023-09-08
fix(schema): update menu schemas

vinothvino42:
I tested it locally in the IDE and it's working fine. Showing completion for all the menus but in the dev studio ensemble it's not showing the autocomplete. I took ensemble_live build from this branch and tested the web studio, it's not showing the console.

In some places, I'm getting this issue - Matches multiple schemas when only one must validate.yaml-schema: Ensemble(0)

Also another one thing we need to do, we've change the way to write Icon widget right? It's not updated in some of the widgets. We've icon schemas called HasIcon and BaseIcon payload. Both looks like same.

### DIFF
--- a/assets/schema/ensemble_schema.json
+++ b/assets/schema/ensemble_schema.json
@@ -3732,15 +3732,9 @@
             ],
             "properties": {
               "icon": {
-                "type": "string",
-                "description": "Icon name from Material Icons or Font Awesome"
-              },
-              "iconLibrary": {
-                "type": "string",
-                "enum": [
-                  "default",
-                  "fontAwesome"
-                ]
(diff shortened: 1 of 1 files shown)


assets/schema/ensemble_schema.json
```diff
@@ -3765,6 +3759,96 @@
...
+            }   
+          }
+        },
+        "items": {
+          "type": "array",
+          "description": "List of menu items (minimum 2)",
+          "items": {
+            "type": "object",
+            "properties": {
+              "icon": {
```
vusters, line 3795: icon / label / page ... should be on the MenuBase. Specific stuff can be on the BottomNavBar

Issue #798, opened 2023-09-06
Fallback widget for backgroundImage


vusters:
backgroundImage should have a fallback widget when we can't load the images via URL. Make sure to do this for both SVG and non SVG. If straightforward we can do this for asset image as well (though these ideally should never happen as an asset).

Please prioritize Column's backgroundImage, but we should do this everywhere backgroundImage is used.

vusters:
If backgroundImage cannot be loaded, this widget will be displayed in its place instead.

vinothvino42:
@vusters PR - 804

vusters:
@vinothvino42 We need the DecorationImage working. Please work on this first thing.

vinothvino42:
Here's the new PR - https://github.com/EnsembleUI/ensemble/pull/806

vinothvino42:
@vusters The fallback style is working fine. Can you check the following code?

```yaml
View:
  styles:
    useSafeArea: true
    backgroundImage:
      source: https://dummyurl.com
      fallback: CustomFallbackWidget

  body:
    Column:
      styles:
        padding: 20
      children:
        - Text:
            text: Flutter
        - Column:
            styles:
              padding: 20
              height: 150
              width: 150
              backgroundColor: green
              backgroundImage:
                source: https://dummyurl.com
                # source: https://firebasestorage.googleapis.com/v0/b/ensembled-web-studio.appspot.com/o/demo_apps%2Fkitchen-sink%2Fbeach_sunset.jpg?alt=media&token=a197bdb1-7b8f-479a-9fee-722ec7857c97
                fallback: SmallWidget
            children:
              - Text:
                  text: SwiftUI Developer
                  styles:
                    backgroundColor: blue
                  textStyle:
                    color: white

SmallWidget:
  Image:
    styles: { fit: cover, height: 100, width: 100 }
    source: https://akm-img-a-in.tosshub.com/indiatoday/images/media_bank/202309/iphone-15-084541573-16x9_3.png?VersionId=7f2lrBsIJhGzRKO6ZdWIqNTf2Tjcfhz7&size=690:388

CustomFallbackWidget:
  Image:
    styles: { fit: cover, height: 700, width: 300 }
    source: https://phoenixcellular.com/wp-content/uploads/2022/09/overview__bcphzsdb4fpu_og.png

```

vinothvino42:
[...]

PR #806, opened 2023-09-13
Update BoxWrapper to support fallback backgroundImage

vinothvino42:
Ticket: [#798](https://github.com/EnsembleUI/ensemble/issues/798)

Removed image from the BoxDecoration and added the `backgroundImage` as a image widget inside the stack along with the child widget.

### DIFF
--- a/lib/framework/model.dart
+++ b/lib/framework/model.dart
@@ -43,8 +43,10 @@ class BackgroundImage {
       image: imageProvider,
       fit: _fit,
       alignment: _alignment,
-      onError: (_, __) =>
-          fallbackWidget != null ? (_, __, ___) => fallbackWidget : null,
+      onError: (_, __) {
+        // TODO: onError doesn't have errorBuilder to set the fallback widget
+        // return fallbackWidget != null ? (_, __, ___) => fallbackWidget : null;
+      },
     );
   }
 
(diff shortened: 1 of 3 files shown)


lib/framework/model.dart
```diff
@@ -43,8 +43,10 @@ class BackgroundImage {
       image: imageProvider,
       fit: _fit,
       alignment: _alignment,
-      onError: (_, __) =>
-          fallbackWidget != null ? (_, __, ___) => fallbackWidget : null,
+      onError: (_, __) {
```
vusters, line 46: remove the function if it doesn't do anything

lib/widget/helpers/widgets.dart
```diff
@@ -56,48 +56,56 @@ class BoxWrapper extends StatelessWidget {
...
+                          offset: boxController.shadowOffset ??
+                              ThemeManager().getShadowOffset(context),
+                          blurStyle: boxController.shadowStyle ??
+                              ThemeManager().getShadowStyle(context))
+                    ],
+            ),
+      child: backgroundImage != null
+          ? Stack(
+              children: [
+                backgroundImage,
```
vusters, line 103: this will cause the Image to contribute to the Stack's size. We only want the main widget to determine the Stack's size, and the Image to act as the background. Use Position.fill()

vusters (changes requested):

lib/widget/helpers/widgets.dart
vinothvino42: I updated it. Can you review it?

Issue #772, opened 2023-08-28
Support infinite scrolling for GridView


kmahmood74:
Customer requirement - 

1. User sees n number of cards on the dashboard. API returns not only the cards but also the token to get the next n cards
2. As the user scrolls down to the last card, new cards are fetched from the server. Meanwhile, a loading indicator shows up at the bottom of the page
3. New cards are added to the bottom of the list

Not urgent and can come in later if too much work: to avoid too many images in memory, GridView supports pageSize for number of widgets before server is called and also a maxBufferCount that dictates max number of child widgets that are kept in the buffer. As more are added, some are removed from the start. As user scrolls back up, the removed images are fetched from the server

evshi:
Proposal to extend `item-template` for infinite scrolling

```yaml
item-template:
   #API name
   api: 
   # Flexible query string key-value map
   pageParams:
      limit:
      offset:
```

Alternatively add new `onScrollEnd` event

```
GridView:
    ....
    onScrollEnd:
       # invoke API with params as usual, bind to stored data value
       invokeApi:
           ....
```

@vusters @vinothvino42 WDYT?

kmahmood74:
I think the GridView/onScrollEnd makes more sense and is more flexible as I can call any action to handle onScrollEnd. The item-template one is too limiting as it assumes that all I would do is to call an api on page end. 

Besides the above, we need the ability to specify the loading container when user reaches the end, this is critical as different customers would want different types of loading containers. Other stuff like maxItemsInMemory etc can be added later as I wrote in the description.

vinothvino42:
@evshi 
Yeah, The second one looks good. User can have their own control when the ListView or GridView or any scrollable widget reach the end.

@kmahmood74 
```yaml
LoadingContainer:
  id: userGrid
  isLoading: ${getPeople.isLoading}
  useShimmer: true
  loadingWidget:
    OurCustomListViewLoaderWithShimmerAndShapeWidget:
  widget:
    GridView:
      onScrollEnd:
        invokeAPI:
          name: getPeople

API:
  getPeople:
    uri: https://randomuser.me/api/?results=6
    method: GET

```

kmahmood74:
@vinothvino42 I am assuming that in the response of invokeAPI that is called onScrollEnd, user will update the array that is bound to item-template with the new data. Updating that array should automatically show the new items. 

Will that work?

evshi:
[...]

PR #828, opened 2023-09-22
Infinite Scroll for ListView and GridView

snehmehta:
Ticket: #772

EDL Definition
```yaml
    ListView:
      onScrollEnd:
        invokeAPI:
          name: getPeople
          onResponse: |
            //@code

            var results = getPeople.body.results;
            var oldResults = ensemble.storage.users;
            var total = oldResults.concat(results);

            ensemble.storage.users = total;

      item-template:
        data: ${ensemble.storage.users}
        name: user 

        template:
          Text:
            text: ${user.name.first}

```

### DIFF
--- a/assets/schema/ensemble_schema.json
+++ b/assets/schema/ensemble_schema.json
@@ -2067,6 +2067,10 @@
                       "type": "number",
                       "description": "Instead of itemHeight, you can set the tile's dimension as a ratio of (item width / item height). For example, a tile with 3x width and 2x height is 3/2 = 1.5. This attribute will be ignored if itemHeight is set.",
                       "minimum": 0
+                    },
+                    "reverse": {
+                      "type": "boolean",
+                      "description": "Show Grid in reverse order. Default is false"
                     }
                   }
                 }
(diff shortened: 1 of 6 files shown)


sharjeelyunus:
@snehmehta 

Just a little confusion, will `onScrollEnd` work at the bottom of the screen only?
because on the messages screen, we need it at the top of the screen as well i.e. onScrollEnd (at the top of the screen), because messages are in reverse order

vusters:
This restricts the use to storage only and there's potential for errors. Please make sure AppDevs understand these:
- items stay in storage, so when you come back to this page, make sure to re-initialize, othewise you'll just add duplicate data to it.
- how to clean up the storage when you move away from the page.

Ideally we have a proper binding between API and ListView/GridView to automatically aggregate the data, but this is useful today.

Please change the onScrollEnd to:
- onScrollToTop
- onScrollToBottom
These are generic enough to be used for other purposes, plus support the use case Sharjeel mentioned.

kmahmood74:
Hi @snehmehta Vu had comments on your PR, could you please fix this as multiple customers are blocked on this

snehmehta:
Updated EDL

```yaml
    ListView:
      style: {reverse: true} # This will start building the list from the bottom instead top and onScrollEnd works the same.
      onScrollEnd:
        invokeAPI:
          name: getPeople
          onResponse: |
            //@code

            var results = getPeople.body.results;
            var oldResults = ensemble.storage.users;
            var total = oldResults.concat(results);

            ensemble.storage.users = total;

      item-template:
        data: ${ensemble.storage.users}
        name: user 

        template:
          Text:
            text: ${user.name.first}

```

PR #853, opened 2023-10-03
Add ios deep link manager

vinothvino42:


lib/ios_deep_link_manager.dart
```diff
@@ -0,0 +1,40 @@
...
+    });
+  }
+
+  /// navigate to screen if the deep link specifies a screenId param
+  void _navigateToScreen(Uri uri) {
+    String? screenId =
+        (uri.queryParameters['screenId'] ?? uri.queryParameters['screenid'])
+            ?.toString();
+    if (screenId != null) {
+      ScreenController().navigateToScreen(Utils.globalAppKey.currentContext!,
```
vusters, line 36: Please modify both iOS and non-iOS deep link manager to take in "screenName" in addition to screenId

vusters (changes requested):

lib/ios_deep_link_manager.dart
vinothvino42: I added a base class called `DeepLinkNavigator` with the navigateToScreen method. Both `DeepLinkManager` and `IOSDeepLinkManager` inherit the DeepLinkNavigator class

vusters (approved):

PR #860, opened 2023-10-05
Add navigate to view group action

vinothvino42:
```yaml
Button:
  label: Navigate To Cast
  onTap:
    navigateViewGroup:
      viewIndex: 2
```

Video:
https://drive.google.com/file/d/1HLBquNs0E1fHlIjyNnKDNFhegcma2V35/view?usp=share_link

### DIFF
--- a/assets/schema/ensemble_schema.json
+++ b/assets/schema/ensemble_schema.json
@@ -2959,7 +2959,7 @@
           "properties": {
             "navigateViewGroup": {
               "type": "object",
-              "description": "Navigate between the screens in the view group",
+              "description": "Navigate between the screens in the view group. It is applicable only when inside a ViewGroup",
               "required": [
                 "viewIndex"
               ],
--- a/assets/schema/ensemble_schema.json
+++ b/assets/schema/ensemble_schema.json
@@ -2951,6 +2951,28 @@
           }
         },
         {
+          "title": "Navigate View Group",
+          "type": "object",
+          "required": [
+            "navigateViewGroup"
+          ],
+          "properties": {
+            "navigateViewGroup": {
+              "type": "object",
+              "description": "Navigate between the screens in the view group",
+              "required": [
+                "viewIndex"
+              ],
+              "properties": {
+                "viewIndex": {
+                  "type": "integer",
+                  "description": "Enter the index of the screen to navigate"
+                }
+              }
+            }
+          }
+        },
+        {
           "title": "Navigate Modal Screen",
           "type": "object",
           "required": [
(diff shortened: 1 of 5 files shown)


assets/schema/ensemble_schema.json
```diff
@@ -2950,6 +2950,28 @@
...
+        {
+          "title": "Navigate View Group",
+          "type": "object",
+          "required": [
+            "navigateViewGroup"
+          ],
+          "properties": {
+            "navigateViewGroup": {
+              "type": "object",
+              "description": "Navigate between the screens in the view group",
```
vusters, line 2962: add " (applicable only when inside a ViewGroup)"

lib/framework/view/page_group.dart
```diff
@@ -113,6 +126,14 @@ class PageGroupState extends State<PageGroup> with MediaQueryCapability {
 
   @override
   Widget build(BuildContext context) {
+    return ListenableBuilder(
```
vusters, line 129: This will recreate the entire ViewGroup from scratch - we don't want that. We simply want to switch the Tab to another screen. Please make sure we don't rebuild items unnecessary. Optimization is important for our framework as the app could be gigantic.

lib/framework/view/page_group.dart
```diff
@@ -125,16 +146,19 @@ class PageGroupState extends State<PageGroup> with MediaQueryCapability {
         Drawer? drawer = _buildDrawer(context, widget.menu);
         bool atStart = (widget.menu as DrawerMenu).atStart;
         return PageGroupWidget(
+            key: UniqueKey(),
```
vusters, line 149: don't do this. Almost never use UniqueKey().

vusters (changes requested):

vinothvino42:
@vusters Can you again review it? Added UniqueKey only in the EnsembleBottomAppBar widget, it's not rebuilding the EnsembleBottomAppBar when the ListenableBuilder called the builder method with index.

lib/framework/view/bottom_nav_page_group.dart
```diff
@@ -99,6 +112,10 @@ class _BottomNavPageGroupState extends State<BottomNavPageGroup>
       floatingAlignment =
           FloatingAlignment.values.byName(fabMenuItem!.floatingAlignment);
     }
+
+    WidgetsBinding.instance.addPostFrameCallback((_) {
```
vusters, line 116: This is suboptimal. Why let it render, then re-render the correct page? Don't use this callback please.

vusters (changes requested):

vusters (approved):

Issue #671, opened 2023-07-24
Have a clear all for TextInput and PasswordInput


vusters:
A clear all button is a common use case for text inputs. Implement this for TextInput/PasswordInput (see Address widget)

https://discord.com/channels/1031982848485359626/1133099908086370488

vinothvino42:
PR - #808

PR #861, opened 2023-10-06
Fix input formatter issue and add disableClear property

vinothvino42:
Ticket: https://github.com/EnsembleUI/ensemble/issues/671

In this PR, I fixed the text field not updating the next characters when using the Input Mask. Also, I added a property called disableClear, which disable the clear suffix icon button


vusters:
As discussed, have an `enableClearText` which is disabled by default.

vusters (approved):

Issue #864, opened 2023-10-10
Allow adding TextInput Label styling


sibteali786:
## Use Case
- I have a client design requirement to have labels like this inside the the TextInput. 

please add this feature and mention the usage example yaml as well.

![image](https://github.com/EnsembleUI/ensemble/assets/62849614/ab671930-7a3f-4264-ade2-3dac7424f700)

CC: @vinothvino42

vinothvino42:
@vusters PR - #868

vinothvino42:
@sibteali786 You can achieve it something like below once the PR got merged.

```yaml
TextInput:
  id: emailId
  onDelayedKeyDuration: 0
  onDelayedKeyPress: |
    //@code
    emailId.value = emailId.value;
  styles:
    {
      enabled: true,
      labelText: '${emailId.value.length == 0 ? "" : "Name"}',
      hintText: Enter your name,
      enabledBorderColor: grey,
      focusedBorderColor: blue,
    }
```

vinothvino42:
@vusters @sibteali786  New PR - https://github.com/EnsembleUI/ensemble/pull/888

```yaml
TextInput:
  id: emailId
  styles:
    {
      label: Email,
      floatLabel: true,
      variant: box,
      floatingLabelStyle:
        { fontWeight: semibold, fontSize: 34, color: green },
      hintText: Enter your email,
      focusedBorderColor: green,
      enabledBorderColor: gray,
    }
```

PR #868, opened 2023-10-12
Add floating label text and styles

vinothvino42:
Ticket - [#864](https://github.com/EnsembleUI/ensemble/issues/864)

Sample EDL:
```yaml
TextInput:
  id: emailId
  onDelayedKeyDuration: 1
  onDelayedKeyPress: |
    //@code
    emailId.value = emailId.value;
  styles:
    {
      floatLabel: true, # Defaults to False
      label: '${emailId.value.length == 0 ? "" : "Name"}',
      labelStyle:
        {
          fontWeight: semibold,
          fontSize: 34,
          color: green,
        },
      hintText: Enter your name,
    }
```

### DIFF
--- a/assets/schema/ensemble_schema.json
+++ b/assets/schema/ensemble_schema.json
@@ -2861,6 +2861,23 @@
                 "focusedErrorBorderColor": {
                   "$ref": "#/$defs/typeColors",
                   "description": "The border color of this input field when it is receiving focus in its error state. This property can be defined in the theme to apply to all Input widgets."
+                },
+                "labelStyle": {
+                  "allOf": [
+                    {
+                      "$ref": "#/$defs/TextStyle"
+                    },
+                    {
+                      "position": {
+                        "type": "string",
+                        "description": "Set the label text position of the TextInput",
+                        "enum": [
+                          "outline",
+                          "inline"
+                        ]
+                      }
+                    }
+                  ]
                 }
               }
             }
--- a/assets/schema/ensemble_schema.json
+++ b/assets/schema/ensemble_schema.json
@@ -2779,10 +2779,6 @@
           "type": "string",
           "description": "Hint text explaining your widget"
         },
-        "floatingLabelText": {
-          "type": "string",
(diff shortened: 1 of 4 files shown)


assets/schema/ensemble_schema.json
```diff
@@ -2759,6 +2759,10 @@
           "type": "string",
           "description": "Hint text explaining your widget"
         },
+        "floatingLabelText": {
```
vusters, line 2762: Isn't this just the label. Why introduce another variable here?

vinothvino42:
```
label: Email
labelStyle: {
   type: outside | inline
}
```

lib/widget/input/form_textfield.dart
```diff
@@ -199,9 +201,23 @@ class TextInputController extends FormFieldController {
...
   TextStyle? textStyle;
+  LabelStyle? labelStyle;
   TextStyle? hintStyle;
 }
 
+enum LabelPosition { outline, inline }
+
+class LabelStyle extends TextStyle {
+  TextStyle? textStyle;
+  LabelPosition position = LabelPosition.outline;
```
vusters: don't overload TextStyle. From the YAML perspective, the user only understand TextStyle. It will be super confusing If we add a position attribute ONLY for form label and nowhere else.
I think you can use boolean `floatLabel` at the style root.

vusters (changes requested):

lib/widget/input/form_textfield.dart
vinothvino42: Okay vu, I'll change then and make the `floatLabel` defaults to false

```yaml
styles:
  {
    label: '${emailId.value.length == 0 ? "" : "Name"}',
    floatLabel: true, # Defaults to false
    labelStyle:
      {
        fontWeight: semibold,
        fontSize: 34,
        color: green,
      },
  }
```

vinothvino42: @vusters Can you review it again? I updated the PR

vusters (approved):

Issue #609, opened 2023-06-29
Theme's Text theme font color always white unless overriden


vusters:
When define any attribute in theme's, the font color changes to white:
```
Text:
  bodyMedium:
     fontSize: 40    # anything attribute will have the color as white (unless you use color of course).
```
I saw other code hardcode the color to Black. You shouldn't need to do this if we are using the light theme (color shouldn't be white by default on light theme).

snehmehta:
I got to take deep dive to understand it as well. 

But that the case, if we pass null as color default it's taking white color and that is the reason I hardcoded black in default style and similarly assigning fallback style color to custom style. here is [PR](https://github.com/EnsembleUI/ensemble/pull/876) of it.

PR #876, opened 2023-10-17
Custom Text Style default color black.

snehmehta:
Ticket: #609

### DIFF
--- a/lib/util/utils.dart
+++ b/lib/util/utils.dart
@@ -1,6 +1,7 @@
 import 'dart:io';
 import 'dart:math';
 import 'dart:ui';
+import 'package:ensemble/framework/theme/theme_manager.dart';
 import 'package:path/path.dart' as p;
 
 import 'package:ensemble/framework/error_handling.dart';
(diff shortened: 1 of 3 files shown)


lib/util/utils.dart
```diff
@@ -428,7 +428,7 @@ class Utils {
           fontStyle: Utils.optionalBool(style['isItalic']) == true
               ? FontStyle.italic
               : FontStyle.normal,
-          color: Utils.getColor(style['color']),
+          color: Utils.getColor(style['color']) ?? Colors.black,
```
vusters, line 431: move hardcoded to ThemeManager and call it defaultTextColor(). This way when we implement light/dark it'll be easier to change in the theme

vusters (approved):

amin-nas:
@snehmehta we used to be able to set the text color globally using this, but it does not work anymore. Is it related to this change?

Screenshot 2023-10-24 at 7 35 49 AM

Button labels have also gone from default white to black.

snehmehta:
> @snehmehta we used to be able to set the text color globally using this, but it does not work anymore. Is it related to this change?
> 
> Screenshot 2023-10-24 at 7 35 49 AM
> 
> Button labels have also gone from default white to black.

Well, if it's broken, this PR is most likely the reason. Checking.

Issue #871, opened 2023-10-12
Support country search


vusters:
Following https://github.com/EnsembleUI/ensemble/issues/867, please add the ability to search for a country by name, alpha-2, and alpha-3.
```
ensemble.utils.findCountry(searchText);
```
Please come up with a proposal before implementation on the Search mechanism.

kmahmood74:
how is it going @fullstackctp ?

fullstackctp:
> how is it going @fullstackctp ?
It is going ohk. I had a few doubts regarding the functionality that needs to be implemented. I asked sharjeel and vinoth to help me out.

fullstackctp:
https://github.com/EnsembleUI/ensemble/assets/75938927/4db3784d-a2e7-42b0-968b-7c90dcbaec46

kmahmood74:
fixed in #880

PR #880, opened 2023-10-17
Added country search

fullstackctp:
Added the functionality for searching countries #867 and #871


lib/framework/data_context.dart
```diff
@@ -543,6 +546,96 @@ class EnsembleStorage with Invokable {
...
 }
 
+/// singleton handling utils
+class EnsembleUtils with Invokable {
+  static final EnsembleUtils _obj = EnsembleUtils._();
+  EnsembleUtils._();
+  factory EnsembleUtils() => _obj;
+
+  @override
+  Map<String, Function> getters() => {'getCountries': () => allCountries};
```
vusters, line 556: You want to move this to method so you can use `utils.getCountries()`. Getters syntax will be `utils.getCountries` which is not what we want.

lib/framework/data_context.dart
```diff
@@ -543,6 +546,96 @@ class EnsembleStorage with Invokable {
...
+
+  @override
+  Map<String, Function> methods() => {
+        'getCountries': (value) {
```
vusters, line 560: Please review the jiras requirements:
#867 - support `getCountries()` and `getCountry(<alpha-2>)`
#871 - support `findCountry(<search string>)`

lib/framework/data_context.dart
```diff
@@ -543,6 +546,96 @@ class EnsembleStorage with Invokable {
...
+  @override
+  Map<String, Function> setters() => {};
+}
+
+class CountryManager {
+  static final CountryManager _instance = CountryManager._();
+  CountryManager._();
+  factory CountryManager() => _instance;
+
+  List<Map<String, dynamic>> getCountryData(String userInput) {
```
vusters, line 575: Please simply this logic.

vusters (commented):
Use this for search:
- always search for name if there's at least 1 character
- additonally, search for alpha-2 if at most 2 character, and search for alpha-3 if at most 3 character

lib/framework/data_context.dart
```diff
@@ -546,45 +546,52 @@ class EnsembleStorage with Invokable {
...
   Map<String, Function> setters() => {};
-}
 
-class CountryManager {
-  static final CountryManager _instance = CountryManager._();
-  CountryManager._();
-  factory CountryManager() => _instance;
+  List<Map<String, dynamic>> getCountry(String val) {
+    String input = val.toLowerCase().trim();
+    List<Map<String, dynamic>> alpha2List = [];
```
vusters, line 571: We shouldn't have to iterate through the list if the # of characters is not 2 - it's wasteful.

Even better, let's search by alpha-2 if 2 characters, and alpha-3 if 3 characters.

lib/framework/data_context.dart
```diff
@@ -546,45 +546,52 @@ class EnsembleStorage with Invokable {
...
+      countryCode = countryCode.toLowerCase();
+      if (countryCode == input) {
+        alpha2List.add(i);
+      }
+    }
+    return alpha2List;
+  }
 
-  List<Map<String, dynamic>> getCountryData(String userInput) {
+  List<Map<String, dynamic>> findCountry(String userInput) {
```
vusters, line 582: Problems with this function:
- you are making 2 passes through the list. We should only do it once for performance reason.
- By concating the list, you essentially break the existing alphabetical sort order. 

Please use 1 pass only and keep the original JSON sort order.

vusters (changes requested):

lib/framework/data_context.dart
```diff
@@ -569,74 +569,41 @@ class EnsembleUtils with Invokable {
   List<Map<String, dynamic>> getCountry(String val) {
     String input = val.toLowerCase().trim();
     List<Map<String, dynamic>> alpha2List = [];
-    for (var i in allCountries) {
-      String countryCode = i['iso']['alpha-2'];
-      countryCode = countryCode.toLowerCase();
-      if (countryCode == input) {
-        alpha2List.add(i);
+    if (input.length == 2) {
```
vusters, line 572: please search alpha-2 for 2 characters, and alpha-3 for 3 characters.

lib/framework/data_context.dart
```diff
@@ -569,74 +569,41 @@ class EnsembleUtils with Invokable {
...
+      for (var i in allCountries) {
+        String countryCode = i['iso']['alpha-2'];
+        countryCode = countryCode.toLowerCase();
+        if (countryCode == input) {
+          alpha2List.add(i);
+        }
```
vusters, line 578: The value is unique. You need to break out and return once found. There is no reason to loop through the entire list.
The return type should be `Map<String, dynamic>?`. Either you find the single result or you don't.

vusters (changes requested):

vusters (approved):

kmahmood74:
@fullstackctp this is incomplete without a kitchen sink example and documentation. No one knows how to use it. Please provide example and documentation.

PR #883, opened 2023-10-18
Refactor iOS Deeplink Manager

vinothvino42:


lib/ios_deep_link_manager.dart
```diff
@@ -3,7 +3,7 @@ import 'package:flutter/services.dart';
 
 class IOSDeepLinkManager extends DeepLinkNavigator {
   static final IOSDeepLinkManager _instance = IOSDeepLinkManager._internal();
-  static const MethodChannel _platform = MethodChannel('com.tin.mobile');
+  static const MethodChannel _platform = MethodChannel('com.ensemble.platform');
```
vusters: com.ensembleui - that's our domain (not com.ensemble)

vusters (changes requested):
Please review the regular deep_link_manager. That was changed to pass the arguments to the next screen. Please make sure to do the same for iOS-specific

lib/ios_deep_link_manager.dart
vinothvino42: I already updated it to ```com.ensembleui.host.platform```

vinothvino42: Added in the starter project. 
PR - https://github.com/EnsembleUI/starter/pull/33

vusters (approved):

Issue #902, opened 2023-10-24
Screen without body should still render


vusters:
This should just render and empty screen and not show errors:
```
View:
```
Basically a body widget should not be required.

PR #905, opened 2023-10-25
view body changes

fullstackctp:
#902 
Added the ability to render an empty view without providing the body or any widget


lib/page_model.dart
```diff
@@ -142,7 +142,7 @@ class SinglePageModel extends PageModel {
   _processModel(YamlMap docMap) {
     super._processModel(docMap);
 
-    YamlMap viewMap = docMap['View'];
+    YamlMap viewMap = docMap['View'] ?? docMap;
```
kmahmood74, line 145: this doesn't seem right. It would mean that a page without the View is fine as well. Someone could ignore View and put options, menu, styles and footer and it would still try to execute it?

fullstackctp: I have made changes to that now. I have added 2 conditions in this. If (docMap.containsKey('View)) and if(docMap['View'] == null). If there is no view then it will throw error, otherwise it will check the value of view. If the value of view is null, then it will make an empty widget as a root widget, otherwise it will proceed as normal.

vusters:
No need to introduce any empty widget. In page_model.dart simply make rootWidget optional. When rendering page.dart if rootWidgetModel is null then return a SizedBox.shrink().

lib/page_model.dart
```diff
@@ -135,7 +135,7 @@ class SinglePageModel extends PageModel {
 
   Map<String, dynamic>? pageStyles;
   ScreenOptions? screenOptions;
-  late WidgetModel rootWidgetModel;
+  late WidgetModel? rootWidgetModel;
```
vusters, line 138: don't use `late` anymore. It was there when it was required.

lib/page_model.dart
```diff
@@ -144,54 +144,50 @@ class SinglePageModel extends PageModel {
 
     if (docMap.containsKey("View")) {
```
vusters, line 145: This block is useless and confusing. Why set rootWidgetModel at all? Why nested the if, only to execute the logic outside of it. Please refactor

vusters (changes requested):

kmahmood74:
@vusters please review as @fullstackctp made the changes

lib/page_model.dart
```diff
@@ -144,11 +144,7 @@ class SinglePageModel extends PageModel {
...
       if (docMap['View'] == null) {
-        rootWidgetModel = ViewUtil.buildModel(
-            YamlMap.wrap({
-              "Shape": {'height': 0, 'width': 0}
-            }),
-            customViewDefinitions);
+        rootWidgetModel = null;
```
vusters, line 147: It is null already, why set it to null again?

vusters (approved):

Issue #914, opened 2023-10-27
Lottie widget events


amin-nas:
As an app developer, I want to control when Lottie animation plays, and have access to events when it starts, ends, ...

```yaml
Lottie:
  autoPlay: # true | false, default to true
  
  #events
  onStart:
  onPause:
  onComplete:
```

I should be able to control the Lottie using code:

```yaml:
Button:
  onTap: |
    myLottie.start();
    myLottie.pause();
    myLottie.restart();
```

hemish11:
Should we have any callback values to be passed down the line?

amin-nas:
> Should we have any callback values to be passed down the line?

Please share your suggestions for any addition or improvement to the requirements.

hemish11:
I will add a new callback and a new method of onReverse and .reverse()

hemish11:
Ahh also ill be adding a bool like shouldRepeat

amin-nas:
> Ahh also ill be adding a bool like shouldRepeat

That sounds great! Please call it `autoRepeat` and with default value of `true`.

hemish11:
Also here's the yaml that I used for testing:
```yaml
View:
  header:
    title: Lottie
  styles:
    scrollableView: true
  Column:
    styles: { gap: 16, padding: 24 }

    children:
      - Row:
          styles: { gap: 8, crossAxis: center }
          children:
            - Icon:
                icon: warning
                styles: { color: orange }
            - Text:
                text: Lottie widget must have explicit width and height for web (this includes the preview in the editor). You do not need to specify width or height for native apps. If width or height is not specified for webapp (html renderer), it defaults to 250 for both
                styles: { overflow: wrap, expanded: true }
      - Lottie:
          id: MyLottie
          autoPlay: false
          repeat: false
          onForward: |
            //@code
            console.log("Forward");
          onReverse: |
            //@code
            console.log("Reverse");
          onPause: |
            //@code
            console.log("Pause");
          onComplete: |
            //@code
            console.log("Completed");
          source: https://assets6.lottiefiles.com/private_files/lf30_rnpgzd17.json
          styles:
            width: 400
            height: 100
      - Lottie:
          source: https://assets3.lottiefiles.com/packages/lf20_lzpnnin5.json
          repeat: true
          styles:
            width: 200
            height: 200
      - Button:
          styles:
            outline: true
            borderColor: 0xffed5742
            labelStyle:
              color: 0xffed5742
          label: Forward
          onTap: |
[...]

PR #915, opened 2023-10-28
LOTTIE - Refactored code, added support for methods and callbacks

hemish11:
Things done in this PR:
- Refactored some code
- Added support for forward(), reverse(), reset() and stop() methods to the Lottie widget.
- Added onForward, onReverse, onComplete, onStop callbacks to the Lottie widget.
- Added autoPlay parameter.

### DIFF
--- a/lib/widget/lottie/lottie.dart
+++ b/lib/widget/lottie/lottie.dart
@@ -32,12 +32,14 @@ class EnsembleLottie extends StatefulWidget
   Map<String, Function> methods() {
     if (kIsWeb) {
       // A little hacky way to check if html renderer is used as only html render would have the lottieController null.
-      final bool isHtml = _controller.lottieController != null;
+      // lottieController is null only for the html renderer.
+      // Cannot use js.context['flutterCanvasKit'] != null to check the html renderer as it requires importing dart:js which will break app in mobile runtime as dart:js is a web only package
+      final bool isNotHtml = _controller.lottieController != null;
 
       return {
         // Method to start animation in forward direction
         'forward': () {
-          if (isHtml) {
+          if (isNotHtml) {
             if (_controller.repeat) {
               _controller.lottieController!.repeat();
             } else {
@@ -49,7 +51,7 @@ class EnsembleLottie extends StatefulWidget
         },
         // Method to run animation in reverse direction
         'reverse': () {
-          if (isHtml) {
+          if (isNotHtml) {
             _controller.lottieController!.reverse();
           } else {
             _controller.lottieAction!.reverse();
@@ -57,7 +59,7 @@ class EnsembleLottie extends StatefulWidget
         },
         // Method to reset animation to initial position
         'reset': () {
-          if (isHtml) {
+          if (isNotHtml) {
             _controller.lottieController!.reset();
           } else {
             _controller.lottieAction!.reset();
@@ -65,7 +67,7 @@ class EnsembleLottie extends StatefulWidget
         },
         // Method to stop animation at current position
         'stop': () {
-          if (isHtml) {
+          if (isNotHtml) {
             _controller.lottieController!.stop();
           } else {
             _controller.lottieAction!.stop();
@@ -139,16 +141,20 @@ class LottieController extends BoxController {
   bool repeat = true;
   bool autoPlay = true;
 
+  // lottieController and lottieAction are different things.
+  // lottieController is a AnimationController which is used to control animation and hook callbacks for all the platforms except web html renderer
+  // lottieAction is a mixin that is used to define all the methods for the html renderer. Cannot use normal AnimationController as html is rendered using iframe and doesn't use Lottie widget
   AnimationController? lottieController;
+  LottieAction? lottieAction;
 
   EnsembleAction? onForward;
   EnsembleAction? onReverse;
   EnsembleAction? onComplete;
   EnsembleAction? onStop;
 
-  LottieAction? lottieAction;
-
+  // method to initialize the AnimationController lottieController
   void initializeLottieController(LottieComposition composition) {
+    // Setting the duration of the animation once the lottie is loaded
     lottieController!.duration = composition.duration;
 
     if (autoPlay) {
@@ -160,6 +166,7 @@ class LottieController extends BoxController {
     }
   }
 
+  // Method to link statusListener with their respective events.
   void addStatusListener(BuildContext context, EnsembleLottie widget) {
     final animationStatusActionMap = <AnimationStatus, EnsembleAction?>{
       AnimationStatus.forward: onForward,
--- a/lib/widget/lottie/lottie.dart
+++ b/lib/widget/lottie/lottie.dart
@@ -27,6 +27,7 @@ class EnsembleLottie extends StatefulWidget
   @override
   Map<String, Function> methods() {
     return {
+      // Method to start animation in forward direction
       'forward': () {
         if (_controller.repeat) {
           _controller.lottieController!.repeat();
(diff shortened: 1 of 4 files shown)


kmahmood74:
@hemish11 always mention the url to the ticket or ticket number with the pound sign in all your PRs. Without it we won't know which ticket you are resolving. 

Also the first thing in your commit should be the URL to the ticket. Github auto-links it with the ticket when you do that. 

for this one, it's https://github.com/EnsembleUI/ensemble/issues/914

lib/widget/lottie/lottie.dart
```diff
@@ -28,7 +26,22 @@ class EnsembleLottie extends StatefulWidget
...
   @override
   Map<String, Function> methods() {
-    return {};
+    return {
+      // Method to start animation in forward direction
+      'forward': () {
+        if (_controller.repeat) {
+          _controller.lottieController!.repeat();
+        } else {
+          _controller.lottieController!.repeat();
```
kmahmood74, line 35: in both cases you are calling repeat, what's the point of else?

lib/widget/lottie/native/lottiestate.dart
```diff
@@ -8,65 +9,139 @@ import 'package:ensemble/widget/widget_util.dart';
...
+      AnimationStatus.forward: controller.onForward,
+      AnimationStatus.reverse: controller.onReverse,
+      AnimationStatus.dismissed: controller.onStop,
+      AnimationStatus.completed: controller.onComplete,
+    };
+
+    controller.lottieController!.addStatusListener(
+      (status) {
+        if (animationStatusActionMap[status] != null) {
+          ScreenController().executeAction(
```
kmahmood74, line 31: this code that executes the action should be inside LottieController. That's what the controller is for. Also that way you can re-use most of the code between web and native

lib/widget/lottie/web/lottiestate.dart
```diff
@@ -14,115 +18,201 @@ import 'dart:js' as js;
...
+    final animationStatusActionMap = <AnimationStatus, EnsembleAction?>{
+      AnimationStatus.forward: controller.onForward,
+      AnimationStatus.reverse: controller.onReverse,
+      AnimationStatus.dismissed: controller.onStop,
+      AnimationStatus.completed: controller.onComplete,
+    };
+
+    controller.lottieController!.addStatusListener(
+      (status) {
+        if (animationStatusActionMap[status] != null) {
```
kmahmood74, line 41: this should be in LottieController, see my previous comment in the native LottieState

lib/widget/lottie/web/lottiestate.dart
```diff
@@ -14,115 +18,201 @@ import 'dart:js' as js;
...
-    double height = widget.controller.height?.toDouble() ?? 250;
-    String _id = widget.controller.id ?? id;
-    String repeat = (widget.controller.repeat == null)
-        ? 'true'
-        : widget.controller.repeat!.toString();
-    String tag = '<div id="$_id"></div>';
-    if (widget.controller.width == null || widget.controller.height == null) {
-      print(
-          'Lottie widget must have explicit width and height for html renderer in the browser (this includes the preview in the editor). You do not need to specify width or height for native apps or for canvaskit renderer. Defaulting to $width for width and $height for height');
+  Widget buildLottieHtml(BoxFit? fit, LottieController controller) {
```
kmahmood74, line 87: widget.controller is available everywhere inside the state. Why then create a variable called controller and then pass it around. 

Most of these changes are unnecessary and make it very hard to actually review the change. Please focus only on the actual change. Remove these unnecessary changes and update your PR

lib/widget/lottie/web/lottiestate.dart
```diff
@@ -14,115 +18,201 @@ import 'dart:js' as js;
...
-            String script =
-                'bodymovin.loadAnimation({container: document.getElementById("$_id"),renderer: "svg",loop: $repeat,autoplay: true,path: "$c"});';
-            if (widget.controller.onTap != null) {
-              script +=
-                  'document.getElementById("$_id").addEventListener("click",() => handleMessage("$_id",""));';
-            }
-            return script;
-          },
-          size: Size(width, height));
+        id: id,
```
kmahmood74, line 106: none of these changes are needed, please cleanup and minimize to only the changes that are required

lib/widget/lottie/web/lottiestate.dart
```diff
@@ -14,115 +18,201 @@ import 'dart:js' as js;
...
+      );
     }
-    return blankPlaceholder();
+
+    return blankPlaceholder(controller);
   }
 
-  Widget buildLottieCanvas(BoxFit? fit) {
-    String source = widget.controller.source.trim();
+  Widget buildLottieCanvas(BoxFit? fit, LottieController controller) {
```
kmahmood74, line 136: same issue, no reason to pass controller when it is available as widget.controller

lib/widget/lottie/web/lottiestate.dart
```diff
@@ -14,115 +18,201 @@ import 'dart:js' as js;
...
+        },
+      );
     }
-    return blankPlaceholder();
+
+    return blankPlaceholder(controller);
   }
 
-  Widget placeholderImage() {
+  Widget placeholderImage(LottieController controller) {
```
kmahmood74, line 186: once again, same issue, cleanup please

kmahmood74 (changes requested):

hemish11:
@kmahmood74 Please review this now

kmahmood74 (approved):
looks good, make sure to test it as follows - 

1. Web - test it with canvasKit AND with html renderer. Compile it with htmlrenderer option and test as well
2. iOS native
3. Android native

hemish11:
@kmahmood74 @snehmehta Finally ready for review!

lib/widget/lottie/lottie.dart
```diff
@@ -28,7 +30,66 @@ class EnsembleLottie extends StatefulWidget
 
   @override
   Map<String, Function> methods() {
-    return {};
+    if (kIsWeb) {
+      // A little hacky way to check if html renderer is used as only html render would have the lottieController null.
+      final bool isHtml = _controller.lottieController != null;
```
kmahmood74, line 35: yes this is a really hacky way. Also I see that you use _controller.lottieController! everywhere.

kmahmood74:
@hemish11 I added some comments

lib/widget/lottie/lottie.dart
```diff
@@ -28,7 +30,66 @@ class EnsembleLottie extends StatefulWidget
...
+          } else {
+            _controller.lottieAction!.stop();
+          }
+        },
+      };
+    }
+
+    return {
+      // Method to start animation in forward direction
+      'forward': () {
```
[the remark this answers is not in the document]
hemish11: There is a difference, lottieController is a normal AnimationController and is passed to lottie widget directly and thus controls animation whereas lottieAction has its methods which are coded specifically for the HTML renderer. I had to do so as HTML render does all the things by running javascript code and thus had to segregate all the things.

lib/widget/lottie/lottie.dart
kmahmood74: @hemish11 this is just super hacky, please make it cleaner so anyone working on the code later can make sense of it

hemish11: The simpler way that I first tried to do was to check by this condition: `js.context['flutterCanvasKit'] != null;` but for this, import `'dart:js'` as js; needs to be imported in the main lottie.dart file which is not possible as the app starts to crash when importing the web-only library when it runs on mobile. It has been imported in web/lottiestate.dart but as it only runs when it is in web there is no issue with that import. Also currently I'm adding more comments in the code so that its easy to understand.

hemish11:
@kmahmood74 Added a few comments in code

kmahmood74 (approved):

Issue #847, opened 2023-09-29
Evaluate expressions passed header, footer, and view's styles


amin-nas:
I want to be able to set properties dynamically across:

1. View style props (e.g. backgroundColor, scrollableView)
2. Header props
3. Footer props

```yaml
View:
  styles:
    backgroundColor: '${flag == true ? "green" : "red"}'
  header:
    title: Home
    styles:
      backgroundColor: '${flag == true ? "green" : "red"}'
  body:
    Column:
      styles:
        backgroundColor: '${flag == true ? "green" : "red"}'
      children:
        - Content
  footer:
    styles:
      backgroundColor: '${flag == true ? "green" : "red"}'
    children:
      - Content

Global: |-
  //@code
  var flag = true;

Content:
  body:
    Spacer:
      styles:
        size: 100
```

kmahmood74:
this doesn't work for title as well. For header/title, the expression is evaluated only once but never updated. The whole header expression code is unfinished or just super buggy. 

@vusters @vinothvino42 

Example - 

```
 header:
    
    title: Welcome ${ensemble.storage.players[ensemble.storage.currentPlayer].name}
```

will work only once. Afterwards even if I change the currentIndex or players array and assign, it won't update. Same code will update fine if it was on a Text instead of title. Super buggy!

kmahmood74:
appreciate if we start fixing these things

kmahmood74:
see the use case in #763 as well

hemish11:
Yes, https://github.com/EnsembleUI/ensemble/issues/763 is a duplicate of this, and for column and other widgets, this does work, it's only breaking for the View.

kmahmood74:
@hemish11 this is not right. You missed the styles under View even though both #763 and #847 mentioned View specifically. I see this code - 

```
    LinearGradient? backgroundGradient = Utils.getBackgroundGradient(
        widget._pageModel.pageStyles?['backgroundGradient']);
    Color? backgroundColor =
        Utils.getColor(widget._pageModel.pageStyles?['backgroundColor']);
        ```
in Page.dart so it is clear that backgroundColor or gradient are not being evaluated. 

Please don't just pick and choose and do this for ALL STYLES on View.

kmahmood74:
@hemish11 please do this at your earliest.

kmahmood74:
@hemish11 I checked in the fix for the backgroundColor on View. Please evaluate expressions for other properties as well

Issue #763, opened 2023-08-24
Need support for changing styles of whole View conditionally


anserwaseem:
need to change the background color of whole View based on a condition like this:
```yaml
View:
  styles:
    backgroundColor: "${ensemble.storage.isAnyEmailEnabled == false ? yellow : blue}" 
```

anserwaseem:
Need this for header styles as well

kmahmood74:
see #847 it's related and might be a duplicate

PR #921, opened 2023-10-31
Support for expressions in header and footer

hemish11:
Fixes https://github.com/EnsembleUI/ensemble/issues/847 and https://github.com/EnsembleUI/ensemble/issues/763

### DIFF
--- a/.android/Flutter/build.gradle
+++ b/.android/Flutter/build.gradle
@@ -30,6 +30,11 @@ group 'com.ensembleui.ensemble'
 version '1.0'
 
 android {
+    // Conditional for compatibility with AGP <4.2.
+    if (project.android.hasProperty("namespace")) {
+        namespace 'com.ensembleui.ensemble'
+    }
+
     compileSdkVersion flutter.compileSdkVersion
     ndkVersion flutter.ndkVersion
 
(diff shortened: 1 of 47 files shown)


snehmehta (commented):
The ticket expects to eval pagestyle as well, have you check it's evaluating or not?

[PageModel - pageStyles](https://github.com/hemish11/ensemble/blob/d66bbf6ff47dc98f51fe3c31ff29a47b57ce5f5f/lib/page_model.dart#L136-L137)

hemish11:
@snehmehta Yes, it does evaluate the page style.

vusters (approved):
Sufficient to unblock folks for now. The proper solution is not to add eval() everywhere but to convert the page into a widget so evaluation and bindings is automatic (just like a normal widget)

kmahmood74:
@hemish11 this is not right. You missed the styles under View even though both #763 and #847 mentioned View specifically. I see this code - 

```
    LinearGradient? backgroundGradient = Utils.getBackgroundGradient(
        widget._pageModel.pageStyles?['backgroundGradient']);
    Color? backgroundColor =
        Utils.getColor(widget._pageModel.pageStyles?['backgroundColor']);
        ```
in Page.dart so it is clear that backgroundColor or gradient are not being evaluated. 

Please don't just pick and choose and do this for ALL STYLES on View.

Issue #688, opened 2023-07-31
Need Datagrid sorting for nested data


anserwaseem:
Need Datagrid sorting for nested data. Currently flat data is supported but most of the time server returns nested data.

![image](https://github.com/EnsembleUI/ensemble/assets/32386713/2df46e00-1b6d-4de2-a975-027298595849)

vusters:
@vinothvino42 consider this as the highest priority please.

vinothvino42:
In Progress PR - #922

vinothvino42:
@vusters Again I updated it - https://github.com/EnsembleUI/ensemble/pull/922#issuecomment-1811855515

PR #922, opened 2023-10-31
Add nested sort functionality for DataGrid

vinothvino42:
Ticket: [#688](https://github.com/EnsembleUI/ensemble/issues/688)

### DIFF
--- a/lib/layout/data_grid.dart
+++ b/lib/layout/data_grid.dart
@@ -1,3 +1,5 @@
+import 'dart:ffi';
+
 import 'package:ensemble/framework/action.dart';
 import 'package:ensemble/framework/data_context.dart';
 import 'package:ensemble/framework/error_handling.dart';
@@ -16,6 +18,7 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:ensemble_ts_interpreter/invokables/invokable.dart';
 import 'package:ensemble/framework/view/page.dart';
+import 'package:json_path/json_path.dart';
 
 enum DataColumnSortType { ascending, descending }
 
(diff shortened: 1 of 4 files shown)


lib/layout/data_grid.dart
```diff
@@ -474,6 +474,54 @@ class DataGridState extends WidgetState<DataGrid>
     setState(() {});
   }
 
+  void customSort(List dataList, String? sortKey, bool isAscendingOrder) {
```
vusters, line 477: extract this to a utility and add unit tests. This is not UI logic and easily testable.

lib/layout/data_grid.dart
```diff
@@ -474,6 +474,54 @@ class DataGridState extends WidgetState<DataGrid>
...
+    if (sortKey == null) return;
+    final lists = JsonPath(sortKey).read(dataList).toList();
+    if (lists.length >= 2) {
+      dataList.asMap().forEach((index, value) {
```
vusters, line 481: does this even work? You loop through each item in the list then sort the entire grid? Please verify correctness throroughly with unit tests. 
Either way this logic is off. Please refactor

vusters (changes requested):

kmahmood74:
@vinothvino42 this is urgently needed by a customer, please look at @vusters 's comments and update quickly. 
cc @anserwaseem

lib/layout/data_grid.dart
vinothvino42: @vusters  Something is wrong here, looks like my code is missing. Not sure, I'll look check and get back to you

kmahmood74:
is it in or not in? @vusters KPN is waiting. @vinothvino42 you said your code was missing but then linked a video that shows its working? Can you please be clear here!

lib/layout/data_grid.dart
vinothvino42: Done

vinothvino42:
@kmahmood74 @vusters Unfortunately my code was not pushed on Nov 2, I did it and then only I took a screen record and re-requested vu to review it and also now my local changes are missing it because I used to replace the ensemble for some reason.

It's done now. Can you review it again?

- Updated the DataGrid widget for sorting
- Moved the data grid sort logic to utils
- Wrote test cases for sorting the list with nested map objects using `key` and `order`

vusters (approved):
THanks for making the change.

Issue #595, opened 2023-06-21
Sound Effects/Haptics


RuthLHanan:
Adding sound effects and other haptics that involve playing with your sense of touch

hemish11:
For the yaml part, I would say that we create a wrapper widget like Haptic and only pass the id to it, and maybe one or two parameters as in haptic type (for iOS, not sure for that tho), and then the user would be able to control haptic using the methods for the given id. How does that sound?

Here's the exact yaml:
```yaml
View:
  header:
    title: Haptic
  Haptic:
    id: MyHaptic
    child:
      Column:
        styles: { gap: 16, padding: 24 }
        children:
          - Button:
              label: Short Haptic
              onTap: |
                MyHaptic.short();
          - Button:
              label: Long Haptic
              onTap: |
                MyHaptic.long();
          - Button:
              label: Double Haptic
              onTap: |
                MyHaptic.double();
```

kmahmood74:
@hemish11 I think it's better for Haptic to be an action so devs can invoke from anywhere. 

Something like - triggerHaptic
and it has properties like short, long etc. 

Actions are the most powerful feature of Ensemble.

hemish11:
Agreed, it would be more beneficial that way.

PR #935, opened 2023-11-02
Added actions for haptic

hemish11:
Closes https://github.com/EnsembleUI/ensemble/issues/595

The methods supported for haptic are:
 - heavyImpactHaptic
 - mediumImpactHaptic
 - lightImpactHaptic
 - selectionClickHaptic
 - vibrateHaptic


snehmehta (changes requested):
[DataContext - Methods](https://github.com/EnsembleUI/ensemble/blob/517dd59eabc98fc1e1cc19b455ced8f56703f890/lib/framework/data_context.dart#L333)

Add action here too, so haptics can be able to call from Js as well.

hemish11:
@snehmehta Added those changes, please re-review it once

snehmehta:
@hemish11 provide sample EDL (yaml) to check the functionality

hemish11:
Heres the EDL:
```yaml
View:
  header:
    title: Testing
  body:
    Column:
      children:
        - Button:
            label: Heavy Impact
            onTap:
              haptic:
                type: heavyImpact
        - Button:
            label: Medium Impact
            onTap:
              haptic:
                type: mediumImpact
        - Button:
            label: Light Impact
            onTap:
              haptic:
                type: lightImpactHaptic
        - Button:
            label: Selection Click
            onTap:
              haptic:
                type: selectionClickHaptic
        - Button:
            label: Vibrate
            onTap:
              haptic:
                type: vibrateHaptic
```

vusters:
Not sure I understand. The haptic is not the primary Action right, as the user most likely want to do something else (while adding the haptic). So in a way it's more like a feature of onTap action i.e. onTapHaptic = heavy | light | .. ?

hemish11:
@vusters I've also added support to run it from js code so I guess the user would be able to use it as a secondary action as well. I didn't understand what on wanted to say on this part `onTapHaptic = heavy | light`

vusters:
What's the YAML looks like for onTap of a button to showing a Toast with a heavy haptic feedback?

hemish11:
@vusters For that it can be something like:
```yaml
Button:
  label: MyButton
  onTap: |-
    //@code
    var toastPayload = {
      message: 'Hello World',
      options: {
        duration: 2,
        alignment: 'bottomLeft'
      }
    };
    var hapticPayload = {
      type: 'heavyImpact'
    };

    ensemble.showToast(toastPayload);
    ensemble.haptic(hapticPayload);
```

kmahmood74:
> Not sure I understand. The haptic is not the primary Action right, as the user most likely want to do something else (while adding the haptic). So in a way it's more like a feature of onTap action i.e. onTapHaptic = heavy | light | .. ?

@vusters I don't agree with that. Haptic should be a primary action. You could be triggering haptics for a variety of reasons - a message came in inbox and want the phone to vibrate, someone liked your post. post a swipe once API has responded and so on and so forth. We must allow Haptics to be independent actions that can be invoked from anywhere. 

also there need to be an onComplete on Haptics as that would allow devs to trigger another action if needed. Unfortunately we don't support parallel actions which would have made a lot of sense in this case. 

Supporting a haptic feedback onTap, if there, should be in addition to an independent haptic action

kmahmood74:
> @vusters For that it can be something like:
> 
> ```yaml
> Button:
>   label: MyButton
>   onTap: |-
>     //@code
>     var toastPayload = {
>       message: 'Hello World',
>       options: {
>         duration: 2,
>         alignment: 'bottomLeft'
>       }
>     };
>     var hapticPayload = {
>       type: 'heavyImpact'
>     };
> 
>     ensemble.showToast(toastPayload);
>     ensemble.haptic(hapticPayload);
> ```

@hemish11 haptic is a noun and not a verb. Call it - ensemble.invokeHaptic

kmahmood74:
@hemish11 please add a onComplete on invokeHaptic action as well. This will allow devs to attach a callback. Callback could be any action. This allows for chaining actions

lib/action/haptic_action.dart
```diff
@@ -17,6 +18,9 @@ class HapticAction extends EnsembleAction {
 
   final HapticTypes type;
 
+  factory HapticAction.fromMap(dynamic inputs) =>
+      HapticAction.fromYaml(payload: Utils.getYamlMap(inputs));
+
   factory HapticAction.fromYaml({Map? payload}) {
     if (payload == null || payload['type'] == null) {
       throw LanguageError("${ActionType.share.name} requires 'type'");
```
vusters, line 26: careful with cut and paste.

lib/action/haptic_action.dart
```diff
@@ -17,6 +18,9 @@ class HapticAction extends EnsembleAction {
 
   final HapticTypes type;
 
+  factory HapticAction.fromMap(dynamic inputs) =>
```
vusters, line 21: combine fromYaml() with this function. takes in dynamic and check if it's a simple `Map` which can be object or yaml.

lib/action/haptic_action.dart
```diff
@@ -17,6 +18,9 @@ class HapticAction extends EnsembleAction {
 
   final HapticTypes type;
```
vusters, line 19: singular HapticType. We are describing a single haptic type.

lib/action/haptic_action.dart
vusters: The compiler doesn't complain about `return HapticAction(hapticType);`? hapticType can be null but you declare it as non-null.

The hapticType can be an expression, meaning you need to evaluate its value when executing it (not at creation time). See how the rest of the Action work. e.g. scopeManager.eval(hapticType)

vusters (commented):
Please do this:
1. HapticAction requires an onComplete. You need to be able to execute other Actions along side (after) the haptic.
2. add `onTapHaptic` to places that we support onTap. The values for onTapHaptic should be the HapticType enum (heavyImpact | mediumImpact | ...)
3. my opinion: call it triggerHaptic. invokeHaptic sounds too heavy duty (we used it for invokeAPI).

kmahmood74:
@vusters I intentionally picked invokeHaptic instead of triggerHaptic. We can say the same thing about triggerAPI. We have to be consistent. Users learn systems that have a consistent naming scheme. If we are using invoke one place, we must continue to use invoke in other places as well. 

So please stick to invokeHaptic

agree with the rest

kmahmood74:
@hemish11 you might have missed @vusters 's comment on adding onTapHaptic wherever onTap is used. See above. That's a short hand as the most common use of haptics is onTap

hemish11:
@kmahmood74 Yes, the code for all is the same for all, but still, I tested all of them on both Android and ios on a physical device

hemish11:
Also @snehmehta and @vusters can you please review it once and approve it if all looks good?

kmahmood74:
all good from my side

kmahmood74 (approved):

snehmehta (approved):

Issue #596, opened 2023-06-22
Ignore invalid API attributes


kmahmood74:
it used to be the case that API attributes that did not make sense were ignored but now the API fails silently. This caused an app to break. 

To test - 

1. create any API
and add

authorization: none

as an attribute to the API's yaml. 

API won't be called and there won't be any error in the console.

kmahmood74:
this is still an issue, should be a quick fix

PR #938, opened 2023-11-03
Added checks for the api parameters

hemish11:
Closes https://github.com/EnsembleUI/ensemble/issues/596

### DIFF
--- a/lib/action/invoke_api_action.dart
+++ b/lib/action/invoke_api_action.dart
@@ -117,7 +117,7 @@ class InvokeAPIController {
         errorResponse = response;
       } catch (error) {
         errorResponse = error;
-        throw error.toString();
+        debugPrint(error.toString());
       }
       _onAPIError(
           context, action, apiDefinition, errorResponse, apiMap, scopeManager);
--- a/lib/action/invoke_api_action.dart
+++ b/lib/action/invoke_api_action.dart
@@ -117,6 +117,7 @@ class InvokeAPIController {
         errorResponse = response;
       } catch (error) {
         errorResponse = error;
+        throw error.toString();
       }
       _onAPIError(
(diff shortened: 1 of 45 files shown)


lib/action/invoke_api_action.dart
```diff
@@ -117,6 +117,7 @@ class InvokeAPIController {
         errorResponse = response;
       } catch (error) {
         errorResponse = error;
+        throw error.toString();
```
vusters, line 120: don't throw the error - this will stop the app flow. THe intention is for the onApiError to execute and notifies the user.

lib/util/http_utils.dart
```diff
@@ -24,28 +24,31 @@ class HttpUtils {
...
-    String? scope = Utils.optionalString(api['authorization']?['scope']);
-    bool forceNewTokens =
-        Utils.getBool(api['authorization']?['forceNewTokens'], fallback: false);
-    if (oAuthService != null && scope != null) {
-      OAuthServiceToken? token = await GetIt.instance<OAuthController>()
-          .authorize(context, oAuthService,
-              scope: scope, forceNewTokens: forceNewTokens);
-      if (token != null) {
-        headers['Authorization'] = 'Bearer ${token.accessToken}';
+    if (api['authorization'] != "none") {
```
vusters, line 27: I don't see anything below that will need this. Is it inside one of the classes used?

vusters (changes requested):

lib/util/http_utils.dart
hemish11: @vusters One part is for the bearer token, one for oauth2 and the final part is for parsing the header and body, and so on.

vusters (approved):

PR #940, opened 2023-11-03
calendar: introducing row span limit and default builder

snehmehta:


snehmehta:
Merging this PR as it's a shell of actual logic which can be found at [snehmehta/table_calendar](https://github.com/snehmehta/table_calendar/commit/fedfcf065479cb1e0522614b2729117bb6de27b1)

PR #950, opened 2023-11-06
Added currentIndex, selectedIndex and indicatorMaxCount properties for Carousel Widget

vinothvino42:
In this PR, I've fixed a few issues and updated the schema

**1. Index Updates:** Ticket: #949 and #955

1. `currentIndex`
    - For setting the initial index of the carousel which starts the carousel from that child of index.
    - To get the current index which is displayed in the carousel
    - Default (0)
2. `selectedIndex` 
    - To get the selectedIndex only when the user clicks on the carousel item otherwise it'll return -1. Default (-1)
3. `indicatorMaxCount` 
    - This property shows a limited number (for ex: 4) of **carousel indicator** when creating the carousel with more (for ex: 100+) items. 
    - Default (null), if this property is null the indicator works normally how it works before (based on the currentIndex property)

**2. Indicator Alignment and Padding** - Ticket: [#952](https://github.com/EnsembleUI/ensemble/issues/952)
- Here, I added multiple alignment (topLeft, bottomCenter, etc) support for the carousel indicator and replaced the `indicatorOffset` with the `indicatorPadding` property

### DIFF
--- a/assets/schema/ensemble_schema.json
+++ b/assets/schema/ensemble_schema.json
@@ -2812,8 +2812,15 @@
                       "type": "string",
                       "description": "Where to display the indicator if specified",
                       "enum": [
-                        "bottom",
-                        "top"
+                        "topLeft",
+                        "topCenter",
+                        "topRight",
+                        "centerLeft",
+                        "center",
+                        "centerRight",
+                        "bottomLeft",
+                        "bottomCenter",
+                        "bottomRight"
                       ]
                     },
                     "indicatorWidth": {
@@ -2858,6 +2865,10 @@
                         }
                       ],
                       "description": "The margin around each indicator"
+                    },
+                    "indicatorPadding": {
+                      "$ref": "#/$defs/Padding-payload",
+                      "description": "The padding around the indicator"
                     }
                   }
                 }
--- a/assets/schema/ensemble_schema.json
+++ b/assets/schema/ensemble_schema.json
@@ -2819,9 +2819,17 @@
                       "type": "boolean",
                       "description": "Determines if carousel should loop infinitely or be limited to item length. Default (false)"
                     },
-                    "selectedItemIndex": {
+                    "selectedIndex": {
                       "type": "integer",
-                      "description": "The initial page to show when first creating the Carousel. Default (0)"
+                      "description": "It returns the selected item index of the Carousel. Default (-1) if not tapped / selected"
+                    },
+                    "currentIndex": {
+                      "type": "integer",
+                      "description": "The initial page to show when first creating the Carousel and the current index of the displaying Carousel Item. Default (0)"
+                    },
+                    "indicatorMaxCount": {
+                      "type": "integer",
+                      "description": "This property is used to show a limited number (for ex: 4) of carousel indicator when creating the carousel with more (for ex: 100+) items. Default (null)"
                     },
                     "autoplayInterval": {
                       "type": "integer",
--- a/assets/schema/ensemble_schema.json
+++ b/assets/schema/ensemble_schema.json
@@ -2809,19 +2809,8 @@
                       ]
                     },
                     "indicatorPosition": {
-                      "type": "string",
-                      "description": "Where to display the indicator if specified",
-                      "enum": [
-                        "topLeft",
-                        "topCenter",
-                        "topRight",
-                        "centerLeft",
-                        "center",
-                        "centerRight",
-                        "bottomLeft",
-                        "bottomCenter",
-                        "bottomRight"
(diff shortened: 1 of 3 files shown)


vusters (changes requested):
Careful with this. selectedItemIndex is used to know if you click on a card. By setting it to 0 you cannot click on the first item. Please re-do this.

vinothvino42:
> Careful with this. selectedItemIndex is used to know if you click on a card. By setting it to 0 you cannot click on the first item. Please re-do this.

We're using this selectedItemIndex in different scenarios.
1. For setting the initial index of the carousel which starts the carousel from that child of index.
2. To get the selectedItemIndex when the user clicks on the card
3. To get the current index which is displaying in the carousel

Looks like it's confusing. Can we expose two properties?

currentIndex - Which will be used for the first and the second point above. 0 is the default value if it's not given by the user
selectedItemIndex - It'll return the value only if the card selected or else it'll have -1 as the value

vusters:
> > Careful with this. selectedItemIndex is used to know if you click on a card. By setting it to 0 you cannot click on the first item. Please re-do this.
> 
> @vusters We're using this selectedItemIndex in different scenarios.
> 
> 1. For setting the initial index of the carousel which starts the carousel from that child of index.
> 2. To get the current index which is displaying in the carousel
> 3. To get the selectedItemIndex only when the user clicks on the card
> 
> Looks like it's confusing. Can we expose two properties?
> 
> `currentIndex` - Which will be used for the first and the second point above. 0 is the default value if it's not given by the user `selectedItemIndex` - It'll return the value only if the card selected or else it'll have -1 as the value

Ok, good ideal. Keep currentIndex internal for us to track, and a `selectedIndex` to expose which card is selected.

vinothvino42:
@vusters Can you review this PR again? See the description for the new changes

kmahmood74:
@vusters need your review on this one as well as Vinoth made changes

assets/schema/ensemble_schema.json
```diff
@@ -2812,8 +2812,15 @@
                       "type": "string",
                       "description": "Where to display the indicator if specified",
                       "enum": [
```
vusters, line 2814: don't hardcode this. There must be an alignment schema that we declare already. If not then create it.

lib/widget/carousel.dart
```diff
@@ -464,4 +463,37 @@ enum IndicatorType {
   custom,
 }
 
-enum IndicatorPosition { bottom, top }
+enum IndicatorPosition {
```
vusters, line 466: why create another enum? Just use Alignment?

lib/widget/carousel.dart
```diff
@@ -149,7 +148,9 @@ class MyController extends BoxController {
   // for multi view this dispatch when clicking on a card
   EnsembleAction? onItemTap;
   EnsembleAction? onItemChange;
-  int selectedItemIndex = -1;
```
vusters, line 152: So whoever used selectedItemIndex previously will have their page ceased to work? Please make sure our App Developers are aware of this so they can make changes.

assets/schema/ensemble_schema.json
```diff
@@ -2819,9 +2819,17 @@
                       "type": "boolean",
                       "description": "Determines if carousel should loop infinitely or be limited to item length. Default (false)"
                     },
-                    "selectedItemIndex": {
+                    "selectedIndex": {
```
vusters, line 2840: I thought we agreed one will be internal and the other to be public. I'm looking at this as a customer and I don't understand the difference between selectedIndex and currentIndex.

vinothvino42: `currentIndex` - Which will be used for the first and the second point above. 0 is the default value if it's not given by the user
`selectedIndex` - It'll return the value only if the card selected or else it'll have -1 as the value

lib/widget/carousel.dart
vinothvino42: Yes, they need to change it to `selectedIndex`

lib/widget/carousel.dart
vinothvino42: Done

assets/schema/ensemble_schema.json
vinothvino42: Done

vinothvino42:
@vusters I updated the carousel indicatorPosition with using `getAlignment` from the Utils and removed IndicatorPosition enum. Also updated the indicatorPosition of carousel schema with `type-alignment` payload.

Can you review it again?

vusters (approved):

Issue #958, opened 2023-11-07
Add callbackAfterInit methods for deep link


vinothvino42:

vinothvino42:
@vusters 
PR - https://github.com/EnsembleUI/ensemble/pull/959

PR #959, opened 2023-11-07
Add callbackAfterInit methods for executing functions after the app is initialized

vinothvino42:
Ticket: #958

In this PR, I added an `addCallbackAfterInitialization` and `getCallbacksAfterInitialization`. Also updated the ensemble app to execute all the registered functions

### DIFF
--- a/lib/ensemble.dart
+++ b/lib/ensemble.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 import 'dart:developer';
+import 'dart:math' as math;
 
 import 'package:ensemble/ensemble_app.dart';
 import 'package:ensemble/ensemble_provider.dart';
@@ -45,6 +46,23 @@ class Ensemble {
   void setExternalScreenWidgets(Map<String, CustomBuilder> widgets) {
     externalScreenWidgets = widgets;
   }
+
+  final List<Map<String, Object?>> _afterInitMethods = [];
+  void addCallbackAfterInitialization(
+      {required Function method,
+      List<dynamic>? positionalArgs,
+      Map<String, dynamic>? namedArgs}) {
+    final function = {
+      'id': math.Random().nextInt(100000),
+      'method': method,
+      'positionalArgs': positionalArgs,
+      'namedArgs': namedArgs
+    };
+    _afterInitMethods.add(function);
+  }
+
+  List<Map<String, Object?>> getCallbacksAfterInitialization() =>
+      _afterInitMethods;
 
   late FirebaseApp ensembleFirebaseApp;
   static final Map<String, dynamic> externalDataContext = {};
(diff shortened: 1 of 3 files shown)


vusters:
@vinothvino42 let's do this right and remove the hacky delay. Try this:
1. Add `Ensemble().addCallbackAfterInitialization(func)`. In your deep link, call this and pass the function to navigate the screen. The function will just add the function to the list of functions it stores internally.
2. in EnsembleApp initState(), after the first frame look at Ensemble().getCallbacksAfterInitialization() and execute them in order. Make sure to clear each func out before you execute, in case some error causes the function to stay there indefinintely and hose the app.

vinothvino42:
@vusters Can you review it again?

lib/ensemble_app.dart
```diff
@@ -117,10 +117,38 @@ class EnsembleAppState extends State<EnsembleApp> with WidgetsBindingObserver {
       Workmanager().initialize(callbackDispatcher, isInDebugMode: false);
       initDeepLink(AppLifecycleState.resumed);
     }
+
+    WidgetsBinding.instance.addPostFrameCallback((_) async {
+      List<int?> functionIds = [];
+      final callbacks = Ensemble().getCallbacksAfterInitialization();
+      for (final Map<String, Object?> callback in callbacks) {
+        final id = callback['id'] as int?;
```
vusters: see my comment above, it'll simplify this code.

lib/ensemble_app.dart
```diff
@@ -117,10 +117,38 @@ class EnsembleAppState extends State<EnsembleApp> with WidgetsBindingObserver {
...
+
+        if (method != null) {
+          await Function.apply(method, positionalPayloads, namedParams);
+          functionIds.add(id);
+        }
+      }
+
+      // Looping function id to remove the functions in the callback object with the id
+      for (final id in functionIds) {
+        callbacks.removeWhere((element) => element['id'] == id);
```
vusters: Please read my original comment for this task. Either you put function execution inside try/catch, or you remove the item then execute (with a try catch). This way if the function throws an exception, our app is not hosed permanently. Remmeber this is what the app will do on start-up - a bad function will mess the app, regardless if you restart the app or not.

vusters (changes requested):
Please read my original comments before you implement this.

lib/ensemble.dart
```diff
@@ -46,6 +47,23 @@ class Ensemble {
     externalScreenWidgets = widgets;
   }
 
+  final List<Map<String, Object?>> _afterInitMethods = [];
+  void addCallbackAfterInitialization(
+      {required Function method,
+      List<dynamic>? positionalArgs,
+      Map<String, dynamic>? namedArgs}) {
+    final function = {
```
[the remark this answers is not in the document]
vinothvino42: @vusters 
I tried changing it to the list of functions, but how can I pass a method with an argument to the ```addCallbackAfterInitialization(method: )``` method?

Ex: 
```dart
Ensemble()
        .addCallbackAfterInitialization(method: navigateToScreen(URL));
```

vusters (approved):

PR #960, opened 2023-11-07
Youtube

fullstackctp:
Added video position indicator

### DIFF
--- a/assets/schema/ensemble_schema.json
+++ b/assets/schema/ensemble_schema.json
@@ -6011,48 +6011,52 @@
           "description": "The URL source to the youtube video"
         },
         "aspectRatio" : {
-          "type": "double",
+          "type": "number",
           "description" : "Video aspect ratio"
         },
         "autoplay": {
-          "type" : "bool",
+          "type" : "boolean",
           "description" : "Automatically start the video when player is loaded. (default False)"
         },
         "videoList" : {
-          "type" : "array",
+          "type" : "object",
           "description" : "List of videos to be played within a single player."
         },
         "startSeconds": {
-          "type" : "double",
+          "type" : "number",
           "description" : "specifies the time from which the first video in the list(or single video) should start playing"
         },
         "endSeconds": {
-          "type": "double",
+          "type": "number",
           "description": "Ends the video after the certain number of seconds (works with Single video)"
         },
         "showAnnotations" : {
-          "type" : "bool",
+          "type" : "boolean",
           "description" : "Showing the annotations of the video"
         },            
         "playbackRate" : {
-          "type": "double",
+          "type": "number",
           "description": "For changing the speed at which the video is displayed"
         },
         "showControls": {
-          "type" : "bool",
+          "type" : "boolean",
           "description" : "For showing the controls on the video within the player(Like in youtube)"
         },
         "showFullScreenButton": {
-          "type" : "bool",
+          "type" : "boolean",
           "description" : "To show the fullscreen button of the video"
         },
         "enableCaptions" : {
-          "type": "bool",
+          "type": "boolean",
           "description" : "To enable any captions in the video(default language of caption is English)"
         },
         "volume": {
           "type": "integer",
           "description" : "Changes the volume. (max = 100, min = 0)"
+        },
+        "videoPosition": {
+          "type": "boolean",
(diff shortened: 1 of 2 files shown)


kmahmood74 (approved):

lib/widget/youtube/youtubestate.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:ensemble/widget/youtube/native/youtube_platform_native.dart'
+    if (dart.library.html) './web/youtube_platform_web.dart';
+
+abstract class YoutubeWeb {
+  factory YoutubeWeb() => getInstance();
+  void createWebInstance();
```
kmahmood74, line 6: it doesn't make sense to call it createWebInstance when the native has to implement it as well. 
Similarly it does not make sense to call this abstract class YoutubeWeb when you know native class has to implement it as well. 

Just call it YouTubeBase

lib/widget/youtube/youtubestate.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:ensemble/widget/youtube/native/youtube_platform_native.dart'
```
kmahmood74, line 1: this doesn't seem right as you are always importing the native libraries. So the native will be imported even for web? is it import or export?

just look at lib/widget/webview/webviewstate.dart to see how it should be done

lib/widget/youtube/native/youtube_platform_native.dart
```diff
@@ -0,0 +1,8 @@
+import 'package:ensemble/widget/youtube/youtubestate.dart';
+
+YoutubeWeb getInstance() => YoutubeNativeInstance();
+
+class YoutubeNativeInstance implements YoutubeWeb {
```
kmahmood74, line 5: doesn't make sense for YoutubeNativeInstance to implement YoutubeWeb. Your code should be understandable and nicely written. This doesn't make sense

kmahmood74 (changes requested):
1. Everywhere change your class names to have capital T i.e. YouTube and YouTubeNative etc. 
2. Have you tested your code with multiple YouTube player instances on the same screen?

lib/widget/youtube/youtubestate.dart
fullstackctp: To fix the issue where the youtubeplayer doesn't work sometimes on the ensemble web studio. This issue is actually an issue from the package side. 
https://github.com/sarbagyastha/youtube_player_flutter/issues/819

In the issue, it was said to load the WebYoutubePlayerIframePlatform(). This requires importing libraries that work on the web only. 

The native library doesn't get imported for the web. It is actually done in the same way as the lottie widget. 
```
export 'native/lottiestate.dart' if (dart.library.html) 'web/lottiestate.dart';
```
I have tested multiple youtube widgets in the same screen. It doesn't load the video in the player in web. That is why I added the videoList for loading the multiple video in a single player. This problem exists in the demo that they provided. https://sarbagyastha.com.np/youtube_player_flutter/#/

kmahmood74:
I understand the issue and saw Vu’s comments on discord. My point is that
you are doing an import and not export. The Lottie code you posted is using
export as well.

All other comments still apply.

lib/widget/youtube/youtubestate.dart
kmahmood74: @fullstackctp think about using export instead of import and have a library, look at webview as an example. It seems cleaner, what do you think?

kmahmood74:
@fullstackctp does it fix the multiple youtube players on the page issue?

kmahmood74:
@fullstackctp please create a ticket or multiple tickets and link here. All your commits must have the ticket as the commit comment

fullstackctp:
> @fullstackctp does it fix the multiple youtube players on the page issue?

Yes. The recent commit i just did fixes that

> @fullstackctp please create a ticket or multiple tickets and link here. All your commits must have the ticket as the commit comment

I have added new tickets and linked to this PR.

kmahmood74:
@snehmehta kindly review this as well. this is deeper into flutter than I am comfortable reviewing

kmahmood74:
@snehmehta please review this

kmahmood74 (approved):

PR #961, opened 2023-11-07
Added Headers in webview

fullstackctp:
Added headers in webview, similar to headers in api

### DIFF
--- a/lib/widget/webview/webview.dart
+++ b/lib/widget/webview/webview.dart
@@ -7,6 +7,7 @@ import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
 import 'package:ensemble/framework/action.dart' as ensemble;
 import 'package:webview_flutter/webview_flutter.dart';
+import 'package:yaml/yaml.dart';
 
 class EnsembleWebView extends StatefulWidget
     with Invokable, HasController<EnsembleWebViewController, WebViewState> {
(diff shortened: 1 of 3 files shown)


lib/widget/webview/webview.dart
```diff
@@ -98,3 +101,17 @@ List<Map<String, dynamic>> getListOfMap(list) {
   }
   return [];
 }
+
+Map<String, String> parseYamlMap(value) {
+  Map<String, String> result = {};
+  if (value is YamlMap) {
+    YamlMap yamlMap = value;
+    for (var i in yamlMap.entries) {
+      result.addAll({
```
vusters, line 110: if either the key is empty, do NOT add it. You are causing unnecessary bugs here.

vusters (changes requested):

lib/widget/webview/webview.dart
```diff
@@ -98,3 +101,17 @@ List<Map<String, dynamic>> getListOfMap(list) {
   }
   return [];
 }
+
+Map<String, String> parseYamlMap(value) {
+  Map<String, String> result = {};
+  if (value is YamlMap) {
+    YamlMap yamlMap = value;
+    for (var i in yamlMap.entries) {
```
kmahmood74, line 109: just a small note - can you please use a descriptive variable name instead of i. I see this in your other PRs as well. Just call it entry in this case. i is the convention for the counter when you are iterating with an index.

kmahmood74 (approved):
approved with a minor change request

Issue #356, opened 2023-03-15
Page inputs need to be explicit


vusters:

kmahmood74:
as long as we allow the inputs to be optional, it's all good.

hemish11:
Didn't got on what exactly to do

amin-nas:
> Didn't got on what exactly to do

@hemish11 In Ensemble, app builder can pass inputs to screens, widgets, and APIs. For Widgets and APIs, app builder explicitly declares the inputs. For Views, it is implicit. We need to make screen inputs also explicit, but for backward compatibility, it should be optional.

first view, linking to another one and passing inputs
```yaml
View:
  body:
    Button:
       onTap:
         navigateScreen:
           name: AnotherView
           inputs:
             foo: bar
```

second view, receiving inputs
```yaml
View:
  inputs:
    - foo
```

kmahmood74:
lots of discussion and final YAML in the associated PR - https://github.com/EnsembleUI/ensemble/pull/965

```
View:
  inputs:
     # short-hand for just name only. This is how it works today
     - occupation
     
     # more details
     - name: age
       type: number
       required: true
       default: 20
```

hemish11:
Also @kmahmood74 Should we include some validation types, like email validations and something like that

kmahmood74:
> Also @kmahmood74 Should we include some validation types, like email validations and something like that

not at this time. Let's start with a simple one with just the primitive types.

PR #965, opened 2023-11-08
WIP: Added support for inputs to be defined explicitly

hemish11:
Closes https://github.com/EnsembleUI/ensemble/issues/356

### DIFF
--- a/assets/schema/ensemble_schema.json
+++ b/assets/schema/ensemble_schema.json
@@ -16,6 +16,10 @@
         "body"
       ],
       "properties": {
+        "inputs": {
+          "type": "array",
+          "description": "List of all the inputs user can send to a given View."
+        },
         "header": {
           "type": "object",
           "description": "Configure the application header",


kmahmood74:
regarding allowing inputs - @hemish11 @vusters @snehmehta  I think we must have an "optional default value" for each input. As we create re-usable widgets and templates, they simply won't work without the default values and default values should not be in the code with checks for null but should be explicitly defined in the input section. 

this will allow people to view the template and then configure it.

hemish11:
Also, it would be good to define whether an input is required or not, and if it is required, then it can throw an error if that parameter is not provided.

kmahmood74:
> Also, it would be good to define whether an input is required or not, and if it is required, then it can throw an error if that parameter is not provided.

I agree. 
Please come up with the YAML proposal for this.

hemish11:
We can do something like:
```yaml
View:
  input:
    - parameter: # Not sure about the name parameter, maybe can change it to something else
        name: first_name # The name of the parameter
        isRequired: true # Defaults to false
    - parameter:
        name: last_name
    - parameter:
        name: words[] # Maybe we can add [] to define it as an array
```

What should be the exact word instead of the parameter? Also wanted suggestions on adding [] in the postfix of a name to make it an array and thereby allow some preliminary validations or so.

kmahmood74:
suggest the following variation on what you proposed. I think we should define the type as well even if we don't enforce it for now. There are just a few types - string, number, object (could be js or json), array, date, boolean, any

any -> its the default. Allows us to be backward compatible. Also useful in places where the screen accepts different types for the same input. For example, can accept names as a string if there is one or as an array if more than one. 

```yaml
View:
  inputs:
    - input: 
        name: first_name # The name of the parameter
        isRequired: true # Defaults to false
        default: John
    - input:
        name: last_name
       default: Doe
       type: string
    - input:
        name: words
        type: array
        default: ['hello']
```
@vusters @snehmehta what do you think? I think we should just do this properly instead of having to go back another dozen times adding things we know we will need for sure

vusters:
Make it backward compatible:
```
View:
  inputs:
     # short-hand for just name only. This is how it works today
     - occupation
     
     # more details
     - name: age
        type: integer
        default: 20
```
This can be the same for API/Custom Widget's inputs.

kmahmood74:
this PR only defines inputs as array in the schema? @hemish11 is it still being developed?

hemish11:
Yes @kmahmood74, It's still under development

kmahmood74:
@hemish11 so we are not doing this? this was closed without any merge?

Issue #962, opened 2023-11-08
Studio freezes with this definition


vusters:
1. Studio will hang with this definition:
```
View:
  body:
    Text:
      text: hello
      styles:
        stackPositionTop: 0
```
stackPositionTop/Bottom/Left/Right should only be used if the widget is inside a Stack. It is not in this case so an error should be shown, instead of the page hung indefinitely. In debug mode Flutter will probably throw and error, but not release mode.

2. It'll also hang with this:
```
View:
  styles:
    useSafeArea: true
  body:
    Text:
      text: hhello
      styles:
        expanded: true
```

The ask is to show a proper error so the user know to fix it. See what we have in StudioDebugger to see how we catch similar issues.

PR #970, opened 2023-11-09
Changes StudioDebugger

fullstackctp:
Added the ability to show error in text widget for when the studio hangs.
#962

### DIFF
--- a/lib/framework/studio_debugger.dart
+++ b/lib/framework/studio_debugger.dart
@@ -1,6 +1,5 @@
 import 'package:ensemble/framework/error_handling.dart';
 import 'package:ensemble/framework/widget/widget.dart';
-import 'package:ensemble/widget/helpers/controllers.dart';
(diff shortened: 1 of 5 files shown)


fullstackctp:
The recording of the web studio
https://drive.google.com/file/d/1d6QdizGNcSct9cF-vjSTn4LVb1mkWJzs/view?usp=sharing

kmahmood74:
> The recording of the web studio https://drive.google.com/file/d/1URkQj8JZszM6g0GcC6BRAJvxQMjduDVN/view?usp=sharing

@fullstackctp  in your video at the 1:23/24 mark, when you save the screen definition after correcting the definition, the error keeps appearing the same way. You have to reload the whole page to make it go away. That is not right. Saving should be enough

kmahmood74:
@vusters this needs review as well

vusters (approved):

Issue #640, opened 2023-07-11
Setting visible property of DataRow in code is not working as expected


anserwaseem:
## Issue
Setting visible property of DataRow in code is not working as expected. Kindly run the given code to reproduce this issue.

## Example Code
```yaml
View:
  body:
    Column:
      children:
        - Button:
            label: Make row visible
            onTap:
              executeCode:
                body: |
                  //@code
                  myRow.visible = true;

        - Column:
            children:
              - DataGrid:
                  DataColumns:
                    - label: Hotel Name
                      type: text
                    - label: Location
                      type: text
                  children:
                    - DataRow:
                        children: 
                          - Text:
                              text: Hello
                          - Text:
                              text: Hello
                    - DataRow:
                        visible: false
                        id: myRow
                        children:
                          - Text:
                              text: Test
                          - Text:
                              text: Test
```

kmahmood74:
@hemish11 this breaks all DataGrids that use item-templates. With item-templates, _children can grow dynamically but rowVisibilities is a static array, will create a PR to comment out some code

kmahmood74:
@hemish11 is this still on or we've given up on it? it's **not** a high priority.

PR #972, opened 2023-11-11
Fixes the bug of setting of visible property of DataRow

hemish11:
Closes https://github.com/EnsembleUI/ensemble/issues/640

### DIFF
--- a/lib/layout/data_grid.dart
+++ b/lib/layout/data_grid.dart
@@ -136,6 +136,7 @@ class EnsembleDataRow extends StatefulWidget
   List<WidgetModel>? children;
   ItemTemplate? itemTemplate;
   bool visible = true;
+  // Callback to fire when visible changes
   Function(bool value)? onVisibleChanged;
 
   EnsembleDataRow({super.key});
@@ -164,10 +165,13 @@ class EnsembleDataRow extends StatefulWidget
     return {
       'visible': (dynamic v) {
         final bool newValue = Utils.getBool(v, fallback: visible);
+
+        // checking if old value and new value are different
         if (visible != newValue) {
           visible = newValue;
 
           if (onVisibleChanged != null) {
+            // Firing the callback to update the state in EnsembleDataGrid
             onVisibleChanged!(visible);
           }
         }
@@ -235,6 +239,10 @@ class DataGridState extends WidgetState<DataGrid>
   final List<DataRow> _rows = [];
   final List<Widget> _children = [];
 
+  // List for the visibilities for individual row
+  List<bool> rowVisibilities = [];
+  bool isFirstTimeRunning = true;
+
   @override
   void initState() {
     widget._controller.addListener(refreshState);
@@ -282,6 +290,13 @@ class DataGridState extends WidgetState<DataGrid>
 
     _buildDataColumn(scopeManager);
     _buildChildren();
+
+    // Ensure that _parseVisibility runs only once as it is inside buildWidget. Cannot put it in initState
+    if (isFirstTimeRunning) {
+      isFirstTimeRunning = false;
+      rowVisibilities = _parseVisibility();
+    }
+
     _buildDataRow();
 
     // Setting sort column index and order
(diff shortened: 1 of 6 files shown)


kmahmood74 (changes requested):
@hemish11 the only commit here should be the data_grid.dart. Please remove all the other accidental commits. Keep it clean please.

lib/layout/data_grid.dart
```diff
@@ -159,7 +163,19 @@ class EnsembleDataRow extends StatefulWidget
...
     return {
-      'visible': (dynamic v) => visible = Utils.getBool(v, fallback: visible)
+      'visible': (dynamic v) {
+        final bool newValue = Utils.getBool(v, fallback: visible);
+
+        // checking if old value and new value are different
+        if (visible != newValue) {
+          visible = newValue;
+
+          if (onVisibleChanged != null) {
```
kmahmood74, line 173: 1. please call it onVisibilityChanged
2. simply by just doing this - onVisibleChanged?.call(visible);

lib/layout/data_grid.dart
```diff
@@ -270,6 +290,13 @@ class DataGridState extends WidgetState<DataGrid>
 
     _buildDataColumn(scopeManager);
     _buildChildren();
+
+    // Ensure that _parseVisibility runs only once as it is inside buildWidget. Cannot put it in initState
+    if (isFirstTimeRunning) {
```
kmahmood74, line 295: this optimization can cause issues if buildWidget is run again and in the meantime the row's visibility has changed. There is no need for this optimization, it can cause more issues than it can solve. Also the performance impact would be probably a few nano seconds at best

lib/layout/data_grid.dart
```diff
@@ -366,8 +412,17 @@ class DataGridState extends WidgetState<DataGrid>
...
         throw Exception("Direct children of DataGrid must be of type DataRow");
       }
+
       EnsembleDataRow child = w;
-      if (!child.visible) {
+
+      // Callback to update the visible of a given row changes
+      child.onVisibleChanged = (value) {
+        setState(() {
+          rowVisibilities[index] = value;
```
kmahmood74, line 421: won't need this once you remove the optimization asI suggested above. You will still need setState but won't need to set the rowVisibilities

lib/layout/data_grid.dart
```diff
@@ -356,7 +398,11 @@ class DataGridState extends WidgetState<DataGrid>
 
   void _buildDataRow() {
     _rows.clear();
-    for (Widget w in _children) {
+    // Added children.asMap().entries to have access for individual element index
+    for (final element in _children.asMap().entries) {
```
kmahmood74, line 402: I don't understand why you are converting a list into a map to get its index. Why not just do - 

for ( int index=0;index<_children.length;index++ ) {
  Widget w = _children[index];
...
}
that's all, why do you need a map for?

lib/layout/data_grid.dart
```diff
@@ -159,7 +163,19 @@ class EnsembleDataRow extends StatefulWidget
...
+      'visible': (dynamic v) {
+        final bool newValue = Utils.getBool(v, fallback: visible);
+
+        // checking if old value and new value are different
+        if (visible != newValue) {
+          visible = newValue;
+
+          if (onVisibleChanged != null) {
+            // Firing the callback to update the state in EnsembleDataGrid
+            onVisibleChanged!(visible);
```
kmahmood74, line 175: this event is not needed. Can you just set the visibility on the widget and simply call setState? Wouldn't that be enough?

lib/layout/data_grid.dart
```diff
@@ -317,6 +344,21 @@ class DataGridState extends WidgetState<DataGrid>
     );
   }
 
+  // Parsers the visibility out of the list of rows
```
kmahmood74, line 347: I don't understand why this is needed?

kmahmood74 (changes requested):

lib/layout/data_grid.dart
```diff
@@ -159,7 +163,19 @@ class EnsembleDataRow extends StatefulWidget
...
     return {
-      'visible': (dynamic v) => visible = Utils.getBool(v, fallback: visible)
+      'visible': (dynamic v) {
+        final bool newValue = Utils.getBool(v, fallback: visible);
+
+        // checking if old value and new value are different
+        if (visible != newValue) {
+          visible = newValue;
+
+          if (onVisibilityChanged != null) {
```
kmahmood74, line 173: don't need the null check. The ? in the statement already checks for that. Dart is pretty cool :-)

kmahmood74 (approved):
suggest a minor change, see comment. Rest looks good. Approving with the minor change

Issue #979, opened 2023-11-13
HTML widget should have maxLines option


sharjeelyunus:
Need maxLines in HTML widget, and overflow should be ellipsis, also the text color, font size and font weight.

kmahmood74:
@hemish11 please treat this as high priority as a customer needs it

hemish11:
@kmahmood74 @vusters @snehmehta Hows this EDL looks like:
```yaml
View:
  styles:
    scrollableView: true
  header:
    title: Home
  body:
    Column:
      children:
        - Html:
            cssStyles: # List of styles similar to CSS
              - style:
                  selector: "#Hello" # or ".Hello" or "div"
                  properties:
                    # All the CSS properties, it will be entered as key value pair just as in CSS
                    color: red
                    borderWidth: 10 # The naming can be of camelCase. Code would convert the camelCase to param-case
              - style:
                  selector: "#tag2"
                  properties:
                    color: blue
                    borderRadius: 10
            styles:
              borderRadius: 20
              borderColor: red
              borderWidth: 2
            text: |
              <div id="hello">
              Until recently, the prevailing view assumed lorem ipsum was born as a nonsense text. “It's not Latin, though it looks like it, and it actually says nothing,” Before & After magazine answered a curious reader, “Its ‘words’ loosely approximate the frequency with which letters occur in English, which is why at a glance it looks pretty real.”

              As Cicero would put it, “Um, not so fast.”

              The placeholder text, beginning with the line “Lorem ipsum dolor sit amet, consectetur adipiscing elit”, looks like Latin because in its youth, centuries ago, it was Latin.

              Richard McClintock, a Latin scholar from Hampden-Sydney College, is credited with discovering the source behind the ubiquitous filler text. In seeing a sample of lorem ipsum, his interest was piqued by consectetur—a genuine, albeit rare, Latin word. Consulting a Latin dictionary led McClintock to a passage from De Finibus Bonorum et Malorum (“On the Extremes of Good and Evil”), a first-century B.C. text from the Roman philosopher Cicero.
              </div>
```

I thought to include support for both css tags and also for the box wrapper. Also flutter has these tags supported in the styles parameter so will be adding suppport for all of them:
- background-color
- color
- counter-increment
- counter-reset
- direction
- display
- font-family
- font-feature-settings
- font-size
- font-style
- font-weight
- height
- letter-spacing
- list-style-image
- list-style-type
- list-style-position
- margin
- padding
- text-align
[...]

PR #987, opened 2023-11-14
Support for styling in html widget

hemish11:
Closes https://github.com/EnsembleUI/ensemble/issues/979


vusters (commented):
Are you done @hemish11 ? it just wraps it in a box. IF you set the width/height it will only affect the outer box. Does the html content scroll properly?

What about the rest of the stylings we talked about?

hemish11:
No no, it's not done yet, That's why I kept it as a draft pr @vusters.

snehmehta (approved):

PR #988, opened 2023-11-14
Revert "Fixes the bug of setting of visible property of DataRow"

kmahmood74:
Reverts EnsembleUI/ensemble#972

This PR does not take into account updating the DataGrid rows dynamically through an item-template. It breaks all grids as the the _children array can grow dynamically and the following code will break (line 420) - 

```
      if (!rowVisibilities[index]) {
        continue;
      }
```
as the rowVisibility is static


kmahmood74:
@hemish11 this breaks all DataGrids that use item-templates. With item-templates, _children can grow dynamically but rowVisibilities is a static array, will create a PR to comment out some code

we need to revert your code immediately. Please review and approve

PR #989, opened 2023-11-15
Revert "LOTTIE - Refactored code, added support for methods and callbacks"

kmahmood74:
Reverts EnsembleUI/ensemble#915


kmahmood74:
regression, can't even run any app. 

image

PR #969, opened 2023-11-09
enhance/selectable-text

snehmehta:
closes: https://github.com/EnsembleUI/ensemble/issues/406

Enable text selection for `Text` widget for web only.

PR #990, opened 2023-11-15
Revert "enhance/selectable-text"

kmahmood74:
Reverts EnsembleUI/ensemble#969


kmahmood74:
@snehmehta just run kitchen sink, appid: e24402cb-75e2-404c-866c-29e6c3dd7992
with this change, none of the buttons even on the first page are clickable

PR #991, opened 2023-11-15
Simplified callbackAfterInit method call logic and handle errors

vinothvino42:
Ticket: https://github.com/EnsembleUI/ensemble/issues/958
Enhancement:

In this PR, I entirely refactored the method calling after an ensemble app initialized.

1. `addCallbackAfterInitialization` - It stores the method in the `_afterInitMethods` (List of type Function)
2. `getCallbacksAfterInitialization` - It returns `_afterInitMethods` (List of type Function)
3. When the ensemble app is initialized or resumed, it executes the `executeCallbacks` method which will remove a method which is stored in the `getCallbacksAfterInitialization` and execute that removed method.
4. Error handled

### DIFF
--- a/lib/ensemble.dart
+++ b/lib/ensemble.dart
@@ -47,12 +47,12 @@ class Ensemble {
     externalScreenWidgets = widgets;
   }
 
-  final List<Function> _afterInitMethods = [];
+  final Set<Function> _afterInitMethods = {};
   void addCallbackAfterInitialization({required Function method}) {
     _afterInitMethods.add(method);
   }
 
-  List<Function> getCallbacksAfterInitialization() => _afterInitMethods;
+  Set<Function> getCallbacksAfterInitialization() => _afterInitMethods;
 
   late FirebaseApp ensembleFirebaseApp;
   static final Map<String, dynamic> externalDataContext = {};
--- a/lib/ensemble.dart
+++ b/lib/ensemble.dart
@@ -47,22 +47,12 @@ class Ensemble {
     externalScreenWidgets = widgets;
   }
 
-  final List<Map<String, Object?>> _afterInitMethods = [];
-  void addCallbackAfterInitialization(
-      {required Function method,
-      List<dynamic>? positionalArgs,
-      Map<String, dynamic>? namedArgs}) {
-    final function = {
-      'id': math.Random().nextInt(100000),
-      'method': method,
-      'positionalArgs': positionalArgs,
-      'namedArgs': namedArgs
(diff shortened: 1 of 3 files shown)


lib/ensemble.dart
```diff
@@ -47,22 +47,12 @@ class Ensemble {
...
-    };
-    _afterInitMethods.add(function);
+  final List<Function> _afterInitMethods = [];
+  void addCallbackAfterInitialization({required Function method}) {
+    _afterInitMethods.add(method);
   }
 
-  List<Map<String, Object?>> getCallbacksAfterInitialization() =>
-      _afterInitMethods;
+  List<Function> getCallbacksAfterInitialization() => _afterInitMethods;
```
vusters, line 55: To prevent duplications, can you change this to a Set? if the same function is added twice my accident, we only call once.

lib/ensemble_app.dart
```diff
@@ -133,31 +133,17 @@ class EnsembleAppState extends State<EnsembleApp> with WidgetsBindingObserver {
...
-    for (final Map<String, Object?> callback in callbacks) {
-      final id = callback['id'] as int?;
-      final method = callback['method'] as Function?;
-      final positionalPayloads = callback['positionalArgs'] as List<dynamic>?;
-      final namedPayloads = callback['namedArgs'] as Map<String, dynamic>?;
-      Map<Symbol, dynamic>? namedParams;
-      namedPayloads?.forEach((key, value) {
-        namedParams = {Symbol(key): value};
+    final callbacks = List.from(Ensemble().getCallbacksAfterInitialization());
+    try {
```
vusters, line 137: do try catch inside each function execution, so if one fail we still continue with the next.

vusters (commented):
I see initDeepLink() is called excessively on initState() but also when the app lifecycle changes. That can happen quite a few times. Does this create multiple of the same functions to be executed? Please test it.

lib/ensemble.dart
vinothvino42: Done. Can you check again?

lib/ensemble_app.dart
vinothvino42: Done

vinothvino42:
@vusters I changed as you suggested. Can you review it again?

1. To prevent duplications, can you change this to a Set? if the same function is added twice by accident, we only call once.
2. do try catch inside each function execution, so if one fail we still continue with the next.

vusters (approved):

PR #994, opened 2023-11-16
Fixed possible issues due to borderWidth and borderColor parameters

### LINKED ISSUE #898: Support adding Border to single side of Widget
There should be a way to specify the side (left, right, top, bottom) on which the border should be applied.
Currently, **borderWidth** is supported and applies borders to all four sides.

hemish11:
Fixes #898

Adding support for setting individual border width and color. It supports both CSS kind off syntax and individual parameters.
CSS syntax would include like `borderWidth: "4 3"` which means to apply 4 border width to top and bottom and 3 to left and right. Similarly for `borderWidth: "4 3 2"` which means it would apply border width of 4 to top, 3 to left and right and 2 to bottom. And `borderWidth: "4 3 2 1"` implies setting borderWidth of 4 to top, 3 to right, 2 to bottom and 1 to left. Same pattern is followed for `borderColor`

Also added individual parameters like `topBorderWidth`, `bottomBorderWidth` and so on.


lib/widget/lottie/web/lottiestate.dart
```diff
@@ -92,7 +92,7 @@ class LottieState extends WidgetState<EnsembleLottie>
     bool repeat = widget.controller.repeat;
     bool autoPlay = widget.controller.autoPlay;
     if (widget.controller.width == null || widget.controller.height == null) {
-      throw LanguageError(
```
vusters: you want to show this error so the AppDev knows to fix this. This is not a silent error

lib/widget/helpers/widgets.dart
```diff
@@ -214,22 +214,22 @@ class BoxWrapper extends StatelessWidget {
                           top: getBorderSide(
                             context,
                             boxController.borderWidth?.top,
-                            boxController.borderColor?.top,
+                            boxController.borderColor.top,
```
vusters, line 217: does your Editor not complain about null safety? User can set borderWidth for example but exclude borderColor, and that will still show the border with default color. Please fix this.

vusters (changes requested):

kmahmood74:
ok so this has been sitting around for over a month. @hemish11 please get on top of your PRs please

hemish11:
@kmahmood74 I already added all the changes requested in it here, waiting for it to be reviewed again

vusters:
@hemish11 you only merged from main. I don't see the changes. Please check this.

hemish11:
Temporarily closing due to monorepo

Issue #964, opened 2023-11-08
BUG: Change switch color


sharjeelyunus:
in Switch widget we have styles property, but nothing works here

I tried this

```yaml
fillColor: 0xFFD6D6CE
borderColor: amber
focusedBorderColor: blue
disabledBorderColor: grey
```

kmahmood74:
@vinothvino42 please review, this is blocking a customer

vinothvino42:
@kmahmood74 Our switch and checkbox have a controller which extends from FormFieldController. So we're having all the above styles and our `OnOff-payload` in schema uses the `FormInput-payload` properties too

Can I have styles like below and see the screenshot

1. activeColor - yellow
2. inactiveColor - blue
3. activeThumbColor - black
4. inactiveThumbColor - grey


<img src="https://github.com/EnsembleUI/ensemble/assets/12869588/d7a8745b-1820-489d-9ebf-c1f35e2aaec8" width="350" />

vinothvino42:
@vusters PR - #995

vinothvino42:
@vusters Can you review it again? I updated it

PR #995, opened 2023-11-16
Add styles for the switch and checkbox widget

vinothvino42:
Ticket: #964

In this PR, I just removed the `form_checkbox.dart` and created two widgets called `Switch` and `Checkbox`

**Sample EDL:**

```yaml
View:
  header:
    title: Switch & Checkbox
  body:
    Column:
      styles: { gap: 24, padding: 24, scrollable: true }
      children:
        - Checkbox:
            leadingText: Leading text
            styles:
              {
                selectedColor: yellow,
                unselectedColor: blue,
                checkColor: pink,
              }

        - Switch:
            leadingText: Set to true and requred
            required: true
            value: true
            styles:
              {
                activeColor: yellow,
                activeThumbColor: black,
                inactiveColor: blue,
                inactiveThumbColor: grey,
              }

```

**Selected State:**

**Unselected State:**


lib/widget/input/form_checkbox.dart
```diff
@@ -149,19 +162,53 @@ class OnOffState extends FormFieldWidgetState<OnOffWidget> {
   /// we adjust the hit area of the checkbox here. 40px is smaller than the default (48px)
   /// but it seem to be reasonable for touch device, plus the alignment inside
   /// form is much better (align well with the rest of the input widgets)
+
   Widget get aCheckbox {
     return SizedBox(
         width: 40,
         height: 40,
         child: Checkbox(
+            side: widget._controller.inactiveColor != null
```
vusters, line 171: Why are we decorating the border if it's inActive? The user sets inActive thinking it's equivalent to !active, but instead the fillColor stays the same but they get a different border? This doesn't make sense.

assets/schema/ensemble_schema.json
```diff
@@ -1569,6 +1569,26 @@
             },
             "trailingText": {
               "type": "string"
+            },
+            "activeColor": {
```
vusters, line 1573: Please don't include all properties that don't make sense. Break these up. Each widget should only get properties that apply to them.

vusters (changes requested):

assets/schema/ensemble_schema.json
vinothvino42: Can you read the above description of the PR? I just updated

lib/widget/input/form_checkbox.dart
vinothvino42: Can you read the above description of the PR? I just updated

vinothvino42:
@vusters I split this into two separate files for two widgets (switch and checkbox) with their own controller and styles. Can you review it?

vusters (approved):

Issue #967, opened 2023-11-09
Fix Image URL source cache issue


vusters:
We cache the image's URL so they can load faster the next time:
```
Image:
  source: https://example.com/1234
```
This seems to be a problem if the image has been updated at the URL (e.g. Profile photo changed and uploaded to the server). When we load this page again, the image still points to the old one since it was cached.

Please check:
- check whether the image should be cached? as in if does the Image bust the cache if its http header says it has changed
- If we can't cache the image by the http header, so we need a flag that say `cache: false` to force the image to be loaded every time?

PR #996, opened 2023-11-16
image_cache 

fullstackctp:
For the ticket #967


lib/widget/image.dart
```diff
@@ -132,7 +134,9 @@ class ImageState extends WidgetState<EnsembleImage> {
       }
 
       return CachedNetworkImage(
```
vusters, line 136: if (widget.controller.cache) 
  use Image.network
else
  use CachedNetworkImage

vusters (changes requested):

fullstackctp:
I have made the required changes however, the problem persists in the web side. The native side requires you to do hot restart for the image.network to work again. 
https://stackoverflow.com/questions/49892074/flutter-image-network-not-updated 

In the flutter image_provider they have added this as a todo
```
// TODO(ianh): Find some way to honor cache headers to the extent that when the
// last reference to an image is released, we proactively evict the image from
// our cache if the headers describe the image as having expired at that point.
```

For now the fix that is provided is to basically add timestamp with the date of last-modified from the cache-headers to provide a difference url if the images changes in the url

vusters:
That's fine. Please use the timestamp as you have had. Go ahead and merge that part in.

snehmehta (approved):

lib/widget/image.dart
```diff
@@ -135,6 +139,20 @@ class ImageState extends WidgetState<EnsembleImage> {
     return rtn;
   }
 
+  Future<String> fetch(String url) async {
+    final http.Response response = await http
+        .get(Uri.parse("${url}timeStamp=${DateTime.now().toString()}"));
```
kmahmood74, line 144: @fullstackctp how can this code ever work - you are putting timeStamp directly after the url without ny query string parameter. 

so if the url is myapp.com/img.png it will become myapp.com/img.pngtimeStamp=.... which will always fail.

You have to check if the url has a query string in it already. If it does, add a query parameter with &. If it doesn't add a new query string with ?

Was this code tested? What am I missing, this would never work

kmahmood74: also you mentioned that you stopped adding the current time but you are doing that here?

lib/widget/image.dart
```diff
@@ -135,6 +139,20 @@ class ImageState extends WidgetState<EnsembleImage> {
...
+    DateTime lastModifiedDateTime =
+        parseHttpDate("${response.headers['last-modified']}");
```
kmahmood74, line 146: what if the last-modified header is not present. You will end up sending "null" as the parameter. 

How will that work?

lib/widget/image.dart
```diff
@@ -135,6 +139,20 @@ class ImageState extends WidgetState<EnsembleImage> {
...
+    if (widget._controller.lastModifiedCache == null ||
+        lastModifiedDateTime.compareTo(widget._controller.lastModifiedCache!) ==
+            1) {
+      widget._controller.lastModifiedCache = lastModifiedDateTime;
+      await EnsembleImageCacheManager.instance.emptyCache();
+    }
+    return "${widget.controller.source}timeStamp=$lastModifiedDateTime";
```
kmahmood74, line 153: where is the query string?

kmahmood74 (changes requested):
This doesn't seem right to me. @vusters has to review this before it is approved.

vusters (approved):

PR #1001, opened 2023-11-17
Disposed animation controller for lottie

hemish11:
Relates to: #914. Forgot to add a dispose for the animation controller which was causing warnings to occur. So added it.


kmahmood74:
@hemish11 you have changed a number of unrelated files due to formatting changes. Please remove those from the commit

kmahmood74 (approved):

PR #1007, opened 2023-11-20
Convert ListView to PagedListView

snehmehta:


assets/schema/ensemble_schema.json
```diff
@@ -3513,14 +3513,26 @@
                       "type": "boolean",
                       "description": "If set true, users can view captured/selected media."
                     },
+                    "instantPreview": {
```
kmahmood74, line 3516: @snehmehta these seem to be related to the camera and not this ticket

snehmehta: yup, it was added by mistake, moving it to it's own PR

kmahmood74 (approved):

PR #1010, opened 2023-11-23
Add backward compatibility for the selectedItemIndex property in the Carousel Widget

vinothvino42:


assets/schema/ensemble_schema.json
```diff
@@ -2898,6 +2898,12 @@
                       "type": "integer",
                       "description": "The initial page to show when first creating the Carousel and the current index of the displaying Carousel Item. Default (0)"
                     },
+                    "selectedItemIndex": {
+                      "type": "integer",
+                      "description": "The initial page to show when first creating the Carousel and the current index of the displaying Carousel Item. Default (0)",
+                      "deprecated": false,
```
snehmehta, line 2904: False? correct me if I am wrong it is supposed to be a deprecated

snehmehta (approved):

assets/schema/ensemble_schema.json
vinothvino42: Yes, but not now. So I added it as false and with deprecation message

vinothvino42: @snehmehta I updated it.

snehmehta (approved):

PR #1015, opened 2023-11-24
Fixed a bug of css styling not being applied for TextOverflow parameter

hemish11:
Closes https://github.com/EnsembleUI/ensemble/issues/979


hemish11:
Also now added the updateMaxLines method

Issue #1017, opened 2023-11-27
keyboard doesn't close clicking inactive places


sharjeelyunus:
see this here, if you click on note input, the keyboard will open up covering the whole input, and there is no way to get out of this keyboard, ideally if you click outside the input keyboard should close automatically

https://studio.ensembleui.com/app/YQGEapU8rrEfhTJJLzz8/screen/gBr5JHleY614Lw4xTGeG

vinothvino42:
@sharjeelyunus PR - https://github.com/EnsembleUI/ensemble/pull/1020

PR #1021, opened 2023-11-29
Add keyboard toolbar with done button

vinothvino42:
Ticket - https://github.com/EnsembleUI/ensemble/issues/1017

In this PR, I added the following
1. Added a toolbar with done button for the keyboard which has input type of `number` and `phone` of `TextInput`
2. `onTapOutside` method in the TextFormField, which will remove the keyboard focus automatically


snehmehta (approved):
So, it unfocus the keyboard and add a done button as well. 
Just check on android as well, if it's UX is same as IOS.

Issue #1024, opened 2023-11-29
Deprecated APIs in Video and QR Code


hemish11:
The current video implementation has a few deprecated APIs being used. Need to move them to newer APIs.

hemish11:
EDIT: Also found the deprecated API inside QR Code, so changing that also

PR #1025, opened 2023-11-29
Changed deprecated APIs for QR code and Video

hemish11:
closes https://github.com/EnsembleUI/ensemble/issues/1024


kmahmood74 (commented):
@hemish11 it's the same packages just new methods? please test the QR code if you haven't already as it has more than a simple change

kmahmood74 (approved):
please test and commit

Issue #1027, opened 2023-11-30
Tapping on push notification not behaving right


anserwaseem:
When app is in **killed** state, 
1/3 times it crashes, 
and other 2/3 times, it doesn't open the specified screen i.e., just opens up Home screen

vinothvino42:
@vusters I checked the issue and found out that getInitialMessage from FirebaseMessaging is not added. I added and tested I'm able to navigate the screen. But there are few other issues.

1. Only for this getInitialMessage, I added handleNotification method to ensemble's addCallbackAfterInitialization like this below to execute this method only after the app is initialized
```dart
Ensemble().addCallbackAfterInitialization(
          method: () => _handleNotification(isFromTerminatedState: true));
```
2. Now it's not navigating to a screen because the (Utils.globalAppKey) NavigatorState's currentContext is null, so it's failed when executing the ScreenController().navigateToScreen with the context.
3. So I added a delay (2 seconds), now the Utils.globalAppKey.currentContext is not null, so navigation is working
4. Now another issue, when the new screen is navigated before completing the home page fully it's got replaced with new widget. See the below video for more info

Link:

PR #1028, opened 2023-11-30
Add getInitialMessage method for handling push notification in Terminated App State

vinothvino42:
Ticket: #1027


vusters (changes requested):
are you missing some files? You are removing the logic to execute callbacks but added somewhere else?

vusters:
Here's how you can resolve the race condition:
- in NotificationManager, have init() returns the initial message.
- in main, await for the NotificationManager's init() and retreieve this initial message.
- if the messge has screenId/screenName, pass them into EnsembleApp.

This will ensure the 1st scren you load will be the screenId/screenName you have.

vinothvino42:
> are you missing some files? You are removing the logic to execute callbacks but added somewhere else?

I added it in the screen widget, it's because the ensemble_app fetches the definition and then only it renders the widget. So I just added it in the screen widget, here the material app is already initialized with the definition.

vinothvino42:
@vusters I'll merge this PR

PR #636, opened 2023-07-11
Add resizeView style for page and device.viewInsetBottom

vinothvino42:
Ticket: #623 
In this PR, I added the following

1. `resizeView` boolean property for the page styles, you can enable or disable it. If enabled, the view will be automatically adjusted by moving the view or input field on top of the keyboard without obscuring it.
2. `device.viewInsetBottom` which provides the keyboard height

PR #1035, opened 2023-12-04
Fix bottom modal sheet scrolling issue

vinothvino42:


snehmehta (commented):
I believe this issue is also occurring other places as well. 
What if we allow ensemble devs tools to add viewInsets bottom.

vinothvino42:
> viewInsets

I added something like this last time - https://github.com/EnsembleUI/ensemble/pull/636#issue-1798591158
But we closed that PR

snehmehta (approved):

PR #1037, opened 2023-12-04
Ability to use actions in js using ensemble object 

### LINKED ISSUE #971: Ability to use actions in js using ensemble object
I want to use following actions in js code using `ensemble` object

1. dismissBottomModal
2. showBottomModal
3. showDialog
4. rateApp

nehal076:
Closes #971


kmahmood74:
@nehal076 once PR has been approved, please document here - https://github.com/EnsembleUI/ensemble_docs/tree/main/actions
and also create Ensemble Kitchen Sink examples. Enhance the existing page or create a new one if needed

lib/framework/action.dart
```diff
@@ -67,6 +67,18 @@ class ShowDialogAction extends EnsembleAction {
   final Map<String, dynamic>? options;
   final EnsembleAction? onDialogDismiss;
 
+  factory ShowDialogAction.from({Invokable? initiator, Map? payload}) {
```
vusters, line 70: Why do we duplicate code? the fromYaml() below should work for regular Map also. Please rename it to `from` and test it with Map.

lib/framework/data_context.dart
```diff
@@ -365,6 +367,16 @@ class NativeInvokable extends ActionInvokable {
         final scope = ScreenController().getScopeManager(buildContext);
         callNativeMethod(buildContext, scope, inputs);
       },
+      ActionType.showBottomModal.name: (inputs) => ScreenController()
```
vusters: This is the old way of registering it. Please put them in the super class ActionInvokable

vusters (changes requested):

lib/action/bottom_modal_action.dart
```diff
@@ -100,12 +96,6 @@ class DismissBottomModalAction extends EnsembleAction {
...
   Future<dynamic> execute(BuildContext context, ScopeManager scopeManager,
       {DataContext? dataContext}) {
-    BuildContext? bottomModalContext =
-        ContextScopeWidget.getRootContext(context);
-    if (bottomModalContext != null) {
-      return Navigator.maybePop(
-          bottomModalContext, scopeManager.dataContext.eval(payload));
-    }
-    return Future.value(null);
+    return Navigator.maybePop(context, scopeManager.dataContext.eval(payload));
```
vusters, line 99: so this will only work for the simplest use case where the widget is defined directly in the same block. The context here can belong to a deeply nested widget - hence the reason we want to travel up to the root to dismiss the dialog.

vusters (changes requested):

lib/action/bottom_modal_action.dart
nehal076: Okay @vusters, actually `ContextScopeWidget.getRootContext(context)` was returning null, so the dialog was never dismissed.

nehal076: @vusters I've done this, if the `rootContext` is not null, it will use root context, else it will use the local context to dismiss the modal

lib/action/bottom_modal_action.dart
```diff
@@ -75,7 +71,7 @@ class ShowBottomModalAction extends EnsembleAction {
               enableDrag: _enableDrag(scopeManager),
               showDragHandle: _enableDragHandler(scopeManager),
               builder: (modalContext) =>
-                  ContextScopeWidget(rootContext: modalContext, child: widget!))
+                  ContextScopeWidget(rootContext: context, child: widget!))
```
kmahmood74, line 74: shouldn't this be modalContext that is being passed in. Why make this change?

kmahmood74 (changes requested):
please remove the unnecessary files like appdelegate, gradle, podfile

lib/action/bottom_modal_action.dart
vusters: and this works even when called from deeply nested widget?

vusters (approved):

PR #1050, opened 2023-12-12
listview: rewrite the simple listview core

snehmehta:
ListView core replaced from packaged to custom slivers.

Introduced simple api for infinite scroll usecase like:
- hasReachedMax ( to stop calling onScrollEnd)
- showLoading setter ( to show / hide loader through EDL)


snehmehta:
merging it as it's required right now, but it does requires a thorough test.

Issue #204, opened 2023-02-07
Event callback for custom widget


vusters:
```yaml
View:
  Column:
    children:
      - CustomWidget:
           inputs:
             label: My Special Widget
           events:
             onDateSelected: 
                 invokeAPI:
                    name: getReservations
                    inputs:
                        date: ${event.date} #this.date should work too I think


CustomWidget:
  inputs:
    - label
  events:
    - onDateSelected:
          data: {date, month, year}
  body:
    Button:
      onTap:
        dispatchEvent:
           onDateSelected:
                 data:
                   date: ${cal.getDate()}
                   month: ${cal.getMonth()}
                   year: ${cal.getYear()}


```

kmahmood74:
assigning to myself as Vu is busy. The Yaml is not right as it confuses inputs and callbacks. Those are two different concepts. 

There are 3 concepts in Custom Widgets (or any widget for that matter) that need to be clearly defined and documented - 

1. Inputs - this is data or configuration that the widget needs to display/act correctly. Think of specific styling or the label of a field being passed as input

2. Events - these are the events that the widget dispatches. For example - onDateSelected dispatched by a Calendar widget. UI components are asynchronous and UI is an event driven system, that's why widgets dispatch events to communicate changes. They don't have outputs i.e. return values that you expect from a synchronous method call

3. Event Callbacks - these are actions that the caller can attach to events. Ensemble actions are perfect for this. 

Here's the proposed YAML. This will also work well in the schema and allow us to auto-complete event handlers
```
View:
  Column:
    children:
      - CustomWidget:
           inputs:
             label: My Special Widget
           events: #I would love to call it eventHandlers or callBacks but events works better to keep consistent with what the custom widget is calling it
             onDateSelected: 
                 invokeAPI:
                    ....


CustomWidget:
  inputs:
    - label
  events:
    - onDateSelected:
          parameters: {date, month, year}
  Button:
    onTap:
      invokeEventHandler:
          onDateSelected:
               parameters:
                   date: ${cal.getDate()}
                   month: ${cal.getMonth()}
                   year: ${cal.getYear()}

```

kmahmood74:
@vusters @amin-nas @snehmehta @evshi please review so I may get started on implementing it. See my comment above for proposed yaml

evshi:
[...]

PR #1116, opened 2024-01-07
https://github.com/EnsembleUI/ensemble/issues/204 support for events …

kmahmood74:
…and callbacks for custom widgets


lib/framework/widget/view_util.dart
```diff
@@ -73,6 +74,10 @@ class ViewUtil {
         payload = item[widgetType];
       }
       def = getDefinition(item);
+      if ( item.keys.length > 1 ) {
```
kmahmood74, line 77: ignore this, it is a no-op and will be in the next change, idea is to detect common issues with tabbing in yaml

kmahmood74:
@vusters please review this. Since I showed you already, this should be a quick one

vusters (approved):

PR #1117, opened 2024-01-08
Change svg image checking logic

vinothvino42:


snehmehta (approved):
let's updated this later

Issue #62, opened 2022-11-02
ability to define re-usable code libraries and import them where needed


kmahmood74:
1. Defining re-usable code
Just like we define screens and widgets as re-usable, code itself is defined as another artifact except this artifact contains javascript code. 
Re-usable code contains - 
- functions such as
```js
function add(a,b) { return a+b;}
```
- vars such as 
```js
var counter = 0;
```
2. Scope of imports
Each import is resolved at the screen level. Code libraries cannot be used to share state across screens or widgets. 
Example
myLib
```js
var counter = 0;
```
Let's say myLib is imported in `Home` and `Profile` screens. The code in the `Home` screen updates the `counter` to 10. The `Profile` screen or any other screen that imported myLib will still get their own independent `counter` as 0. 

3. How are the variables and functions in re-usable libraries resolved in case of naming conflicts
We apply the rules that ES5/HTML uses i.e. the last one wins except when the same library has been included multiple times. 
Example - 
commonUtils
```js
var counter = 0;
function add(a,b) {return a + b;}
```
apiUtils
```js
var counter = 10;
function add(a,b,c) {return a+b+c;}
```

Let's say the screen imports commonUtils and then apiUtils. Value of the counter will be 10 and the add function from apiUtils will be one invoked when anyone calls add in the page. 

Same applies to re-usable widgets as well. Code libraries at the screen level are automatically inherited by all the widgets in the page. Widgets can have their own libraries as well. If widget imports the same library that is already imported in the screen it resides in, it is *not* imported again. Any new libraries are imported in the order defined.

5. Importing code is supported in both Screens and Widgets as follows. Currently importing code within a code library is not supported. It is being tracked here - #1142 

Importing Code in a screen - 

```
Import:
  - apiUtils
  - common

View:
    # Optional - style the screen
    styles:
      scrollableView: true

    # Optional - set the header for the screen
    header:
      title: Home
    onLoad: |
      saveName('Khurram','Mahmood');
      console.log('name is ='+storageName);

    # Specify the body of the screen
    body:
      Column:
        styles:
          padding: 24
          gap: 8
        children:
          - Text:
              text: Hello ${ensemble.storage.jslibtest.name.first} ${ensemble.storage.jslibtest.name.last}
```
Importing in a re-usable widget - 

```
Import:
  - common
  - apiUtils
Widget:
    inputs:
      - inputName
    body:
      Column:
        children:
          - Text:
              text: ${inputName.first} ${inputName.last}
[...]

PR #1144, opened 2024-01-14
Reusable code libs 62

kmahmood74:
see detailed description in https://github.com/EnsembleUI/ensemble/issues/62

### DIFF
--- a/lib/framework/scope.dart
+++ b/lib/framework/scope.dart
@@ -37,6 +37,7 @@ class ScopeManager extends IsScopeManager with ViewBuilder, PageBindingManager {
   final PageData pageData;
   ScopeManager? _parent;
 
+
   // TODO: have proper root scope
   RootScope rootScope = Ensemble().rootScope();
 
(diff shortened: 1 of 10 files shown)


kmahmood74:
see detailed description in #62

lib/framework/scope.dart
```diff
@@ -151,11 +152,13 @@ class ScopeManager extends IsScopeManager with ViewBuilder, PageBindingManager {
...
   }
 
-  /// create a copy of the parent's data scope
+  /// create a copy of the parent's data scope. the initialMap is generally coming from the imports
   @override
-  ScopeManager createChildScope({bool ephemeral = false}) {
+  ScopeManager createChildScope({bool ephemeral = false, List<ParsedCode>? importedCode}) {
     ScopeManager childScope =
-        ScopeManager(dataContext.clone(), pageData, ephemeral: ephemeral);
+        ScopeManager(dataContext
```
vusters: So evalImports makes sense at the screen level as we want the screens to have separate context, but within a screen I would want my widgets, apis, item-templates,... to use the same code context. This is how the screen's Code block works today. createChildScope() is used for these purposes.

kmahmood74: I am not following you. A re-usable custom widget may have its own imports. So when we evaluate custom widget, we evaluate its imports as well.

vusters: talked to Khurram. This is not what I understood. All good

vusters (approved):

Issue #1135, opened 2024-01-11
Image | pinch to zoom


sharjeelyunus:
there should be a property in image `pinchToZoom: true/false` allowing you to zoom in the images via pinch and on release it should get back to normal state

see this: https://pub.dev/packages/pinch_zoom_release_unzoom

sharjeelyunus:
actually these should be an option for on release, if true go back to normal state, if false maintain the zoomed-in position

PR #1151, opened 2024-01-17
image enhance pinch to zoom

snehmehta:
## Describe the change
- New feature added to image, `pinchToZoom`: true | false, as the name suggest allow image to be zoomed in by pinching.

## Issue number
closes: https://github.com/EnsembleUI/ensemble/issues/1135

## EDL
```yaml
Image:
  source: https://i.gr-assets.com/images/S/compressed.photo.goodreads.com/hostedimages/1578239390i/28715127._SX540_.jpg
  pinchToZoom: true

```

### DIFF
--- a/lib/widget/image.dart
+++ b/lib/widget/image.dart
@@ -1,4 +1,3 @@
-import 'dart:convert';
 import 'dart:io';
 
 import 'package:cached_network_image/cached_network_image.dart';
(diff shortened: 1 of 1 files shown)


snehmehta:
The implementation is done using InteractiveViewer, here the InteractiveViewer can adds interactivity like pinch to zoom on any widget. So, this implementation can be extracted out and add like visibility where any with can be set to have pinch to zoom interactivity or similar interaction.

Issue #1158, opened 2024-01-22
Background of AppIcon in the app switcher is always default to pink


hemish11:
The background color of the app icon when the app is in app switching for Android 13 and above always defaults to pink. It doesn't adapt according to the Ensemble theme.

Here are the screenshots for the same:

![Screenshot_20240115-1150122 (1)](https://github.com/EnsembleUI/ensemble/assets/46677360/041159fc-cbbb-43b1-99be-7fdc532d73a7)

![Screenshot_1705833317](https://github.com/EnsembleUI/ensemble/assets/46677360/0453b42c-8612-4470-a374-cacb94b3e925)

PR #1159, opened 2024-01-22
Fixed the color bug for the appIcon background

hemish11:
Closes https://github.com/EnsembleUI/ensemble/issues/1158


snehmehta (approved):
so some issue with `copyWith`?

hemish11:
Yes @snehmehta, so I was able to debug that the background color for the logo was not getting updated only when the seed color was passed using copyWith method. We can pass by all other ways but copyWith method doesn't work. This behavior should be a bug in the Flutter framework, so I just changed the code to not use copyWith.

Issue #1154, opened 2024-01-17
Switch | Intermediate State


sharjeelyunus:
By default there should be only 2 states, but if user configures for 3, Switch should have 3 states, like below.

here we have data coming from the api, and we select 2 or more dates, if availablity is open then we keep it false, if its closed then true, and if its true and false at the same time on different select dates (mixed availability) then we want to show switch in between.

In between state should also have special styling properties like mentioned bellow

![Screenshot_2024-01-18-00-56-31-426-edit_com figma mirror](https://github.com/EnsembleUI/ensemble/assets/61178058/6db5f6db-cc71-4cbb-b026-335d4e7032a1)

sharjeelyunus:
![image](https://github.com/EnsembleUI/ensemble/assets/61178058/2a1403ec-7a0b-4e4c-9e21-9fae1c7e0617)

sharjeelyunus:
![image](https://github.com/EnsembleUI/ensemble/assets/61178058/b4fb4bda-a0a0-422a-9d63-49b1475a8115)

vinothvino42:
PR - #1160

PR #1160, opened 2024-01-22
Add triple state switch widget

vinothvino42:
Ticket: #1154

`value = off | mixed | on`

Doc - https://github.com/EnsembleUI/ensemble_docs/pull/123

**Sample EDL** 

```yaml
View:
  header:
    title: Triple Switch
  body:
    Column:
      styles: { gap: 24, padding: 24, scrollable: true }
      children:
        - TripleSwitch:
            leadingText: "Entire Place "
            value: mixed
            enabled: true
            styles:
              {
                activeThumbColor: white,
                inactiveThumbColor: grey,
                activeColor: 0xFFD6D6CE,
                inactiveColor: 0xFF425C59,
                mixedColor: 0xFF858174,
              }
            onChange: |
              //@code
              console.log("Switch Value: "+this.value);

```

https://github.com/EnsembleUI/ensemble/assets/12869588/3c88ebde-4e0e-46d0-b222-f9e9f7c3c3ed


assets/schema/ensemble_schema.json
```diff
@@ -1849,6 +1849,35 @@
       ]
     },
     "Switch-payload": {
+      "type": "object",
```
snehmehta, line 1852: hey vinoth, the schema needs to add in ensemble-web-studio ts files, check the studio repo readme

lib/widget/triple_state_switch.dart
```diff
@@ -0,0 +1,296 @@
...
+
+  framework.EnsembleAction? onChange;
+}
+
+class EnsembleTripleStateSwitchState
+    extends FormFieldWidgetState<EnsembleTripleStateSwitch> {
+  void onToggle(SwitchState newValue) {
+    widget.onToggle(newValue);
+    widget._controller.value = newValue;
+    if (widget._controller.onChange != null) {
```
snehmehta, line 83: add a dispatch event here, so if any listener is bind to switch state change will update as well. 
Something like this 
https://github.com/EnsembleUI/ensemble/blob/afa3e8dcfa0eb0e6fbc16a783f03426b2b77b71f/lib/layout/tab_bar.dart#L160-L161

lib/widget/triple_state_switch.dart
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    // wraps around FormField to get all the form effects
+    return InputWrapper(
+      type: EnsembleTripleStateSwitch.type,
+      controller: widget._controller,
+      widget: FormField<bool>(
+        key: validatorKey,
+        validator: (value) {
+          if (widget._controller.required) {
```
snehmehta, line 118: do test this validation case, from the look of it won't work as expected

snehmehta (approved):

lib/widget/triple_state_switch.dart
```diff
@@ -0,0 +1,296 @@
...
+    return {'value': () => _controller.value.name};
+  }
+
+  @override
+  Map<String, Function> methods() {
+    return {};
+  }
+
+  @override
+  Map<String, Function> setters() {
```
vusters, line 37: why copy and pasted all these code? 99% of it identical to the regular switch. You should at least use a base class instead of replicating all these.

lib/widget/triple_state_switch.dart
```diff
@@ -0,0 +1,296 @@
...
+    return {
+      'value': (value) => _controller.value =
+          SwitchState.values.from(Utils.getString(value, fallback: 'start')) ??
```
vusters, line 40: You have two "start" values here plus the default "start" - that is 3 different places to change if you decide to change the default state. Potential cause for bug unnecessary.
Use this: 
`_controller.value = SwitchState.values.from(value) ?? _controller.value`

vinothvino42: Done

lib/widget/triple_state_switch.dart
```diff
@@ -0,0 +1,296 @@
...
+      middleBackgroundColor: widget._controller.intermediateColor,
+      endBackgroundColor: widget._controller.inactiveColor,
+      disable: widget._controller.enabled == false,
+      state: widget._controller.value,
+      onChanged: isEnabled() ? (value) => onToggle(value) : (_) {},
+    );
+  }
+}
+
+enum SwitchState {
```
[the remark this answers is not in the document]
vinothvino42: Changed it

lib/widget/triple_state_switch.dart
vinothvino42: Refactored

lib/widget/triple_state_switch.dart
```diff
@@ -0,0 +1,296 @@
...
+import 'package:ensemble/framework/extensions.dart';
+import 'package:ensemble/screen_controller.dart';
+import 'package:ensemble/util/utils.dart';
+import 'package:ensemble/widget/helpers/widgets.dart';
+import 'package:ensemble/widget/input/form_helper.dart';
+import 'package:ensemble/framework/action.dart' as framework;
+import 'package:ensemble_ts_interpreter/invokables/invokable.dart';
+import 'package:flutter/material.dart';
+
+class EnsembleTripleStateSwitch extends StatefulWidget
```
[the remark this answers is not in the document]
vinothvino42: Now, I'm extending the FormFieldController for the TripleStateSwitchController. Currently, the FormFieldController is not migrated to the new way (extending the EnsembleWidgetController instead of WidgetController). I tried changing the FormFieldController so I am supposed to migrate all the widget which uses the FormFieldController to new way of building widgets (EnsembleWidget<Controller>).

So as of now I done in the old way only

assets/schema/ensemble_schema.json
vinothvino42: Okay I'll update in the studio once this PR is merged

lib/widget/triple_state_switch.dart
snehmehta: Yes, my bad. It works. For some reason i thought the case of tabbar and footer not updating on tab change because of event but I was wrong it update I check my theory.

lib/widget/switch.dart
```diff
@@ -32,15 +42,16 @@ class EnsembleSwitch extends StatefulWidget
...
-          _controller.value = Utils.getBool(value, fallback: false),
+      'value': (value) => setSwitchValue(value),
       'leadingText': (text) =>
           _controller.leadingText = Utils.optionalString(text),
       'trailingText': (text) =>
           _controller.trailingText = Utils.optionalString(text),
       'activeColor': (color) => _controller.activeColor = Utils.getColor(color),
       'inactiveColor': (color) =>
           _controller.inactiveColor = Utils.getColor(color),
+      'intermediateColor': (color) =>
```
vusters, line 53: mixedColor

lib/widget/switch.dart
```diff
@@ -50,28 +61,30 @@ class EnsembleSwitch extends StatefulWidget
     };
   }
 
-  void onToggle(bool newValue) {
-    setProperty('value', newValue);
+  void setSwitchValue(dynamic value) {
```
vusters: Switch and TripleSwitch's value should be succinct. The former takes in true/false, the latter 3 states. There should be no guessing like this, so please refactor so their inputs are separate. The schema should reflect tis too

lib/widget/switch.dart
```diff
@@ -50,28 +61,30 @@ class EnsembleSwitch extends StatefulWidget
...
-  bool value = false;
-  String? leadingText;
-  String? trailingText;
-  Color? activeColor;
-  Color? activeThumbColor;
-  Color? inactiveColor;
-  Color? inactiveThumbColor;
 
-  framework.EnsembleAction? onChange;
+  void onToggle(dynamic newValue) {
```
vusters, line 64: same thing here. Each widget should have its own onToggle

lib/widget/switch.dart
```diff
@@ -102,12 +115,13 @@ class SwitchState extends FormFieldWidgetState<EnsembleSwitch> {
 
     // wraps around FormField to get all the form effects
     return InputWrapper(
-      type: EnsembleSwitch.type,
+      // type: EnsembleTripleStateSwitch.type,
+      type: '',
```
vusters: put the type in. This is to help with exceptions.

lib/widget/switch.dart
vinothvino42: Changed it

lib/widget/switch.dart
vinothvino42: Added `onToggle` in those widgets with their state as input parameter

lib/widget/switch.dart
vinothvino42: Updated it

lib/widget/switch.dart
vinothvino42: Done

vinothvino42:
@vusters Can you review it? Just updated it, also checked the validation and it's working fine

- Moved the `onToggle` to Switch and TripleSwitch with their input as parameter and updated the logic in the SwitchBase class for value-changing
- Changed the widget name from `TripleStateSwitch` to `TripleSwitch`
- Added the `type` to the InputWrapper widget
- Changed the property from `intermediateColor` to `mixedColor`

vusters (approved):

Issue #1166, opened 2024-01-24
Notifications | Screen Name is compulsory???


sharjeelyunus:
I'm getting this as notification payload

```
{
    “conversation_id”,
    “guest_id”,
    “host_id”,
    “property_id”,
    “property_city”,
    “notification_type”
}
```

and based on notification type I want to navigate the user to a screen. apparently, this isn't possible, it won't even accept the notification, in console I'm getting: `No screenId nor screenName provided on the notification. Ignoring...`

this shouldn't be the case, When I click on a notification, it should open the app and take me to the default "Home Screen", and from there I can navigate the user to whichever screen I want based on notification_type.

vusters:
on notification, if screenId nor screenName is specified, takes the user to the home screen (don't just open the app) with the following payload:
```json
notificationPayload: {
  notification: {
    // .. notification block
  }
  data: {
    // .. data block
  }
  ..// other block
}
The home screen can access it by:
```
View:
  onLoad: |
    // has notification
    if (notificationPayload != null) {
      
    }
    ```

vinothvino42:
@vusters Can you review it? Here's the PR - #1169

vusters:
Fixed under https://github.com/EnsembleUI/ensemble/pull/1175.
@sharjeelyunus Please disregard the above. The notification payload will be:
```
notificationPayload: {
  title: "optional title",
  body: "optional body",
  data: {
    // data if sent by the server. Empty otherwise
  }
}
```
You can do this on the screen that handles notification:
```
View:
  onLoad: |
    if (notificationPayload != null) {
      // process notification
    } else {
        // regular page logic
    }
```

PR #1169, opened 2024-01-25
WIP: Add process notification action and update handle notification

vinothvino42:
Ticket: #1166
In this PR, I updated the following

1. Created a new notification action called `ProcessNotificationAction`. Which will return the terminated notification object if it's not null or else return the `Ensemble.externalDataContext` which contains the `notificationPayload` object. User can able to get the data using notificationPayload key (it contains the title, body and data of the notification)
2. Updated the firebase message listeners (onMessage, onMessageOpened, etc) to store the payload into the `Ensemble.externalDataContext`

Sample EDL:
```yaml
View:
  styles:
    useSafeArea: true
  header:
    title: Push Notification Test
  onLoad: |
    //@code
    ensemble.storage.payloadData = "Not Found";
    if (notificationPayload != null) {
      ensemble.storage.payloadData = notificationPayload;
    }
  body:
    Column:
      styles:
        gap: 20
        padding: 20
        margin: 20
      children:
        - Button:
            label: Print Device Token
            styles:
              backgroundColor: blue
              borderRadius: 8
            onTap:
              getDeviceToken:
                onSuccess: |-
                  console.log("Device Token: "+event.data.token);

        - Divider

        - Button:
            label: Get Notification Payload
            styles:
              backgroundColor: blue
              borderRadius: 8
            onTap: |
              console.log("Get Notification");
              if (notificationPayload != null) {
                console.log(notificationPayload);
              }

        - Markdown:
            text: ${ensemble.storage.payloadData}

        - Divider
        - Button:
            label: PROCESS Notification Action
            onTap:
              processNotification:
                onNotification: |
                  console.log("processNotification - Notification Payload");
                  console.log(event.data.notificationPayload);
                onNoNotification: |
                  console.log("Failed to process notification");

```

### DIFF
--- a/lib/framework/data_context.dart
+++ b/lib/framework/data_context.dart
@@ -274,6 +274,10 @@ class DataContext {
       [SourceLocation? startLoc]) {
     try {
       _contextMap['getStringValue'] = Utils.optionalString;
+      // TODO: Added temporarily because the externalDataContext datas are not exposed when it's inside the nested callback methods
+      // Ex: Not working when this (widget.dataContext.addDataContext({'messageText': 'Data - EnsembleUI'});) inside the
+      // addPostFrameCallback method in screen.dart
+      _contextMap.addAll(Ensemble.externalDataContext);
       return JSInterpreter(codeBlock, program, _contextMap).evaluate();
     } on JSException catch (e) {
       if (e.detailedError is EnsembleError) {
--- a/lib/framework/notification_manager.dart
+++ b/lib/framework/notification_manager.dart
@@ -37,6 +37,11 @@ class NotificationManager {
     }
   }
 
+  // Returns the remote notification only when the app is opened from terminated state otherwise it'll be null
+  Future<RemoteMessage?> getInitialMessage() {
+    return FirebaseMessaging.instance.getInitialMessage();
+  }
+
   /// get the device token. This guarantees the token (if available)
   /// is the latest correct token
   Future<String?> getDeviceToken() async {
@@ -101,8 +106,8 @@ class NotificationManager {
   }
 
   void initGetInitialMessage() {
-    // This is called when the user taps on the notification and the app is opened from the terminated state
-    FirebaseMessaging.instance.getInitialMessage().then((message) {
+    // Returns the remote notification only when the app is opened from the terminated state or else it'll be null
+    getInitialMessage().then((message) {
       if (message == null) return;
 
       Ensemble.externalDataContext.addAll({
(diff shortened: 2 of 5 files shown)


kmahmood74:
@vinothvino42 @vusters I don't get it, when would someone invoke a `processNotification` onTap of a button? I thought this action would be invoked by the platform when a notification arrives. 

It makes sense for user to register an action handler for this action outside of any screens, at the app level just like we have app settings etc.

lib/framework/notification_manager.dart
```diff
@@ -120,17 +125,15 @@ class NotificationManager {
...
+        messageData?['screenId'] == null && messageData?['screenName'] == null;
+    ScreenController().navigateToScreen(Utils.globalAppKey.currentContext!,
         screenId: messageData!['screenId'],
         screenName: messageData['screenName'],
         pageArgs: messageData,
-      );
-    } else {
-      log('No screenId nor screenName provided on the notification. Ignoring ...');
-    }
+        routeOption: screenNotFound ? RouteOption.clearAllScreens : null);
```
vusters, line 143: i know we talked about this but it's probably not safe to clear all the screens. It's OK to have duplicate screens

lib/framework/data_context.dart
```diff
@@ -274,6 +274,10 @@ class DataContext {
       [SourceLocation? startLoc]) {
     try {
       _contextMap['getStringValue'] = Utils.optionalString;
+      // TODO: Added temporarily because the externalDataContext datas are not exposed when it's inside the nested callback methods
+      // Ex: Not working when this (widget.dataContext.addDataContext({'messageText': 'Data - EnsembleUI'});) inside the
+      // addPostFrameCallback method in screen.dart
+      _contextMap.addAll(Ensemble.externalDataContext);
```
vusters, line 280: This is wrong. I don't understand why you need this. Also notification payload should be sent only ONCE and ONE ONLY, when the notification was delivered (or when you click on it).

lib/framework/notification_manager.dart
```diff
@@ -124,7 +130,8 @@ class NotificationManager {
   }
 
   Future<void> _handleNotification() async {
-    Map<String, dynamic>? messageData = Ensemble.externalDataContext['data'];
+    Map<String, dynamic>? messageData =
```
vusters, line 133: why do you have to read the data from externalDataContext? What is externalDataContext? The RemoteMessage is already there.

vusters:
@vinothvino42 please abandon this review. There some serious issues with this when the fix is very simple. See it https://github.com/EnsembleUI/ensemble/pull/1175/commits/aa69419668a00f37e8e1ee9e100bfa7dd11f990a

Issue #1186, opened 2024-01-31
Dropdown Bug


Ahsen6:
Once an auto-complete's drop down's value is selected, it remains there even if i delete/remove the text. I encountered this issue while attempting to make a button enabled only if the user had selected a certain dropdown value, and disabled otherwise. But removing the value after initially selecting it doesn't disable the button. Works fine for a regular dropdown but not for dropdown with autoComplete set to true. 

Deleting the text that pertains to the dropdown value should delete the value. Here's the sample yaml:

``` 

- Dropdown:
      hintText: For e.g poster design
      id: autoComplete
      autoComplete: true
      items: ${getAllCategories()}
      label: Gig category
      styles:
         fillColor: transparent

vinothvino42:
PR - https://github.com/EnsembleUI/ensemble/pull/1188

PR #1188, opened 2024-02-01
Trigger value property when dropdown textfield value changes

vinothvino42:
Ticket: [#1186](https://github.com/EnsembleUI/ensemble/issues/1186)

Sample EDL:

```yaml
View:
  header:
    title: Dropdown
  styles:
    scrollableView: true
  onLoad:
    invokeAPI:
      name: getPeople
      onResponse: |
        //@code
        var list = [];
        response.body.map(function(item) {
          list.push(item.name);
        });
        dropdownId.items = list;

  body:
    Column:
      styles:
        gap: 30
        padding: 24

      children:
        - Markdown:
            text: |
              ## Autocomplete

        - Conditional:
            conditions:
              - if: ${dropdownId.value == 'Patricia Lebsack'}
                Button:
                  label: Patricia Lebsack - Selected
              - else:
                Text:
                  text: Patricia Lebsack - Not Selected

        - Dropdown:
            hintText: Select the person
            id: dropdownId
            autoComplete: true
            onChange: |
              //@code
              console.log("Autocomplete Data: "+dropdownId.value);

API:
  getPeople:
    uri: https://jsonplaceholder.typicode.com/users
    method: GET

```

### DIFF
--- a/lib/widget/input/dropdown.dart
+++ b/lib/widget/input/dropdown.dart
@@ -387,13 +387,21 @@ class SelectOneState extends FormFieldWidgetState<SelectOne>
                   FocusNode fieldFocusNode,
                   VoidCallback onFieldSubmitted) {
                 return TextField(
-                    enabled: isEnabled(),
-                    showCursor: true,
-                    style: const TextStyle(
-                        color: Colors.black, fontWeight: FontWeight.w500),
-                    controller: fieldTextEditingController,
-                    focusNode: fieldFocusNode,
-                    decoration: inputDecoration);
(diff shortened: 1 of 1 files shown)


snehmehta (approved):
May be someday we may need to add debouncer here to prevent from abuse of setProperty notifier on frequent change.

vinothvino42:
> May be someday we may need to add debouncer here to prevent from abuse of setProperty notifier on frequent change.

Yes. We've to do it for the root form input field controller itself

Issue #1184, opened 2024-01-31
Enable reveal (eye icon) and ability to copy on Password field that is disabled


amin-nas:
When password field is disabled, I should still be able to click the eye icon to view and COPY the value.

vinothvino42:
PR - #1202

PR #1202, opened 2024-02-06
Add selectable and readOnly property to text input field

vinothvino42:
Ticket: [#1184](https://github.com/EnsembleUI/ensemble/issues/1184)

It applicable for both TextInput and PasswordInput widgets

Doc - https://github.com/EnsembleUI/ensemble_docs/pull/143

Schema (BaseTextInput) - https://github.com/EnsembleUI/ensemble-web-studio/pull/711

**Sample EDL**
```yaml
PasswordInput:
  label: Password
  value: EnsembleUI
  obscureToggle: true
  readOnly: true    # If true, the text input is to be read-only, text cannot be modified
  selectable: true     # If true, It will enable features such as long-pressing the TextField to select text and show the cut/copy/paste menu, and tapping to move the text caret text.   
```

### DIFF
--- a/lib/widget/input/form_textfield.dart
+++ b/lib/widget/input/form_textfield.dart
@@ -140,6 +140,10 @@ abstract class BaseTextInput extends StatefulWidget
           _controller.enableClearText = Utils.optionalBool(value),
       'obscureToggle': (value) =>
           _controller.obscureToggle = Utils.optionalBool(value),
+      'readOnly': (value) =>
+          _controller.readOnly = Utils.getBool(value, fallback: false),
+      'selectable': (value) =>
+          _controller.selectable = Utils.optionalBool(value),
       'toolbarDone': (value) =>
           _controller.toolbarDoneButton = Utils.optionalBool(value),
       'keyboardAction': (value) =>
@@ -213,7 +217,8 @@ class TextInputController extends FormFieldController {
 
   // applicable only for Password or obscure TextInput, to toggle between plain and secure text
   bool? obscureToggle;
-
+  bool readOnly = false;
+  bool? selectable;
   bool? toolbarDoneButton;
 
   model.InputValidator? validator;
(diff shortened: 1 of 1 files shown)


snehmehta (approved):

amin-nas:
@vinothvino42 @snehmehta we already have a `enabled` field for inputs such as this which makes the field, read only. why do we need a new property called `readOnly`? how's that different than `enabled: false`

vinothvino42:
> @vinothvino42 @snehmehta we already have a `enabled` field for inputs such as this which makes the field, read only. why do we need a new property called `readOnly`? how's that different than `enabled: false`

The `enabled` property doesn't allow the property to have an onTap event, it completely makes the field non-interactive one. But with `readOnly` property in the input field, you can do the following

1. Toggle the password visibility on/off
2. If on, you can able to select and copy the text but you can't enter any text
3. Also the input field will have outline color when you tap the field

amin-nas:
@vinothvino42 thanks for the details. I get it, but still find a bit odd to have all these props. For instance, what happens if I do both. I assume enabled supersede the other behaviors?

```yaml
enabled: false
readOnly: true
```

If we do want these, I suggest we apply it to other inputs as well. I'll create a ticket for that once others chime in with their thoughts. @vusters @kmahmood74 what do you think?

Also tagging @evshi as these props should work and behave the same way on React side.

vinothvino42:
> @vinothvino42 thanks for the details. I get it, but still find a bit odd to have all these props. For instance, what happens if I do both. I assume enabled supersede the other behaviors?
> 
> ```yaml
> enabled: false
> readOnly: true
> ```
> 
> If we do want these, I suggest we apply it to other inputs as well. I'll create a ticket for that once others chime in with their thoughts. @vusters @kmahmood74 what do you think?
> 
> Also tagging @evshi as these props should work and behave the same way on React side.

In this case, `enabled` will be the priority one and the `readOnly` will be ignored

PR #1217, opened 2024-02-11
https://github.com/EnsembleUI/ensemble_live/issues/48 providing live …

kmahmood74:
…context when debug == true

see it with https://github.com/EnsembleUI/ensemble_live/pull/49 to make sense of it


kmahmood74:
@vusters this is the first checkin to get it going. I still need to add `actions` with their method signatures and add a lot more getters to widgets including nested styles. Once this infra piece is done, we'll keep adding more context independent of the live app or the studio

Issue #1220, opened 2024-02-11
Add readyOnly and selectable prop for free-form input field


amin-nas:
For these widgets, i want to be able to set readOnly and selectable:

1. TextInput
2. PasswordInput
3. Form

Acceptance criteria:
1. Default values are `readOnly: false` and `selectable: true`
2. When I only set `readOnly: true`, and do not set enabled or selectable props, user should be able to select the value and copy it. User should not be able to modify the value.
3. When I only set `enabled: fase`, widget should ignore the `readOnly` prop. The widget will not be clickable at all.
4. When I set `readOnly` at the parent Form widget, the value must cascade to children of the Form that are TextInputs and PasswordInputs.

vinothvino42:
PR - #1234

vusters:
completed.

PR #1234, opened 2024-02-13
Update input field readOnly and selectable

vinothvino42:
Ticket: [#1220](https://github.com/EnsembleUI/ensemble/issues/1220)

Acceptance criteria:
1. Default values are `readOnly: false` and `selectable: true`
2. When I only set `readOnly: true`, and do not set enabled or selectable props, user should be able to select the value and copy it. User should not be able to modify the value.
3. When I only set `enabled: fase`, widget should ignore the `readOnly` prop. The widget will not be clickable at all.
4. When I set `readOnly` at the parent Form widget, the value must cascade to children of the Form that are TextInputs and PasswordInputs.

Sample EDL:

```yaml
View:
  header:
    title: Form
  styles:
    scrollableView: true

  body:
    Column:
      styles:
        gap: 16
        padding: 24
      children:
        - Form:
            styles:
              readOnly: true
            children:
              - TextInput:
                  id: name
                  label: Name
                  value: EnsembleUI
                  required: true
                  readOnly: false
                  # enabled: false

              - PasswordInput:
                  label: Password
                  value: ABCD
                  obscureToggle: true
                  # readOnly: false
                  # enabled: false

```


lib/widget/input/form_textfield.dart
```diff
@@ -217,7 +216,7 @@ class TextInputController extends FormFieldController {
 
   // applicable only for Password or obscure TextInput, to toggle between plain and secure text
   bool? obscureToggle;
-  bool readOnly = false;
+  bool? readOnly;
   bool selectable = true;
```
vusters, line 220: Why is this still here?

lib/layout/form.dart
```diff
@@ -158,6 +161,15 @@ class FormState extends WidgetState<EnsembleForm>
     for (Widget formItem in formItems) {
       if (formItem is HasController &&
           formItem.controller is FormFieldController) {
+        // Adding Form's readOnly value to the TextInput and PasswordInput
```
vusters, line 164: This is wrong @vinothvino42 @snehmehta. The Form should have zero idea what a TextInput is, nor trying to reach inside its children and modify the child's data.

vusters: Instead inside the child (i.g. TextInput) it should reach out to its parent Form and look at its attribute instead.

vusters (changes requested):

lib/widget/input/form_textfield.dart
vinothvino42: Actually, both **TextInput** and **PasswordInput** field's controller extends from (`TextInputController -> FormFieldController -> WidgetController`)
But the Form widget's controller extends from (`FormController -> WidgetController`)
So added the `readOnly` property in two places (`TextInputController` and `FormController`)

`Form's controller's` `readOnly` property is to set all the `TextInput` and `PasswordInput` field's `readOnly` property in the **Form** widget

TextInput and PasswordInput's controller's readOnly property is used for `inline styling`, to `override` the parent widget (Form controller's readOnly property)

lib/layout/form.dart
vinothvino42: Refactored it. Now I accessed the parent Form's readOnly property from the child widget (BaseTextInput's state class) and assigned the value to it if the BaseTextInput controller's readOnly property is null

vinothvino42:
@vusters Can you review it again?

lib/widget/input/form_textfield.dart
```diff
@@ -317,6 +316,14 @@ class TextInputState extends FormFieldWidgetState<BaseTextInput>
         }
       }
     });
+
+    // Checking for readOnly from parent widget and assign the value to TextInput and PasswordInput if it's readOnly property is null
+    final formController =
```
vusters, line 321: Dont' waste unnecessary cycles. There's no need to traverse the tree unless widget._controller.readOnly is not set.

lib/widget/input/form_textfield.dart
```diff
@@ -317,6 +316,14 @@ class TextInputState extends FormFieldWidgetState<BaseTextInput>
...
+        context.findAncestorWidgetOfExactType<EnsembleForm>()?.controller;
+
+    if (formController != null && widget._controller.readOnly == null) {
+      widget._controller.readOnly = formController.readOnly == true;
```
vusters, line 325: You are inheriting this property from Form, so it's not proper to overwrite it. This cause unnecessary bugs if the user is reading it back, and also waste another rendering cycle. You should simple do a getter `isReadOnly() => isReadOnly ?? form.isReadOnly`

vusters (changes requested):
It's important to remove attributes you added inadvertently.

lib/widget/input/form_textfield.dart
vinothvino42: Why? `selectable` property also needed right. User can control (selectable: true or false) it anytime based on some condition

lib/widget/input/form_textfield.dart
vinothvino42: How can I get EnsembleForm widget's `formController` to get the readOnly property?

Our TextInput and PasswordInput field's controller extends from (`TextInputController -> FormFieldController -> WidgetController`)
But the Form widget's controller extends from (`FormController -> WidgetController`).

Now I want to get the `Form widget's readOnly` property and to set that parent value (Form's readOnly) to the children's (TextInput or PasswordInput) field's `readOnly` property

lib/widget/input/form_textfield.dart
vinothvino42: Updated it

vusters (approved):

Issue #1218, opened 2024-02-11
Ensemble Live in Web/Studio - top two tiles in kitchen sink app are not clickable


kmahmood74:
1. login to the studio
2. go here - https://studio.ensembleui.com/app/e24402cb-75e2-404c-866c-29e6c3dd7992/screen/nAvv0XLuq92N82qpKPCI
3. Pick any option like Form Widgets or Layout widgets
4. top two tiles are not clickable, rest work fine

Same issue when running kitchen sink separately in Ensemble Live on the web. I think this has been an issue for a while

cc @amin-nas @vusters

vinothvino42:
This is fixed now. It is happening only when the build is built with HTML renderer, its working fine with Canvaskit renderer

 It's an issue caused by the Lottie widget for the web, the HTMLElementView widget is used for it which is not allowing other views to receive the mouse scroll event. 

It can be fixed with a package (https://pub.dev/packages/pointer_interceptor). Already we have a style property called `captureWebPointer` boolean property for the widgets. Now I added it for the page styles too, here's the PR for it - https://github.com/EnsembleUI/ensemble/pull/1238

kmahmood74:
@vinothvino42 why are you marking it as released when the PR is still open? This is not released

kmahmood74:
@vinothvino42 your explanation is hard to understand. 

1. Why did this just start happening when Lottie has been there for at least a year if not more for html renderer?
2. why is it just blocking the top two tiles?
3. there is no lottie on those pages?

We can't fix a new regression by adding another flutter package. You have to investigate the issue properly. We can't put band-aids like this. 

cc @snehmehta

vinothvino42:
> @vinothvino42 your explanation is hard to understand.
> 
> 1. Why did this just start happening when Lottie has been there for at least a year if not more for html renderer?
> 2. why is it just blocking the top two tiles?
> 3. there is no lottie on those pages?
> 
> We can't fix a new regression by adding another flutter package. You have to investigate the issue properly. We can't put band-aids like this.
> 
> cc @snehmehta

@kmahmood74 First I need to tell you that I didn't implemented this package, we're already having it https://github.com/EnsembleUI/ensemble/blob/1978c082248a2a017756317a4eaa523ced386a92/pubspec.yaml#L98
[...]

PR #1238, opened 2024-02-14
Add captureWebPointer for page widget styles

vinothvino42:
#1218 is the issue
This allows to get mouse events if the view is on top of the HTMLElementView


vinothvino42:
It's not needed, so I am closing it. 
Another PR that solves the issue in that Lottie widget itself - #1240

Issue #1210, opened 2024-02-08
Blank definition should render a blank screen instead of error


vusters:
A blank definition could be a valid start state, and so Runtime should render a blank screen instead of showing an error today.

![Image](https://github.com/EnsembleUI/ensemble/assets/622269/510cee56-3d3a-4274-83be-4a9c5bd36534)

vinothvino42:
WIP PR - #1239

vusters:
Completed

PR #1239, opened 2024-02-14
Show empty view instead of throwing an error when the EDL is empty

vinothvino42:
Ticket: [#1210](https://github.com/EnsembleUI/ensemble/issues/1210)


snehmehta (approved):

lib/ensemble_provider.dart
```diff
@@ -60,8 +60,9 @@ class EnsembleDefinitionProvider extends DefinitionProvider {
     }
 
     if (content == null) {
-      throw LanguageError(
-          "Invalid screen content: ${screenId ?? screenName ?? 'Home'}");
+      throw LanguageError('invalid-content',
```
vusters, line 63: We are saying this is NOT an error condition, so we should NOT be throwing error. Throwing an error can cause all sort of unknown and un-traceable issues.

You should refactor the code such that empty content is accepted by the system.

vusters (changes requested):

lib/ensemble_provider.dart
vinothvino42: @vusters I updated it, can you review it?

snehmehta (approved):

vusters (approved):

PR #1240, opened 2024-02-15
Fix lottie web widget not to capture other view input events

vinothvino42:
https://github.com/EnsembleUI/ensemble/issues/1218

It fixes two issues

1. This allows to get mouse/input events for other views on top of it or on the next page in the region of the Lottie widget on the previous page

2. Also it fixes the onTap issue with HTML Renderer also


snehmehta (approved):

vusters (approved):

vinothvino42:
> Please make sure to remove all the hacks from the Kitchen Sink code before merging this in.

I added only the captureWebPointer to all the custom widgets and removed that already

Issue #1247, opened 2024-02-17
Context issues with global variables and invokeAPI calls


kmahmood74:
This combines - 
#1243
and
#1226

PR #1248, opened 2024-02-17
https://github.com/EnsembleUI/ensemble/issues/1247 fixes the global v…

kmahmood74:
…ar and API issues with context.


lib/framework/data_context.dart
```diff
@@ -50,15 +51,24 @@ import 'package:jsparser/jsparser.dart';
 
 /// manages Data and Invokables within the current data scope.
 /// This class can evaluate expressions based on the data scope
-class DataContext {
+class DataContext implements Context {
+  static const String globalContextKey = '__internal__parent__context__';
```
kmahmood74, line 55: I created the parent context as a key inside the context. Not my ideal solution but had to do it to make sure that our copy operations copy over the parent context as well

vusters (approved):
Looks fine to me. Good luck :)

PR #1272, opened 2024-02-27
Added semantics label to widgets

hemish11:
Closes https://github.com/EnsembleUI/ensemble/issues/1308


lib/widget/helpers/controllers.dart
```diff
@@ -223,6 +225,7 @@ abstract class WidgetController extends Controller with HasStyles {
     return {
       'flexMode': (value) => flexMode = FlexMode.values.from(value),
       'flex': (value) => flex = Utils.optionalInt(value, min: 1),
+      'automationId': (value) => automationId = Utils.optionalString(value),
```
kmahmood74, line 228: please add a getter as well, it's weird to have a setter but not a getter, we have had many issues due to this asymmetry

lib/widget/helpers/controllers.dart
```diff
@@ -398,6 +403,7 @@ abstract class EnsembleWidgetController extends EnsembleController with HasStyle
...
```
kmahmood74, line 406: please add a getter as well, it's weird to have a setter but not a getter, we have had many issues due to this asymmetry

kmahmood74 (changes requested):
have you tested with GridViewListView i.e. repeating containers?

kmahmood74 (changes requested):
@hemish11 we need to do this behind a runtime configuration. Note that semantic labels are used for accessibility. We don't want this code to always take effect as our accessibility case will not work in the future. 

For native apps - pass a runtime argument --testmode when launching the simulator and then read it in your code. If true, only then set the semantic label the way you are doing. 

Merge today please as I need to make some further changes for allowing automation on flutter web. I will also revert to label/text if the widget doesn't have either an id or an automation id

kmahmood74 (approved):
Please merge after making the requested changes and testing containers like GridView and ListView as well

lib/framework/ensemble_widget.dart
```diff
@@ -105,7 +105,8 @@ abstract class EnsembleWidgetState<W extends EnsembleWidget> extends State<W> {
       }
 
       // To enable test mode, we need to add --dart-define="testmode=true"
-      const isTestMode = String.fromEnvironment("testmode") == "true";
+      final envString = const String.fromEnvironment("testmode").toLowerCase();
```
kmahmood74, line 108: environment variables should be retrieved at the load time and stored in a config object, they must not be deep in widget code. That way it's easier to understand and change how env variables are set and the rest of the code only depends on the value of the env variables and not where or how it is set. 

Please move this to EnvConfig and make EnvConfig a singleton. It should be a singleton anyway

lib/framework/widget/widget.dart
```diff
@@ -154,7 +154,8 @@ abstract class WidgetState<W extends HasController> extends BaseWidgetState<W> {
       }
 
       // To enable test mode, we need to add --dart-define="testmode=true"
-      const isTestMode = String.fromEnvironment("testmode") == "true";
+      final envString = const String.fromEnvironment("testmode").toLowerCase();
```
kmahmood74, line 157: same comment here

kmahmood74 (changes requested):

lib/framework/widget/widget.dart
```diff
@@ -26,6 +29,38 @@ mixin UpdatableContainer<T extends Widget> {
 abstract class WidgetState<W extends HasController> extends BaseWidgetState<W> {
   ScopeManager? scopeManager;
 
+  @override
+  void changeState() {
```
kmahmood74, line 33: @hemish11 is this accidental? why all these changes that are not related to this PR?

lib/framework/widget/widget.dart
```diff
@@ -114,43 +153,18 @@ abstract class WidgetState<W extends HasController> extends BaseWidgetState<W> {
         ///    So if we put Expanded on the Column's child, layout exception will occur
         rtn = Expanded(child: rtn);
       }
-    }
-    return rtn;
-  }
 
-  /// build your widget here
```
kmahmood74, line 121: why is all this deleted?

kmahmood74 (changes requested):

kmahmood74 (approved):

kmahmood74:
how to use this?

PR #1280, opened 2024-02-29
improving list view scroll

snehmehta:
I have introduced mixin `AutomaticKeepAliveClientMixin` to every listview element. 
This keeps the element alive even if the item goes out of the viewport.

Reasoning: Our `ListView` item can be very heavy in terms of number of widget it is built width. If we have to built the heavy item we scroll this make scroll janky. So by keep them alive, list view can reuse them.

If we keep item alive even if we go out of view port, this will increase the memory usage as we scroll to very large number of items.

### DIFF
--- a/lib/layout/list_view.dart
+++ b/lib/layout/list_view.dart
@@ -221,6 +221,7 @@ class ListViewState extends WidgetState<ListView>
           }
         }
         if (itemWidget != null) {
+          itemWidget = KeepAliveListItem(itemWidget: itemWidget);
           return widget._controller.onItemTap == null
               ? itemWidget
               : flutter.InkWell(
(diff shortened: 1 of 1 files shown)


snehmehta:
### Ensemble widget tree
https://github.com/EnsembleUI/ensemble/assets/36638657/6d4eece1-9b59-4ba0-9a62-629e039f900f

### Flutter widget tree
https://github.com/EnsembleUI/ensemble/assets/36638657/5bd80b48-c0e3-4c05-82c9-10ce91a692a3

Our ensemble widget tree for similar widget is quite heavy compare to flutter implementation tree. This make quite a difference in scroll performance between two for similar UI.

So, short term solution is in this PR reusing the already built heavy widget by keeping them alive even after the go out of viewport but long term will be to optimise the ensemble widget tree as well.

fyi: both video and scroll are taken in debug mode.

lib/layout/list_view.dart
```diff
@@ -221,6 +221,7 @@ class ListViewState extends WidgetState<ListView>
           }
         }
         if (itemWidget != null) {
+          itemWidget = KeepAliveListItem(itemWidget: itemWidget);
```
kmahmood74, line 224: Flutter's documentation for `KeepAlive` states this - 

```
[ListView] and
/// [GridView], have an `addAutomaticKeepAlives` feature, which is enabled by
/// default, and which causes [AutomaticKeepAlive] widgets to be inserted around
/// each child, causing [KeepAlive] widgets to be automatically added and
/// configured in response to [KeepAliveNotification]s.
```
We use `SilverList` and that is already setting `addAutomaticKeepAlives` to true so why do we need to do this again here?

PR #1307, opened 2024-03-11
Add support handling deeplink

vinothvino42:
Added missing support to handle deeplink with passed-in URL

Another PR (ensemble_deeplink) - https://github.com/EnsembleUI/ensemble_deeplink/pull/1


lib/action/deep_link_action.dart
```diff
@@ -70,6 +70,58 @@ class DeepLinkInitAction extends EnsembleAction {
...
+      );
+    }
+    throw LanguageError('DeferredDeepLink: Missing inputs for handleDeepLink');
+  }
+
+  @override
+  Future execute(BuildContext context, ScopeManager scopeManager) async {
+    try {
+      GetIt.I<DeferredLinkManager>().handleDeferredLink(url!, (linkData) {
+        return ScreenController().executeAction(context, onLinkReceived!,
```
vusters, line 113: null check

lib/action/deep_link_action.dart
```diff
@@ -70,6 +70,58 @@ class DeepLinkInitAction extends EnsembleAction {
...
+            event: EnsembleEvent(initiator, data: {'link': linkData}));
+      });
+    } catch (e) {
+      return ScreenController().executeAction(context, onError!,
```
vusters, line 117: null check

vusters (changes requested):

lib/action/deep_link_action.dart
vinothvino42: Fixed

lib/action/deep_link_action.dart
vinothvino42: Fixed

vusters (approved):

Issue #1289, opened 2024-03-06
Notification and deep link handlers at the app level


kmahmood74:
**Problem Statement:**
Our current notification and deep link handling is not right as the handlers need to be specified on the home screen. So when the notification or deep link arrives, it first takes the user to the home screen and from there routes to the right screen. This make the user interaction suboptimal and is also not good developer experience. 

**Proposed Solution:**
App level handlers - 

`Notification Handler` 
 - specified at the app level as a js function. 
 - the js function is in one of the code libraries in the project
 - at the app level there is an optional environment variable
   - `ensemble_notification_handler` - specifies the js function to be called in this format - `<script>.<function>`. For example - `commonUtils.handleNotification`

`Deeplink handler`
 - specified at the app level as a js function. 
 - the js function is in one of the code libraries in the project
 - at the app level there is an optional environment variable
   - `ensemble_deeplink_handler` - specifies the js function to be called in this format - `<script>.<function>`. For example - `commonUtils.handleDeeplink`
   - Deeplink handler will be passed the Deeplink object as a json document
  
**Runtime:**
 Deeplink or Notification handler should be able to run any js code to do logic and return
-  `null` - to specify there is nothing to do
- `error` json object with `{status: error, code: 123, message: 'error message'}` 
- `payload` for the `navigateScreen` action. This payload must have the name of the screen and may optionally also have `inputs`
 
After calling the handler, ensemble framework should read the response and 
- if response is null, do nothing and just load the app as usual. 
- if response is error, show the errorcode and message as toast message. 
- else verify the payload and call `navigateScreen` with the payload.

kmahmood74:
@vusters @vinothvino42 there is more to notifications and deep links than what I wrote above right? how about notification badges, notifications when the app is in the background or foreground? notifications when the app is killed?

Please add comments here

snehmehta:
TODO:-
support deeplink handler as well

kmahmood74:
@sharjeelyunus please review the deeplinkhandler description in the description of this ticket and make sure it captures what you need

sharjeelyunus:
For Deep Link app-level integration make suse it is initialized at app level as well. Right now we can only initialize it at 1 screen, and that doesn't work if the user is on another screen. + If we try to initialize it on more than 1 screen it kind of works for some time but if the user closes the app and opens again it clears the storage i.e user token becomes null and we're back to square 1.

[...]

PR #1310, opened 2024-03-12
handling notification using global script

snehmehta:
closes: https://github.com/EnsembleUI/ensemble/issues/1289

Getting `navigateScreen` payload of `handleNotification` from the global script whose path is specified in environment variable.


lib/framework/notification_manager.dart
```diff
@@ -120,37 +124,66 @@ class NotificationManager {
...
     });
   }
 
-  Future<void> handleNotification(RemoteMessage message) async {
-    Map<String, dynamic> payload = {
-      'notificationPayload': {
-        'title': message.notification?.title,
-        'body': message.notification?.body,
-        'data': message.data,
+  Future<void> handleNotification(String message) async {
```
vusters, line 134: What about the existing customers with the existing format. They'll break

vusters (changes requested):

vusters (approved):
This looks find, but please test Tin's use case to make sure it still work.

snehmehta:
okay

kmahmood74:
@snehmehta please merge and close if you have tested the Tin changes, we need to close PRs quickly after they have been approved

Issue #1317, opened 2024-03-15
Share action not working on iPad


hemish11:
The share action doesn't seem to work on iPad, the app seems to be frozen.

PR #1318, opened 2024-03-15
Fixed the bug for share action for iPad

hemish11:
Closes https://github.com/EnsembleUI/ensemble/issues/1317


lib/action/misc_action.dart
```diff
@@ -72,8 +72,13 @@ class ShareAction extends EnsembleAction {
...
   @override
   Future execute(BuildContext context, ScopeManager scopeManager) {
-    Share.share(scopeManager.dataContext.eval(_text),
-        subject: Utils.optionalString(scopeManager.dataContext.eval(_title)));
+    final box = context.findRenderObject() as RenderBox?;
+
+    Share.share(
+      scopeManager.dataContext.eval(_text),
+      subject: Utils.optionalString(scopeManager.dataContext.eval(_title)),
+      sharePositionOrigin: box!.localToGlobal(Offset.zero) & box.size,
```
kmahmood74, line 80: @hemish11 I see that this is what the `share_plus` package's documentation gave as an example but what does this mean in terms of position. I see that the share box will appear right next to the widget that triggered this action? is that true, have you tested on iPad simulator to see where it will appear?

hemish11: Yes @kmahmood74, that is exactly what happened I tested it on an iPad, let me also add a screenshot for the same:

simulator_screenshot_C7C090AD-ACC4-46BD-81B2-B3AEB9A700D7

kmahmood74 (approved):

Issue #1372, opened 2024-04-19
Assert the use of stackPosition* inside a Stack


vusters:
https://discord.com/channels/1031982848485359626/1230452676509044781/1230452676509044781

Setting a stackPositionTop/down/left/right will freeze the screen if the widget which used it is not a direct child of a Stack. We need to assert that this is the case and give user a warning (see StudioDebugger's assertions)

vusters:
@hemish11 sending this to you.

PR #1378, opened 2024-04-22
Added fix for the assertion for stack widget

hemish11:
Closes https://github.com/EnsembleUI/ensemble/issues/1372

There was a bug that the assertion for the immediate parent to be stack if we have parameters like stack positions then it would freeze the entire studio as the flutter app would crash due to improper layout. So fixed those assertions

### DIFF
--- a/lib/framework/studio/studio_debugger.dart
+++ b/lib/framework/studio/studio_debugger.dart
@@ -5,7 +5,6 @@ import 'package:ensemble/layout/box/box_layout.dart';
 import 'package:ensemble/widget/helpers/controllers.dart';
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter/rendering.dart';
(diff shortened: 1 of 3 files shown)


vusters (changes requested):
THis doesn't look right. The context's widget cannot be the parent. Double check this please,

lib/framework/studio/studio_debugger.dart
```diff
@@ -133,8 +133,18 @@ class StudioDebugger {
...
+    context.visitAncestorElements((element) {
+      if (element.widget.runtimeType == Stack) {
+        isStack = true;
+      }
+
+      // This will ensure that it doesnt check for any other parent widgets
+      return false;
+    });
+
+    print(isStack);
```
vusters, line 145: always clean up these prints/logs. Customers use console log to detect errors

vusters (approved):
Make sure to remove the prints

PR #1437, opened 2024-06-02
firebase analytics

sharjeelyunus:
- restructured analytics to make it flexible to support more operations
- added setUserId support

previous EDL:

```yaml
logEvent:
  name: "Restaurant_Clicked"
  parameters:
     restaurant_name: ${name}
     restaurant_Id: ${cafeId}
     timestamp: ${ensemble.formatter.now().getDateTime()}
  ```
  
  new EDL:
  
  to log event
  ```yaml
  logEvent:
    provider: firebase # default firebase
    operation: logEvent # default logEvent
    name: "Button_Clicked" # required if operation is logEvent
    parameters:
      button_name: "Login"
      screen_name: "Login Screen"
      timestamp: "2024-06-04T12:00:00"
```

to set user id
```yaml
logEvent:
  provider: firebase
  operation: setUserId
  userId: "user123" # required if operation is setUserId
```

with this structure, analytics can be much more powerful
1. We can add more analytics libraries in `provider` 
2. currently there only 2 operations
    - logEvent
    - setUserId
  - but we can add more operations like `setUserProperty`, `logViewItem`, `logViewCart` etc in `operation` for more flexibility

### DIFF
--- a/modules/ensemble/lib/action/Log_event_action.dart
+++ b/modules/ensemble/lib/action/Log_event_action.dart
@@ -8,30 +8,43 @@ import 'package:ensemble_ts_interpreter/invokables/invokable.dart';
 import 'package:flutter/cupertino.dart';
 
 class TrackEvent extends EnsembleAction {
-  final String eventName;
+  final String? eventName;
   final Map<dynamic, dynamic>? parameters;
+  final String? provider;
+  final String? operation;
+  final String? userId;
   final String logLevel;
 
   TrackEvent({
     required Invokable? initiator,
-    required this.eventName,
+    this.eventName,
     required this.logLevel,
+    required this.provider,
+    this.operation,
+    this.userId,
     this.parameters,
   }) : super(initiator: initiator);
 
   factory TrackEvent.from({Invokable? initiator, Map? payload}) {
     payload = Utils.convertYamlToDart(payload);
     String? eventName = payload?['name'];
-    if (eventName == null) {
+    String? operation = payload?['operation'];
+
+    if (operation == 'logEvent' && eventName == null) {
       throw LanguageError(
           "${ActionType.trackEvent.name} requires the event name");
+    } else if (operation == 'setUserId' && payload?['userId'] == null) {
+      throw LanguageError("${ActionType.trackEvent.name} requires the user id");
     }
 
     return TrackEvent(
       initiator: initiator,
-      eventName: eventName,
+      eventName: eventName ?? '',
       parameters: payload?['parameters'] is Map ? payload!['parameters'] : null,
       logLevel: payload?['logLevel'] ?? LogLevel.info.name,
+      provider: payload?['provider'] ?? 'firebase',
+      operation: payload?['operation'],
+      userId: payload?['userId'],
     );
   }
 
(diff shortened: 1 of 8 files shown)


modules/ensemble/lib/action/Log_event_action.dart
```diff
@@ -7,31 +7,47 @@ import 'package:ensemble/util/utils.dart';
 import 'package:ensemble_ts_interpreter/invokables/invokable.dart';
 import 'package:flutter/cupertino.dart';
 
-class LogEvent extends EnsembleAction {
-  final String eventName;
+class TrackEvent extends EnsembleAction {
```
kmahmood74, line 10: why change the name? it is Logevent as you are logging events to the server that may or may not be used for tracking, please change it back to LogEvent.

modules/ensemble/lib/action/Log_event_action.dart
```diff
@@ -7,31 +7,47 @@ import 'package:ensemble/util/utils.dart';
...
     payload = Utils.convertYamlToDart(payload);
     String? eventName = payload?['name'];
-    if (eventName == null) {
+    String? operation = payload?['operation'];
+
+    if (operation == 'logEvent' && eventName == null) {
       throw LanguageError(
-          "${ActionType.logEvent.name} requires the event name");
+          "${ActionType.trackEvent.name} requires the event name");
+    } else if (operation == 'setUserId' && payload?['userId'] == null) {
```
kmahmood74, line 36: this is not object oriented programming. Don't overload the same logevent by adding operation and then changing all the code everywhere to handle it. Have a base class that's LogEvent and extend it with LogSetUserIdEvent and set whatever you need to set in that. 

Also have a clear logSetUserId as a method on the provider so you don't have to hide things in parameters. Fix this whole thing up. better to start fresh

kmahmood74 (changes requested):

modules/ensemble/lib/action/Log_event_action.dart
sharjeelyunus: setUserId is supposed to be called only once when user starts his session i.e after login or when user reopens the app, not with every event, that's why I added it as operation. + we might have to support more operations as well, as client was also requesting for `setUserProperty` which they realized its not needed right now, but later on they would need it.

Firebase Analytics has a lot of operations, and it the current structure is not flexible enough to add more in future. We have to add operations to make it easier for app dev side.

modules/ensemble/lib/action/Log_event_action.dart
kmahmood74: that's ok, this is generic right, other tools may call it something else. TrackEvent just doesn't make sense.

modules/ensemble/lib/framework/logging/log_provider.dart
```diff
@@ -8,8 +8,7 @@ abstract class LogProvider {
   bool shouldAwait = false;
   Map? options;
   String? ensembleAppId;
-  Future<void> log(
-      String event, Map<String, dynamic> parameters, LogLevel level);
+  Future<void> handleAnalytics(Map<String, dynamic> config);
```
kmahmood74, line 11: keep it as log please. You are just logging things, it's not handling analytics.

kmahmood74 (commented):
please fix the names. I see your point about adding 'operation', makes sense to keep it extensible as you have done. Once you change the names, it's good to go

kmahmood74 (approved):
thank you please merge

PR #1439, opened 2024-06-06
corrected the phone validator function for text input in form widget.

TheNoumanDev:
The phone validator function was coming from the following package:
image

 and was accepting strings greater then 7 and less then 15 consisting only of digit with regExp "^\d{7,15}$" which is correct but the bug was in the function.
 
image
the function is removing the non-digit number then checking the phoneExp thats why it is returning true.

so am just using simple regExp function that will do the work using the above mentioned Expression and the message doesnt mattar, we are just returning the bool.

### DIFF
--- a/modules/ensemble/lib/util/input_validator.dart
+++ b/modules/ensemble/lib/util/input_validator.dart
@@ -11,6 +11,7 @@ class InputValidator {
 
   /// check if value is a valid phone number
   static bool phone(String value) {
-    return null == ValidationBuilder().phone().test(value);
+    final RegExp PhoneRegExp = RegExp(r'^\d{7,15}$');
+    return null == ValidationBuilder().regExp(PhoneRegExp,"Please enter a valid Phone Number").test(value);
   }
 }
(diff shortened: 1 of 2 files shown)


vusters (changes requested):
Please do this:
1. Create a ValidationUtils class with validatePhone() method and put the code there.
2. Ask GPT to write a few unit tests inside ValidationUtils_test.dart.
3. inside your phone() just return validatePhone()

TheNoumanDev:
1. Above i just showed function from a package, code for `validatePhone()` was already implemented with class as `InputValidator` in Util folder (https://github.com/EnsembleUI/ensemble/blob/nouman/modules/ensemble/lib/util/input_validator.dart#L13) so ig there was not need of creating another class. i just edited the existing function, not added a new one.
2. For validation, there were already file existed named `validators_test.dart`, so i just write unit tests for phone function validation. (https://github.com/EnsembleUI/ensemble/blob/nouman/modules/ensemble/test/validators_test.dart).

(Sorry in advance if i understood wrong😅).

vusters (approved):

Issue #1342, opened 2024-03-27
Cannot set value on a dropdown with autocomplete


amin-nas:
https://studio.ensembleui.com/app/e24402cb-75e2-404c-866c-29e6c3dd7992/screen/3tBTU9CUTu0IegpytNYa

See kitchen sink example. The value set on the dropdown is ignored.

vusters:
@hemish11 most likely the same issue as https://github.com/orgs/EnsembleUI/projects/2/views/1?pane=issue&itemId=55714028.

Issue #1292, opened 2024-03-07
Dropdown Bugs


sharjeelyunus:
1. Autocomplete doesn't work with itemTemplate
2. If `autoComplete` is true, value isn't working
3. without autoComplete, the selected text overflows

![image](https://github.com/EnsembleUI/ensemble/assets/61178058/a774c492-4a9c-469b-96b2-80c40ed53ea8)

kmahmood74:
@hemish11 any update here? this is showing in progress, are you working on it?

kmahmood74:
@hemish11 ?

sharjeelyunus:
Include this as well
https://github.com/EnsembleUI/ensemble/issues/1342

CC: @hemish11

PR #1459, opened 2024-06-27
Fixed the bugs with the drop down

hemish11:
Closes https://github.com/EnsembleUI/ensemble/issues/1342 and https://github.com/EnsembleUI/ensemble/issues/1292


modules/ensemble/lib/widget/input/dropdown.dart
```diff
@@ -285,6 +285,8 @@ class SelectOneState extends FormFieldWidgetState<SelectOne>
         validatorKey.currentState!.validate();
       }
     });*/
+    textEditingController = TextEditingController(text: widget.getValue());
```
kmahmood74, line 288: also check the getValue code which in turn calls isValueInItems which has the following - 

```dart
      if (_controller.itemTemplate != null) {
        // TODO: we have no way to look into the itemTemplate to see
        // if the value matches one of them. Return true for now
        return true;
      }
```
looks like it just skips over when it is itemTemplate. Not sure how it will impact your fix

kmahmood74 (changes requested):

modules/ensemble/lib/widget/input/dropdown.dart
```diff
@@ -351,6 +353,7 @@ class SelectOneState extends FormFieldWidgetState<SelectOne>
 
     return DropdownButtonFormField2<dynamic>(
         key: validatorKey,
+        isExpanded: true,
```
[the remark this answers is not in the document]
hemish11: @kmahmood74 This one is primarily there to ensure that the child, ie the drop-down entities are properly in the size of the dropdown like it makes the items expanded and thus it resolves that issue in which the item was overflowing

modules/ensemble/lib/widget/input/dropdown.dart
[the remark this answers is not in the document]
hemish11: Yes, I'll add that to the dispose method, and also testing that value of

modules/ensemble/lib/widget/input/dropdown.dart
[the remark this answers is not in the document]
kmahmood74: I know why it's there but what is the impact of this on the actual layout? I am not sure if we can hardcode isExpanded like this?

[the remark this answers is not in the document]
hemish11: Yes, it wont affect the layout, I did tested that one

modules/ensemble/lib/widget/input/dropdown.dart
hemish11: Umm, Vu wrote this one, and yes we don't have a way by which we can get items from the item template, which is the same issue by which we cannot add autoCorrect with item template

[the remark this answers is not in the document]
kmahmood74: think about it Hemish - you are setting the value in initState which is called only once. How will it work when the developer changes value in js after the widget has already been built. Have you tested that? Can you please how it works?

modules/ensemble/lib/widget/input/dropdown.dart
```diff
@@ -246,6 +249,8 @@ class SelectOneController extends FormFieldController with HasTextPlaceholder {
   SelectOneInputFieldAction? inputFieldAction;
   List<SelectOneItem>? items;
 
+  TextEditingController textEditingController = TextEditingController();
```
kmahmood74, line 252: this should be handled internally in the widget, there is no need to expose it in the controller

modules/ensemble/lib/widget/input/dropdown.dart
```diff
@@ -73,7 +73,10 @@ abstract class SelectOne extends StatefulWidget
   Map<String, Function> setters() {
     var setters = _controller.textPlaceholderSetters;
     setters.addAll({
-      'value': (value) => _controller.maybeValue = value,
+      'value': (value) {
+        _controller.textEditingController.value = TextEditingValue(text: value);
```
kmahmood74, line 77: this should be handled internally in the widget, there is no need to expose it in the controller

kmahmood74 (changes requested):

modules/ensemble/lib/widget/input/dropdown.dart
hemish11: I needed to add it to the controller so that its available in the setters

kmahmood74: you don't need it to be available in the setters. You can update the value in the onchange event inside the state.

modules/ensemble/lib/widget/input/dropdown.dart
```diff
@@ -285,6 +289,9 @@ class SelectOneState extends FormFieldWidgetState<SelectOne>
         validatorKey.currentState!.validate();
       }
     });*/
+    widget.controller.textEditingController =
```
kmahmood74, line 292: why not just set value, why create a new textEditingController?

modules/ensemble/lib/widget/input/dropdown.dart
kmahmood74: what happens when user selects a value from the dropdown? You are setting the value when it is set in setters but what about when it is selected by user interaction? Will the textEditingController update, I don't see the code?

kmahmood74 (changes requested):

modules/ensemble/lib/widget/input/dropdown.dart
hemish11: Because to set the initial value of a text field, we must need to pass a `TextEditingController` so we cannot just update a value variable

modules/ensemble/lib/widget/input/dropdown.dart
kmahmood74: @hemish11 I am not sure why this is needed as you just override and create a new one inside initState everytime? That was my point

modules/ensemble/lib/widget/input/dropdown.dart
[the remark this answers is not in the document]
kmahmood74: @hemish11 please test the isExpanded: true with long text. Create a regular dropdown and an auto-complete dropdown each with items with very long labels that overflow multiple lines. Try it with and without isExpanded: true and report the results.

kmahmood74 (approved):

PR #1488, opened 2024-07-12
1433 networkinfo

### LINKED ISSUE #1433: Action to get network info
As an app developer, I want to get the current device connection details, including IP address

https://pub.dev/packages/network_info_plus

### Proposed EDL

```yaml
View:
  onLoad:
    getNetworkInfo:
      onError:
      onCompleted:
         # provide a map of result here
```

kmahmood74:
fixes #1433 

cc @amin-nas

### DIFF
--- a/modules/ensemble/lib/action/get_network_info_action.dart
+++ b/modules/ensemble/lib/action/get_network_info_action.dart
@@ -1,0 +1,135 @@
+import 'package:ensemble/framework/action.dart';
+import 'package:ensemble/framework/device.dart';
+import 'package:ensemble/framework/event.dart';
+import 'package:ensemble/framework/scope.dart';
+import 'package:ensemble/framework/stub/location_manager.dart';
+import 'package:ensemble/screen_controller.dart';
+import 'package:ensemble/util/utils.dart';
+import 'package:ensemble_ts_interpreter/invokables/invokable.dart';
+import 'package:flutter/cupertino.dart';
+import 'package:get_it/get_it.dart';
+
+import '../framework/stub/network_info.dart';
+
+class GetNetworkInfoAction extends EnsembleAction {
+  NetworkInfoManager networkInfo = GetIt.I<NetworkInfoManager>();
+  EnsembleAction? onSuccess;
+  EnsembleAction? onError;
+  EnsembleAction? onDenied;
+  EnsembleAction? onLocationDisabled;
+
+  GetNetworkInfoAction({super.initiator,
+    super.inputs,
+    this.onSuccess,
+    this.onError,
+    this.onDenied,
+    this.onLocationDisabled});
+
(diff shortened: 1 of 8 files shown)


modules/ensemble/lib/action/get_network_info_action.dart
```diff
@@ -0,0 +1,135 @@
...
+  EnsembleAction? onLocationDisabled;
+
+  GetNetworkInfoAction({super.initiator,
+    super.inputs,
+    this.onSuccess,
+    this.onError,
+    this.onDenied,
+    this.onLocationDisabled});
+
+  factory GetNetworkInfoAction.from({Invokable? initiator,dynamic payload}) =>
```
vusters, line 29: you don't need this. just the next one.

modules/ensemble/lib/action/get_network_info_action.dart
```diff
@@ -0,0 +1,135 @@
...
+      } else
+      if (locationStatus == LocationPermissionStatus.unableToDetermine.name
+          && onError != null) {
+        return ScreenController().executeAction(context, onError!,
+            event: EnsembleEvent(initiator,
+                data: {
+                  'status': locationStatus
+                }));
+      } else {
+        try {
```
vusters, line 70: You want to throw error if onSuccess is not specified? If you want to force it required, throw a LanguageError if not specified instead.

vusters (commented):
You also need to add this Action to action_invokable so Javascript will work. See https://github.com/EnsembleUI/ensemble/wiki/Add-New-Action

kmahmood74:
@vusters the PR is in draft mode, not ready to be reviewed

modules/ensemble/lib/action/get_network_info_action.dart
kmahmood74: will do, thanks

modules/ensemble/lib/action/get_network_info_action.dart
kmahmood74: I do need it for javascript calls, when coming from js, it's not yaml it's json

kmahmood74:
> You also need to add this Action to action_invokable so Javascript will work. See https://github.com/EnsembleUI/ensemble/wiki/Add-New-Action

@vusters this action will *not* work in js. We don't have the ability to do callbacks in js. The callbacks have to be js functions and not actions, today we assume that onSuccess etc are all actions. I'll create a separate ticket to add the ability to do callbacks as people need it to invokeAPI as well. On another thought, perhaps we shouldn't as not having that would encourage the use of yaml. Let me know what you think

kmahmood74:
@vusters ready to be reviewed. also please see my comment about js invokation

kmahmood74:
@vusters let me know when you have had a chance to review this. This is required for a customer

vusters (approved):

Issue #1493, opened 2024-07-16
Ability to dismiss toast when view changes


amin-nas:
Currently, a toast message stays on top of view when user navigates between screens. Example:
https://studio.ensembleui.com/app/63jj6KiQ1jHhXCBPvgSt/screens

I want to be able to set a toast to be specific to the current screen so that if the view changes, the toast is dismissed automatically.

## Use case:
1. User enters a password on a login screen
2. If password is inaccurate a toast message is displayed
3. When user re-enters the password and logs in, user is taken to Home screen
4. I want the toast to be automatically dismissed

## Proposed property

```yaml
showToast:
  options:
    dismissOnNavigate: true # default is false
```

Alternatively, we could also expose an action to dismissToast. Then app developer can decide when to trigger this.

hemish11:
Sample EDL:

```yaml
View:
  title: "Action: navigateScreen"
  styles:
    scrollableView: true

  body:
    Column:
      styles:
        gap: 16
        padding: 24
      children:
        - Button:
            label: Show Success
            onTap:
              showToast:
                message: Success! by default at screen's top right
                options:
                  dismissable: true
                  type: success
                  duration: 3
                styles:
                  backgroundColor: 0xFFEDEDED
                  padding: 16 24

        - Button:
            label: Show Modal
            onTap:
              navigateModalScreen:
                name: Hello Home

        - Button:
            label: Navigate and clear all screen history
            onTap:
              navigateScreen:
                name: Hello Home
                options:
                  clearAllScreens: true
                  closeToasts: true
```

PR #1499, opened 2024-07-21
Added closeToasts attribute to the nagivateModal and navigateScreen

hemish11:
Closes https://github.com/EnsembleUI/ensemble/issues/1493


hemish11:
Sample EDL:

```yaml
View:
  title: "Action: navigateScreen"
  styles:
    scrollableView: true

  body:
    Column:
      styles:
        gap: 16
        padding: 24
      children:
        - Button:
            label: Show Success
            onTap:
              showToast:
                message: Success! by default at screen's top right
                options:
                  dismissable: true
                  type: success
                  duration: 3
                styles:
                  backgroundColor: 0xFFEDEDED
                  padding: 16 24

        - Button:
            label: Show Modal
            onTap:
              navigateModalScreen:
                name: Hello Home

        - Button:
            label: Navigate and clear all screen history
            onTap:
              navigateScreen:
                name: Hello Home
                options:
                  clearAllScreens: true
                  closeToasts: true
```

modules/ensemble/lib/framework/widget/toast.dart
```diff
@@ -28,6 +28,10 @@ class ToastController {
     return _instance;
   }
 
+  void closeToast() {
+    _toast.removeQueuedCustomToasts();
```
kmahmood74, line 32: this will only get rid of the queued ones, what about the present one? should there be also be call like this - 

`fToast.removeCustomToast()`

also there is a method called `Fluttertoast.cancel()` that gets rid of all of them, did you try that?

modules/ensemble/lib/screen_controller.dart
```diff
@@ -194,6 +194,16 @@ class ScreenController {
         } else if (action.options?['replaceCurrentScreen'] == true) {
           routeOption = RouteOption.replaceCurrentScreen;
         }
+
+        if (action.options?['closeToasts'] ?? true) {
+          ToastController().closeToast();
```
kmahmood74, line 199: @hemish11 we talked about it in the standup that the default should be true for both navigateScreen and navigateModalScreen

modules/ensemble/lib/screen_controller.dart
```diff
@@ -194,6 +194,16 @@ class ScreenController {
...
+        }
+      }
+
+      // No code for NavigateModalScreenAction so need to add this
+      if (action is NavigateModalScreenAction &&
+          (action.options?['closeToasts'] ?? true)) {
```
kmahmood74, line 205: default should be true for both navigateScreen and navigateModalScreen

kmahmood74 (changes requested):

modules/ensemble/lib/framework/widget/toast.dart
hemish11: No, I thought the same at first, but then I looked at the code and we are using FToast and not Fluttertoast, so the package provides two different classes. And due to that I was using the FToast one, and the _toast is the object of the same. Also, I tested and it does remove all the toasts

modules/ensemble/lib/screen_controller.dart
hemish11: Yes, it defaults to true. So the `??` true verifies that if the `action.options?['closeToasts']` is null and when we don't pass any it null then it will default to true.

modules/ensemble/lib/screen_controller.dart
hemish11: Same as the above comment

modules/ensemble/lib/screen_controller.dart
kmahmood74: got it, thanks

modules/ensemble/lib/framework/widget/toast.dart
kmahmood74: got it, I see that the OverlayEntry.remove will get rid of the overlay. Thanks for explaining.

kmahmood74 (approved):

amin-nas:
@hemish11 we need this before end of this week. Anything left to do so you can merge?

kmahmood74:
@hemish11 please get in the habbit of merging immediately after approval, merging now

Issue #1495, opened 2024-07-17
Enable autofill on textinput


amin-nas:
I want to configure a data type for a text field so that the corresponding data is suggested to user.
https://api.flutter.dev/flutter/material/TextField/autofillHints.html

## Example
phone number suggested above keyboard here:

![image](https://github.com/user-attachments/assets/09d69865-465f-4671-ba01-3706e9fa1bd0)

PR #1503, opened 2024-07-24
Added autofillHints attribute for text fields

hemish11:
Closes https://github.com/EnsembleUI/ensemble/issues/1495

Added the autofillHints attribute for the TextInput and also added autofillEnabled for the OTP field

### DIFF
--- a/modules/ensemble/lib/util/utils.dart
+++ b/modules/ensemble/lib/util/utils.dart
@@ -287,7 +287,7 @@ class Utils {
   }
 
   static List<String>? getListOfStrings(dynamic value) {
-    if (value is YamlList) {
+    if (value is YamlList || value is List) {
       List<String> results = [];
       for (var item in value) {
         if (item is String) {
--- a/modules/ensemble/lib/widget/input/form_textfield.dart
+++ b/modules/ensemble/lib/widget/input/form_textfield.dart
@@ -156,6 +156,8 @@ abstract class BaseTextInput extends StatefulWidget
       'maxLines': (value) =>
           _controller.maxLines = Utils.optionalInt(value, min: 1),
       'textStyle': (style) => _controller.textStyle = Utils.getTextStyle(style),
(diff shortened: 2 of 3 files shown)


modules/ensemble/lib/widget/input/form_textfield.dart
```diff
@@ -156,6 +156,8 @@ abstract class BaseTextInput extends StatefulWidget
       'maxLines': (value) =>
           _controller.maxLines = Utils.optionalInt(value, min: 1),
       'textStyle': (style) => _controller.textStyle = Utils.getTextStyle(style),
+      'autofillHints': (value) =>
```
kmahmood74, line 159: @hemish11 this won't work when someone is setting `autofillHints` in js. 

```js
myField.autofillHints = ['a','b'];
```
look at the implementation of `Utils.getListOfStrings` it's terrible code that returns null if it is not yaml. 

Please ALWAYS test the js property setting too. Your code is hardcoded to work only for yaml

kmahmood74 (changes requested):

modules/ensemble/lib/util/utils.dart
```diff
@@ -287,7 +287,7 @@ class Utils {
   }
 
   static List<String>? getListOfStrings(dynamic value) {
-    if (value is YamlList) {
+    if (value is YamlList || value is List) {
```
kmahmood74, line 290: @hemish11 YamlList is List, isn't it? so the multiple check is not needed.

kmahmood74 (approved):
please look at my comment about not needing the YamlList check. Change if needed. Approved

modules/ensemble/lib/util/utils.dart
hemish11: Umm I did take a look at the implementation of YamlList and we do need this check, so ya closing it without changing it

PR #1506, opened 2024-07-25
Added support for aggregate quesries in firebase firestore.

TheNoumanDev:
Added support for aggregate quesries in firebase firestore. on get query, once PR gets approved will update documention for that as well. Tests for aggregate queries is also added.
- Right now haven't updated `subscribeToApi` function in main file as continuously updation of content is not available for aggregate function.

issue: (https://github.com/EnsembleUI/ensemble/issues/1507)

### DIFF
--- a/modules/ensemble/lib/framework/apiproviders/firestore/firestore_app.dart
+++ b/modules/ensemble/lib/framework/apiproviders/firestore/firestore_app.dart
@@ -3,7 +3,7 @@ import 'package:cloud_firestore/cloud_firestore.dart';
 class FirestoreApp {
   final FirebaseFirestore firestore;
   FirestoreApp(this.firestore);
-  Query getQuery(Map api, {bool isCollectionGroup = false}) {
+  Future<dynamic> getQuery(Map api, {bool isCollectionGroup = false}) async {
     String path = api['path'];
     Query collection;
     if (isCollectionGroup) {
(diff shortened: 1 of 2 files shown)


modules/ensemble/lib/framework/apiproviders/firestore/firestore_app.dart
```diff
@@ -116,6 +116,23 @@ class FirestoreApp {
         query = query.limitToLast(queryMap['limitToLast']);
       }
     }
+    if (api.containsKey('aggregate')) {
+        String aggregateType = api['aggregate']['type'];
```
kmahmood74, line 120: @TheNoumanDev firestore supports aggregates over multiple fields. Your schema and PR code is only allowing one field, why is that?

modules/ensemble/lib/framework/apiproviders/firestore/firestore_app.dart
```diff
@@ -3,7 +3,7 @@ import 'package:cloud_firestore/cloud_firestore.dart';
 class FirestoreApp {
   final FirebaseFirestore firestore;
   FirestoreApp(this.firestore);
-  Query getQuery(Map api, {bool isCollectionGroup = false}) {
+  Future<dynamic> getQuery(Map api, {bool isCollectionGroup = false}) async {
```
kmahmood74, line 6: @TheNoumanDev getQuery is also called inside subscribeToApi function and that expects a Query object back. Does your code even compile anymore? did you even test the firestore api provider?

kmahmood74 (changes requested):
this is just not right. Don't work on it anymore. Cancel this PR. I'll take care of it.

PR #1519, opened 2024-08-02
fontFamily now properly propagates to all widgets. Refactored Buttons…

vusters:
… code


kmahmood74:
@vusters is this still pending?

Issue #1521, opened 2024-08-02
ConfirmationInput should have obscureText property like we have for TextInput


mehsaandev:
ConfirmationInput should have obscureText property like we have for TextInput

PR #1525, opened 2024-08-04
ConfirmationInput pinType support

TheNoumanDev:
related issue: https://github.com/EnsembleUI/ensemble/issues/1521

properties:
- `pinType`: to have the input as password, possible values are (password/custom/none(default))
- `pinTypeCustom`: to deal if user select `custom` in `pinType` to have there own choice of pin text to displayed.

Example EDL:
```yaml
ConfirmationInput:
  styles:
    length: 6
    fieldHeight: 50
    fieldWidth: 50
    gap: 10
    borderRadius: 16
    borderWidth: 2
    pinType: custom # can be password/custom/none
    pinTypeCustom: / # by default if is *
  onChange: |
    console.log(this.text);
```

Demo Video: 

https://github.com/user-attachments/assets/578abdac-6214-4eae-8655-fa02630af89a

before merging the PR will update the studio schema and kitchen sink example once changes get approved.

### DIFF
--- a/modules/otp_pin_field/lib/src/otp_pin_field_state.dart
+++ b/modules/otp_pin_field/lib/src/otp_pin_field_state.dart
@@ -371,20 +371,16 @@ class OtpPinFieldState extends State<OtpPinField>
   String _getPinDisplay(int position) {
     var display = '';
     var value = pinsInputed[position];
-    if( widget.obscureText == true ) {
-      display = '*'; // Use asterisks to obscure text
-    } else {
-        switch (widget.otpPinFieldInputType) {
-          case OtpPinFieldInputType.password:
-            display = '*';
-            break;
-          case OtpPinFieldInputType.custom:
-            display = widget.otpPinInputCustom;
-            break;
-          default:
-            display = value;
-            break;
-        }
+    switch (widget.otpPinFieldInputType) {
+      case OtpPinFieldInputType.password:
+        display = '*';
+        break;
+      case OtpPinFieldInputType.custom:
+        display = widget.otpPinInputCustom;
+        break;
+      default:
+        display = value;
+        break;
     }
     return value.isNotEmpty ? display : value;
   }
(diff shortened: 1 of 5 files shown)


kmahmood74:
what's the difference between obscureText and OtpPinFieldInputType.password?

modules/otp_pin_field/lib/src/otp_pin_field_state.dart
```diff
@@ -371,16 +371,20 @@ class OtpPinFieldState extends State<OtpPinField>
...
+          case OtpPinFieldInputType.custom:
+            display = widget.otpPinInputCustom;
+            break;
+          default:
+            display = value;
+            break;
+        }
     }
     return value.isNotEmpty ? display : value;
   }
```
sharjeelyunus, line 390: @TheNoumanDev this doesn't seems right, looks like this was already implemented in `otp_pin_field` module with `OtpPinFieldInputType.password` but wasn't exposed to the widget. I think the whole `OtpPinFieldInputType` enum is not exposed. and instead of duplicating the functionality, please just expose the `pinFieldType` with options `none`, `password` and `custom` and by default it should be none

TheNoumanDev: @sharjeelyunus @kmahmood74  updated PR and comment with new fields explanation and example EDL.

sharjeelyunus (approved):
LGTM

kmahmood74:
@vusters can you please take a look as ConfirmationInput is your implementation. Customer requirement is that the OTP should be obscured and not be visible as text.

vusters:
Why not just used almost what you had originally?
- obscureText: true | false
- obscureSymbol: string (single character)

vusters (approved):

kmahmood74:
@TheNoumanDev please always put the final EDL in the ticket as well. Also add documentation, it's not there

TheNoumanDev:
@kmahmood74 i updated the final EDL in the description and also updated the docs and kitchen sink example.

PR #1536, opened 2024-08-06
fix: syntax typo

snehmehta:


snehmehta:
Merging it as it's just syntax fix

Issue #1527, opened 2024-08-05
TextInput's cursor is acting weirdly on real device


ridsashabbir:
i am trying to edit text in TextInput, so there should be some text in the TextInput beforehand, and when i click on the edit button from the bottom modal, then it is supposed to show the cursor after the text, but it is showing it weirdly, and it shows the cursor first, then it shows the text in the TextInput, what i want is, text should be there first, and cursor should appear after the text, so that user could edit the text easily... it is working fine for studio, but it is not working on ensemble preview.

here is the studio example for that:
https://studio.ensembleui.com/app/fvWvR0DqRko5dpkfI7v4/screen/Sslp51siQdNx0nbz9lqh

recording of this bug: 
https://drive.google.com/file/d/1oRpKJVA7Aj5BbYClVfx-gmgaNkkXhZok/view?usp=drive_link

sharjeelyunus:
This issue is still there.
Steps : 
1. open dialog with TextInput (cursor will be at the end of the field) 
2. Move the cursor to the starting of field and type something => the cursor will move to last position.

CC: @snehmehta

kmahmood74:
this was due to a bad app design where the the storage value that the field was bound to was being modified onKeyPress of the field. That's not right. I already proposed a design that works fine. Why do we need this?

kmahmood74:
you need to set the value when closing the dialog. The app logic is missing that step. 

```yaml
        - Button:
            label: Save
            styles:
              margin: 16 0 0 0
            onTap:
              executeActionGroup:
                actions:
                  - executeCode:
                      body: |
                        ensemble.storage.announcementText = ensemble.storage.announcementText_live;
                  - dismissBottomModal:
         ```

PR #1539, opened 2024-08-06
fix: move cursor to end while rebuild

snehmehta:
ticket: https://github.com/EnsembleUI/ensemble/issues/1527


modules/ensemble/lib/widget/input/form_textfield.dart
```diff
@@ -346,11 +346,22 @@ class TextInputState extends FormFieldWidgetState<BaseTextInput>
...
     super.didUpdateWidget(oldWidget);
     widget.controller.inputFieldAction = this;
+
+    // Making sure to move cursor to end when widget rebuild
+    // issue: https://github.com/EnsembleUI/ensemble/issues/1527
+    int cursorPosition = widget.textController.selection.baseOffset;
+    if (widget.textController.text.length != cursorPosition &&
+        focusNode.hasFocus) {
+      widget.textController.selection = TextSelection.fromPosition(
+        TextPosition(offset: widget.textController.text.length),
```
kmahmood74, line 356: I don't understand - wouldn't this move the cursor to the very end even when the user has it placed in the middle of the text while editing it? Isn't that the actual bug?

snehmehta: No, the bug is when textfield rebuild which happens when textfield in bottomModal, the cursor starts from the end. 
The reason behind it, essentially flutter correctly puts at the end but bottomModal append bottomInsets when keyboard is displayed which makes our ensemble widget rebuild and put cursor again in starting position. 

The solution just make sure to put it at the end if required.

vusters: A better behavior would be to get the previous cursor position and go to that exact position. If the old text is longer somehow (which it shouldn't) the go to the end. You already have access to the old controller

vusters (approved):

PR #1547, opened 2024-08-07
added support for icon and string as obscure in confirmationWidget

TheNoumanDev:
Added support for icon and string as obscure in confirmationWidget.

Example EDl:
```yaml
ConfirmationInput:
  styles:
    length: 6
    fieldHeight: 50
    fieldWidth: 50
    gap: 10
    borderRadius: 16
    borderWidth: 2
    obscureText: custom # can be password/custom/none
    #obscureSymbol: / # by default if is *
    obscureSymbol:
      icon:
         name: code
  onChange: |
    console.log(this.text);
```

Tested with following inputs:
```yaml
obscureSymbol:                     # (output * )
obscureSymbol: /                  # (output / )
obscureSymbol: //                 # (output * )
obscureSymbol:                     # (output * )
  icon:
obscureSymbol:                     # (output * )
  icon:
    name:
obscureSymbol:                     # (output icon )
  icon:
    name: code
obscureSymbol:                      # (output icon )
  icon:
    name: code 
    library: fontAwesome
obscureSymbol:                      # (output icon )
  icon: 'code fontAwesome'
obscureSymbol:                      # (output icon )
  icon:
    name: code 
    library: fontAwesome
    size: 20
obscureSymbol:                      # (output icon )
    icon: {name: code}
```
using JS:
```js
confirmationInputId.obscureText = true;
confirmationInputId.obscureSymbol = {icon: {name: 'code'}} ;
```


modules/ensemble/lib/widget/confirmation_input.dart
```diff
@@ -35,7 +37,7 @@ class ConfirmationInput extends StatefulWidget
           _controller.fieldType = Utils.optionalString(input),
       'inputType': (type) => _controller.inputType = Utils.optionalString(type),
       'obscureText': (type) => _controller.obscureText = Utils.optionalString(type),
```
kmahmood74, line 39: @TheNoumanDev `obscureText` should be a boolean. We need to be consistent with TextInput. See Vu's message on discord as well. We need two things - 

obscureText: boolean 
obscureSymbol: string | icon

if obscureText  == true, only then you process the obscureSymbol. And the default obscureSymbol is * Please fix yoru code and repost

kmahmood74 (changes requested):

modules/ensemble/lib/widget/confirmation_input.dart
TheNoumanDev: Hey @Khurram!
Below is just description of wot happening bcz first i kept it boolean and sharjeel bhai told me to not change types that are already defined😅 , then bcz of Icon requirement i changed neccesary property and code but kept rest same as @vusters approved last changes but i understand it should be boolean to keep it consistent, should i change it or keep it same? 🙂 

It actually has 3 states defined in enum that's why it is in string:
`none`: to display plain text (default)
`password`: to have obscure text just as `*`.
`custom`: to have obscure text but with their own choice of character or Icon and `obscureSymbol` will only be processed if `custom` is selected.

kmahmood74: we can make it backward compatible as follows - 

1. if you get `null` returned by `Utils.optionalBool(obscure)`, parse it as a string
2. update the schema so that for all future apps, `obscureText` is a bool

I don't know what `password` is needed as it seems to be the exact same as `obscureText` at least in this case.

TheNoumanDev:
@kmahmood74 updated code to have:
- `obscureText`: boolean
- `obscureSymbol`: string(length == 1) | icon

if `obscureText == true`, only then `obscureSymbol` is processed. And the default obscureSymbol is `*` if wrong input or not given.

Example EDL:
```yaml
ConfirmationInput:
  styles:
    length: 6
    fieldHeight: 50
    fieldWidth: 50
    gap: 10
    borderRadius: 16
    borderWidth: 2
    obscureText: true # can be true/false
    #obscureSymbol: / # by default if is *
    obscureSymbol:
      icon:
         name: code
```
- test inputs remains the same!!!

modules/ensemble/lib/widget/confirmation_input.dart
```diff
@@ -34,8 +36,8 @@ class ConfirmationInput extends StatefulWidget
       'fieldType': (input) =>
           _controller.fieldType = Utils.optionalString(input),
       'inputType': (type) => _controller.inputType = Utils.optionalString(type),
-      'obscureText': (type) => _controller.obscureText = Utils.optionalString(type),
-      'obscureSymbol': (typeCustom) => _controller.obscureSymbol = Utils.optionalString(typeCustom),
+      'obscureText': (type) => _controller.obscureText = Utils.optionalBool(type),
```
kmahmood74, line 39: @TheNoumanDev looks like you just ignored the backward compatability comment that I had posted. How is this backward compatible!

kmahmood74:
1. if obscureText is a bool -> do what you are doing
2. if obscureText is a string -> set _controller.obscureText to true and set _controller.obscureSymbol = obscureText as string
rest of your code will not change right. This will ensure that existing uses of obscureText won't break

modules/ensemble/lib/widget/confirmation_input.dart
```diff
@@ -201,7 +203,7 @@ class ConfirmationInputState extends framework.WidgetState<ConfirmationInput>
       keyboardType: widget.keyboardType,
       otpPinFieldDecoration: controller.fieldType?.otpPinField ??
           OtpPinFieldDecoration.defaultPinBoxDecoration,
-      otpPinFieldInputType: controller.obscureText?.otpPinType ?? OtpPinFieldInputType.none,
+      otpPinFieldInputType: controller.obscureText == true ? OtpPinFieldInputType.password : OtpPinFieldInputType.none,
```
kmahmood74, line 206: why is this needed?

modules/ensemble/lib/widget/confirmation_input.dart
```diff
@@ -244,24 +246,21 @@ class ConfirmationInputState extends framework.WidgetState<ConfirmationInput>
...
-String _validatePinTypeCustom(String? value) {
-    if (value == null || value.length != 1) {
-      return "*";
+dynamic _validatePinTypeCustom(dynamic value) {
+  if ( value is String ) {
+    if (  value.length != 1) {
+      return "*"; // Default symbol if string length is not 1
     }
     return value;
+  } else if (value is Map && value['icon'] != null) {
```
kmahmood74, line 255: what if it's YamlMap? I believe YamlMap is Map but please test. also test when the value is being set in js

modules/ensemble/lib/widget/confirmation_input.dart
TheNoumanDev: we are using `otp_pin_field` module to create `confirmationInput` and obsure property is defined here https://github.com/EnsembleUI/ensemble/blob/bf9a9948913d4c7c610399c72238a6076b007406/modules/otp_pin_field/lib/src/otp_pin_field_widget.dart#L17 with dataType being enum defined at https://github.com/EnsembleUI/ensemble/blob/bf9a9948913d4c7c610399c72238a6076b007406/modules/otp_pin_field/lib/src/otp_pin_field_input_type.dart#L1. and its manipulation that how it will be displayed is done in `otp_pin_field_state.dart` that's why adjusted the `confirmationInput` code accordingly.

modules/ensemble/lib/widget/confirmation_input.dart
TheNoumanDev: Yup testes with multiple inputs, added js inputs in description as well.

modules/ensemble/lib/widget/confirmation_input.dart
kmahmood74: sorry no need for backward compatability and obscureText is a new property anyway that just went in recently

kmahmood74:
> 1. if obscureText is a bool -> do what you are doing
> 2. if obscureText is a string -> set _controller.obscureText to true and set _controller.obscureSymbol = obscureText as string
>    rest of your code will not change right. This will ensure that existing uses of obscureText won't break

please ignore my above comment

modules/ensemble/lib/widget/confirmation_input.dart
```diff
@@ -36,7 +36,10 @@ class ConfirmationInput extends StatefulWidget
       'fieldType': (input) =>
           _controller.fieldType = Utils.optionalString(input),
       'inputType': (type) => _controller.inputType = Utils.optionalString(type),
-      'obscureText': (type) => _controller.obscureText = Utils.optionalBool(type),
+      'obscureText': (type) {
```
kmahmood74, line 39: this is not needed, sorry for the backward compatibility comment. 

you can just revert to 

'obscureText': (type) => _controller.obscureText = Utils.optionalBool(type),

thanks and sorry again, approving with this change request

kmahmood74 (approved):
please see my comment and make that small change before merging.

modules/ensemble/lib/widget/confirmation_input.dart
TheNoumanDev: Thanks khurram for making me think in that prespective as well, will surely keep that in mind if somewhere it is needed.

Issue #1548, opened 2024-08-07
provide `maxLength` on all TextInputs and PasswordInput


kmahmood74:
we need two properties to be exposed - 

```
maxLength
maxLengthEnforcement - none, enforced, truncateAfterCompositionEnds
```
https://api.flutter.dev/flutter/material/TextField-class.html

PR #1550, opened 2024-08-08
TextInput/PasswordInput: maxLength and maxLengthEnforcement 

sharjeelyunus:
Closes: https://github.com/EnsembleUI/ensemble/issues/1548

EDL:

```yaml
 - TextInput:
       label: Hello world
       maxLength: 10
       maxLengthEnforcement: none # none, enforced, truncateAfterCompositionEnds
- PasswordInput:
       maxLength: 8
       maxLengthEnforcement: enforced
```

image

### DIFF
--- a/modules/ensemble/lib/widget/input/form_textfield.dart
+++ b/modules/ensemble/lib/widget/input/form_textfield.dart
@@ -158,6 +158,9 @@ abstract class BaseTextInput extends StatefulWidget
       'textStyle': (style) => _controller.textStyle = Utils.getTextStyle(style),
       'autofillHints': (value) =>
           _controller.autofillHints = Utils.getListOfStrings(value),
+      'maxLength': (value) => _controller.maxLength = Utils.optionalInt(value),
+      'maxLengthEnforcement': (value) =>
+          _controller.maxLengthEnforcement = _getMaxLengthEnforcement(value),
     });
     return setters;
   }
@@ -234,6 +237,8 @@ class TextInputController extends FormFieldController with HasTextPlaceholder {
   bool? multiline;
   int? minLines;
   int? maxLines;
+  int? maxLength;
+  MaxLengthEnforcement? maxLengthEnforcement;
 
   List<String>? autofillHints;
 }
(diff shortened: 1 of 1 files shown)


amin-nas:
@sharjeelyunus  Is there another state beside `none` and `enforced`?
Boolean seems more inline with how we do these properties:
```yaml
maxLengthEnforced: true # default to false
```

Please link the relevant ticket. That's where we should discuss the EDL ahead of implementation.

kmahmood74:
> @sharjeelyunus Is there another state beside `none` and `enforced`? Boolean seems more inline with how we do these properties:
> 
> ```yaml
> maxLengthEnforced: true # default to false
> ```
> 
> Please link the relevant ticket. That's where we should discuss the EDL ahead of implementation.

there is - truncateAfterCompositionEnds it's there in the code as well. 

  /// Users can still input text if the current value is composing even after
  /// reaching the max length limit. After composing ends, the value will be
  /// truncated.

kmahmood74:
every PR must reference its ticket and must put the yaml in the ticket as well. @sharjeelyunus please provide that

kmahmood74 (approved):
I am assuming maxLengthEnforcement is optional and the default behavior is 'enforced' when maxLength is specified and maxLengthEnforcement is not.

sharjeelyunus:
> I am assuming maxLengthEnforcement is optional and the default behavior is 'enforced' when maxLength is specified and maxLengthEnforcement is not.

hey, the default is none, let me know if you want me to make it enforced by default

kmahmood74:
> > I am assuming maxLengthEnforcement is optional and the default behavior is 'enforced' when maxLength is specified and maxLengthEnforcement is not.
> 
> hey, the default is none, let me know if you want me to make it enforced by default

default is not none in flutter, it is enforced. We should keep the same default. 

from flutter - 

  /// Determines how the [maxLength] limit should be enforced.
  ///
  /// Defaults to [MaxLengthEnforcement.enforced].
  ///
  /// {@macro flutter.services.textFormatter.maxLengthEnforcement}
  final MaxLengthEnforcement? maxLengthEnforcement;

PR #1579, opened 2024-08-18
[Button] loading state

sharjeelyunus:
- loading
- loadingIndicator

EDL:
```yaml
Button:
  label: Continue
  loadingIndicator:
     Progress:
        #any configuration for Progress widget
  loading: ${getPeople.isLoading} #whether to display or not, could be bound to anything
```

https://github.com/user-attachments/assets/d15a7eea-3f57-45cf-b2ea-8119bd7dc516

Some example loadings achieved with this change 

https://github.com/user-attachments/assets/3845e995-f178-4b2d-a021-954946ba6d1a


kmahmood74 (changes requested):
as described here, we need to make it a generic loadingWidget that could be any widget. Your code is hardcoded for the loading indicator, it won't work for many customers.

modules/ensemble/lib/widget/button.dart
```diff
@@ -148,6 +154,39 @@ class ButtonState extends EWidgetState<Button> {
   }
 
   Widget _buildButtonChild() {
+    if (widget._controller.loading == true) {
+      // Measure text size to keep the button consistent
+      final textSpan = TextSpan(
```
kmahmood74, line 159: this could be the default loading container. But we still need to provide the ability for devs to put their own custom widget as a loading container. Please see https://github.com/EnsembleUI/ensemble/issues/1584

kmahmood74 (changes requested):

kmahmood74:
@sharjeelyunus any update here?

kmahmood74:
@sharjeelyunus we don't want this to just be ignored. Let me know if you don't have time. We need to implement it as a generic loading widget

modules/ensemble/lib/widget/button.dart
```diff
@@ -149,6 +155,24 @@ class ButtonState extends EWidgetState<Button> {
   }
 
   Widget _buildButtonChild() {
+    if (widget._controller.loading == true) {
+      return widget._controller.loadingWidget != null && scopeManager != null
+          ? scopeManager!
```
kmahmood74, line 160: @vusters does this look ok to you, I know you changed how widgets are created recently?

modules/ensemble/lib/widget/button.dart
```diff
@@ -115,6 +118,9 @@ class ButtonController extends BoxController {
 
   IconModel? startingIcon;
   IconModel? endingIcon;
+
+  bool? loading;
```
vusters, line 122: use `isLoading` to be consistent with other widgets

modules/ensemble/lib/widget/button.dart
```diff
@@ -115,6 +118,9 @@ class ButtonController extends BoxController {
...
+  YamlMap? loadingWidget;
```
vusters, line 123: use `dynamic` type. This may throw an exception if we put anything but Map into the definition.

modules/ensemble/lib/widget/button.dart
```diff
@@ -149,6 +155,24 @@ class ButtonState extends EWidgetState<Button> {
   }
 
   Widget _buildButtonChild() {
+    if (widget._controller.loading == true) {
+      return widget._controller.loadingWidget != null && scopeManager != null
+          ? scopeManager!
+              .buildWidgetFromDefinition(widget._controller.loadingWidget)
+          : // default loading widget
+          SizedBox(
```
vusters, line 163: more of a design question for @amin-nas. Should the default loading state simply be the same button with reduced opacity? Probably the most common case that won't scream out "I need to update this right away".

vusters (approved):
Make sure you change what was asked and you should be good.

modules/ensemble/lib/widget/button.dart
sharjeelyunus: actually, the most common loading type is Circular loading, examples from our platform: MyPlace, UniConnect, and Chowmill all use circular loadings, and currently doing it with conditionals

image

kmahmood74 (approved):

Issue #1153, opened 2024-01-17
Request: Dropdown Widget Styling


omermahmood:
1. Ability to style the font and background of the Dropdown Widget Values
2. Dropdown Widget: Ability to style the border to match the Widget (round edges)

<img width="184" alt="image" src="https://github.com/EnsembleUI/ensemble/assets/18559315/c01f5de0-5ea9-4873-a153-6ab00246006c">

amin-nas:
The requested styling items are added except for for border color.

TheNoumanDev:
# Problem
For dropdown before if we were using the below EDL:
```yaml
- Dropdown:
    styles:              
      borderColor: red
      borderWidth: 4
    items:
      - Item 1
      - Item 2
```
It gives the below result, as `borderColor` was applied on the `border` of the dropdown.
![image](https://github.com/user-attachments/assets/5b97e732-04c2-4711-bc7c-9e6b32198471)

# Fixed Solution
so updated that bug, now styling attributes that start with `dropdown` is styling the list of options that open up, e.g. `dropdownBackgroundColor` and `dropdownBorderColor` and separately, we have `borderColor` and `borderWidth` that style the input box only.

Exampl EDL:
```yaml
Dropdown:
   label: Colors
   floatLabel: true
   items:
       - Red
       - Blue
       - Green
    styles:
       # variant: underline
       # autoComplete: true
       focusedBorderColor: black
       borderRadius: 50
       borderColor: red
       dropdownBorderWidth: 5
       dropdownBorderColor: green
       borderWidth: 5
```
Issue: https://github.com/EnsembleUI/ensemble/issues/1153

PR #1588, opened 2024-08-21
Dropdown border property

TheNoumanDev:
# Problem
For dropdown before if we were using the below EDL:
```yaml
- Dropdown:
    styles:              
      borderColor: red
      borderWidth: 4
    items:
      - Item 1
      - Item 2
```
It gives the below result, as `borderColor` was applied on the `border` of the dropdown.
image

# Fixed Solution
so updated that bug, now styling attributes that start with `dropdown` is styling the list of options that open up, e.g. `dropdownBackgroundColor` and `dropdownBorderColor` and separately, we have `borderColor` and `borderWidth` that style the input box only.

Exampl EDL:
```yaml
Dropdown:
   label: Colors
   floatLabel: true
   items:
       - Red
       - Blue
       - Green
    styles:
       # variant: underline
       # autoComplete: true
       focusedBorderColor: black
       borderRadius: 50
       borderColor: red
       dropdownBorderWidth: 5
       dropdownBorderColor: green
       borderWidth: 5
```
Issue: https://github.com/EnsembleUI/ensemble/issues/1153


kmahmood74:
@TheNoumanDev please put the detailed description in the #1153 no one reads the pull requests, ticket should be our source of truth

modules/ensemble/lib/widget/input/dropdown.dart
```diff
@@ -439,6 +498,61 @@ class SelectOneState extends FormFieldWidgetState<SelectOne>
                         : null,
                     fillColor: widget._controller
                         .fillColor, // Background color for the field
+                    enabledBorder: widget._controller.dropdownBorderColor != null ||
```
kmahmood74, line 501: the autocomplete enabledBorder code seems like a copy paste of the one before, please re-use by creating a function instead of copy/pasting

kmahmood74 (changes requested):

kmahmood74:
@TheNoumanDev is there no way to define the borders for the actual dropdown field (not the dropdown panel but the field itself) or is that handled by the box properties?

kmahmood74 (approved):

modules/ensemble/lib/widget/input/dropdown.dart
TheNoumanDev: Created.

PR #1597, opened 2024-08-27
Fixed the shimmer effect

hemish11:
Fixes the shimmer effect.


modules/ensemble/lib/widget/loading_container.dart
```diff
@@ -53,7 +56,7 @@ class LoadingContainer extends StatefulWidget
       'defaultShimmerPadding': (value) => _controller.padding =
           Utils.getInsets(value), // Backward compatibility
       'shimmerOptions': (options) {
-        if (options is Map<String, dynamic>) {
+        if (options is YamlMap) {
```
kmahmood74, line 59: how will this work when options are being set in javascript? this is something I keep bringing up in every meeting

modules/ensemble/lib/widget/loading_container.dart
```diff
@@ -148,42 +142,45 @@ class LoadingContainerState extends EWidgetState<LoadingContainer> {
...
+              const Color(0xFFEBEBF4).withOpacity(0.3),
+          widget._controller.baseColor ?? const Color(0xFFEBEBF4),
+        ];
+
+    return LinearGradient(
+      begin: _getGradientBegin(),
+      end: _getGradientEnd(),
+      colors: colors,
+      stops: widget._controller.gradientStops ?? const [0.1, 0.3, 0.4],
+      tileMode: TileMode.clamp,
```
kmahmood74, line 158: this should not be hardcoded. Default could be clamp but give the user options to pick something else

modules/ensemble/lib/widget/loading_container.dart
kmahmood74: did you test it with js?

kmahmood74 (approved):
approving, I'll fix the js part. please merge

PR #1602, opened 2024-08-28
Git based version with melos

snehmehta:


modules/ensemble/lib/framework/data_context.dart
```diff
@@ -464,9 +465,24 @@ class NativeInvokable extends ActionInvokable {
       'formatter': () => Formatter(),
       'utils': () => _EnsembleUtils(),
       'device': () => Device(),
+      'version': () => getEnsembleVersion(),
```
vusters, line 468: cache and mutex this. We'll be doomed if user put this in a loop. Less desirable but can also do this once at start up.

modules/ensemble/pubspec.yaml
```diff
@@ -15,7 +15,7 @@ description: Ensemble Runtime
 # This version is used _only_ for the Runner app, which is used if you just do
 # a `flutter run` or a `flutter make-host-app-editable`. It has no impact
 # on any other native host app that you embed your Flutter project into.
-version: 1.0.0+5
+version: 1.0.1
```
vusters: Today this auto-incment (in iOS). Do we have to reember to change this manually every time we push to Testflight?

modules/ensemble/lib/framework/data_context.dart
snehmehta: I have added cache. It will cache on first run not at startup at the moment

modules/ensemble/pubspec.yaml
snehmehta: No will be auto-increment only.

Issue #1598, opened 2024-08-27
BUG: `shimmerEffect: vertical` does not work on the native devide


kmahmood74:
https://studio.ensembleui.com/app/e24402cb-75e2-404c-866c-29e6c3dd7992/screen/3wNZSfVkdi56zmTtFeBT

change shimmerEffect to vertical. works fine on the web but not on native device. Check on iOS simulator

PR #1604, opened 2024-08-28
Fixed the Diagonal shimmer effect and also added tileMode

hemish11:
Closes https://github.com/EnsembleUI/ensemble/issues/1598


kmahmood74 (approved):
thanks. Please update the schema, documentation and the kitchen sink example as well as you just added the tileMode property

Issue #1549, opened 2024-08-07
Custom Inline time picker


ridsashabbir:
we need a time picker for customer app, where first we have a option to select 'am' or 'pm' then there is time, user can scroll it up or down to select specific time, and the time that is being selected will be displayed on screen with formatting like '4:45 pm', like the one we use to set alarm in real device

here is the screenshot of what exactly we need to built:
https://imgur.com/sqwB6cL

kmahmood74:
@hemish11 any update on this one, it shows in progress?

hemish11:
Sample EDL:
```yaml
View:
  header:
    title: Inline TImePicker
  styles:
    scrollableView: true
  body:
    Column:
      children:
        - InlineTimePicker:
            id: timePicker
            initialTime: "11:44"
            mode: hms
            showHourLabel: false

        - Button:
            label: Get Time
            onTap: |
              //@code
              console.log(timePicker.selectedTime);
```

hemish11:
@kmahmood74 here's the video of it's working:

https://github.com/user-attachments/assets/49af2ca4-0dac-4f48-90bd-ba5def68e560

kmahmood74:
@hemish11 please implement the time picker as clearly described in the description of the ticket. You have implemented a duration widget and not a time widget.

kmahmood74:
assigning to @TheNoumanDev Nouman, the PR that @hemish11 created does not provide the inline time picker, instead it provides a time range widget. Please ignore that PR and implement this ticket from scratch. 

We already have a `Time` widget. 

1. There should be a property on the time widget called `userIOSStyleTimePicker` as a boolean. This should be true by default on iOS and false on all other platforms. User can set it to false to override the default behavior on iOS as well. 
2. When set to true, the `CupertinoDatePicker` with the time mode should appear. 
3. Expose properties such as `use24hFormat` on the actual Time widget and make sure the existing time picker and the cupertino time picker both process these properties

Let me know if you have any questions. Also once you have it ready for testing, check with Ridsa to make sure her use case is covered, use her app to configure it under the calendar as she needs it and then test

@ridsashabbir the time picker will appear like this - 
<img width="332" alt="image" src="https://github.com/user-attachments/assets/a56a941e-b2f6-4d43-8a3f-0ff22d626f8d">

TheNoumanDev:
[...]

PR #1626, opened 2024-09-06
updated time widget to support iOS style

TheNoumanDev:
Related #1549 

Example Video & EDL is added in ticket.

once PR gets approved, will create schema PR and update docs.


modules/ensemble/lib/widget/input/form_time.dart
```diff
@@ -101,30 +125,68 @@ class TimeState extends FormFieldWidgetState<Time> {
...
+      _updateTime(picked);
+    }
+  }
+
+  void _showCupertinoTimePicker(BuildContext context) {
+    showCupertinoModalPopup<void>(
+      context: context,
+      builder: (BuildContext context) => Container(
+        height: 216,
+        padding: const EdgeInsets.only(top: 6.0),
```
kmahmood74, line 156: this should come from syles

modules/ensemble/lib/widget/input/form_time.dart
```diff
@@ -101,30 +125,68 @@ class TimeState extends FormFieldWidgetState<Time> {
...
+        margin: EdgeInsets.only(
+          bottom: MediaQuery.of(context).viewInsets.bottom,
+        ),
+        color: CupertinoColors.systemBackground.resolveFrom(context),
```
kmahmood74, line 160: see the original time picker and use theme

modules/ensemble/lib/widget/input/form_time.dart
```diff
@@ -42,33 +46,53 @@ class Time extends StatefulWidget
...
     }
-  }
 
-  TimeOfDay? initialValue;
+    String formattedTime;
+    if (use24hFormat) {
+      formattedTime =
+          '${value!.hour.toString().padLeft(2, '0')}:${value!.minute.toString().padLeft(2, '0')}';
+    } else {
+      formattedTime = value!.format(context);
```
kmahmood74, line 84: will this format based on the locale?

modules/ensemble/lib/widget/input/form_time.dart
```diff
@@ -42,33 +46,53 @@ class Time extends StatefulWidget
...
+  
   Text prettyValue(BuildContext context) {
-    if (value != null) {
-      return Text(value!.format(context),
-          style: TextStyle(fontSize: fontSize?.toDouble()));
-    } else {
-      return Text(placeholder ?? 'Select a time', style: placeholderStyle);
+    if (value == null) {
+      return Text(
+        placeholder ?? 'Select a time',
```
kmahmood74, line 74: is there a localized default placeholder that Flutter has?

kmahmood74 (changes requested):

modules/ensemble/lib/widget/input/form_time.dart
TheNoumanDev: added style properties to adjust these for now like `pickerHeight`, `pickerPadding`, and `pickerBackgroundColor` applied to iOS picker.
Did try to update to `showTimePicker` as well but it has its own multiple properties to adjust the style, to customize the material widget style, will have to add 6,7 new properties that would only affect `showTimePicker`, if you say so, then will expose them other it's having the default style and `CupertinoDatePicker` is being updated correctly.

modules/ensemble/lib/widget/input/form_time.dart
TheNoumanDev: before it was using `format` which was not formating for 24 hours format, that's why I used it but I say `MaterialLocalizations` have a function that format the date to local based on 24h flag as well so using it now also it has a default hint text so using it as well.

modules/ensemble/lib/widget/input/form_time.dart
TheNoumanDev: yup, using it from `MaterialLocalizations `.

modules/ensemble/lib/widget/input/form_time.dart
```diff
@@ -112,7 +113,7 @@ class TimeController extends FormFieldController with HasTextPlaceholder {
   TimeOfDay? value;
   TimeOfDay? initialValue;
   EnsembleAction? onChange;
-  bool useIOSStyleTimePicker = Platform.isIOS;
+  bool useIOSStyleTimePicker = kIsWeb ? false : Platform.isIOS;
```
kmahmood74, line 116: if someone is running the webapp on safari on iOS, it seems ok to show them the iOSStyleTimePicker. Why make it false?

TheNoumanDev: As i showed yesterday, web build was failing, it was due to that `Platform` library was not available on Web, so I put a check on if its not web then check.
But yeah I can do the following scenario:

- if it's web, check the if on Apply platform then true otherwise false.
- if not web then check native platform, if iOS then true else false..

kmahmood74 (approved):

Issue #1628, opened 2024-09-07
Need support for `tooltip`


TheNoumanDev:
Need support for a Tooltip widget that allows users to view additional information when hovering or long-pressing on a widget. This would improve the user experience by offering helpful context or explanations for specific elements. Tooltips are essential for enhancing usability, especially in complex interfaces.

something like that.
![image](https://github.com/user-attachments/assets/14387c85-8592-4fe2-9f29-d8e6e8802ff8)

TheNoumanDev:
Default for mobile is: `Tap`
Default for web is: `onHover`

Example EDL:
```yaml
ToolTip:
  message: "This is a tooltip"
  height: 100
  decoration:
    backgroundColor: grey
    borderRadius: 20
    borderColor: red
    borderWidth: 3
  textStyle:
    fontSize: 25
    fontWeight: bold
    color: '#D32F2F'
  child:  
    Text:
      text: I am inside ToolTip.
```

PR #1630, opened 2024-09-12
added ToolTip widget

TheNoumanDev:
ticket: https://github.com/EnsembleUI/ensemble/issues/1628

Example Video & EDL is added in ticket.

once PR gets approved, will create schema PR and update docs.

### DIFF
--- a/modules/ensemble/lib/widget/tooltip.dart
+++ b/modules/ensemble/lib/widget/tooltip.dart
@@ -1,5 +1,7 @@
+import 'package:ensemble/framework/event.dart';
 import 'package:ensemble/framework/extensions.dart';
-import 'package:ensemble/widget/calendar.dart';
+import 'package:ensemble/framework/theme/theme_manager.dart';
+import 'package:ensemble/widget/helpers/controllers.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:ensemble/framework/widget/widget.dart';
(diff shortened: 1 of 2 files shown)


modules/ensemble/lib/widget/tooltip.dart
```diff
@@ -0,0 +1,182 @@
...
+  @override
+  Map<String, Function> methods() {
+    return {
+      'show': () => _controller.show(),
+      'hide': () => _controller.hide(),
+    };
+  }
+}
+
+class ToolTipController extends Controller {
```
kmahmood74, line 69: see if you can extend one of the existing controllers that already capture these properties

modules/ensemble/lib/widget/tooltip.dart
```diff
@@ -0,0 +1,182 @@
...
+    return {
+      'isVisible': () => _controller.isVisible,
+    };
+  }
+
+  @override
+  Map<String, Function> setters() {
+    return {
+      'message': (value) => _controller.message = Utils.optionalString(value),
+      'child': (value) => _controller.child = value,
```
kmahmood74, line 43: everywhere else we call it a `widget` suggest calling it a `widget` instead of `child`

modules/ensemble/lib/widget/tooltip.dart
```diff
@@ -0,0 +1,182 @@
...
+      'textStyle': (value) => _controller.textStyle = Utils.getTextStyle(value),
+      'height': (value) => _controller.height = Utils.optionalDouble(value),
+      'padding': (value) => _controller.padding = Utils.getInsets(value),
+      'margin': (value) => _controller.margin = Utils.getInsets(value),
+      'verticalOffset': (value) => _controller.verticalOffset = Utils.optionalDouble(value),
+      'preferBelow': (value) => _controller.preferBelow = Utils.optionalBool(value),
+      'excludeFromSemantics': (value) => _controller.excludeFromSemantics = Utils.optionalBool(value),
```
kmahmood74, line 50: is this for accessibility?

modules/ensemble/lib/widget/tooltip.dart
TheNoumanDev: Yeah, sorry for oversight, i implemented `BoxController` and handled box decoration and other properties from it.

modules/ensemble/lib/widget/tooltip.dart
TheNoumanDev: Updated.

kmahmood74 (approved):

PR #1642, opened 2024-09-23
small fix related to fontsize in checkbox, switch

TheNoumanDev:
Label in `checkBox` and `Switch` has a property name `fontSize` under `labelStyle` which was not working, else it had a independent property, which shouldn't be the case, So to make it consistent,  fixed that so `fontSize` under `labelStyle` should work.

Example EDL:
```yaml
Checkbox:
  leadingText: hello
  styles:
    labelStyle:
      fontSize: 50
```


modules/ensemble/lib/widget/helpers/form_helper.dart
```diff
@@ -292,7 +292,7 @@ abstract class FormFieldWidgetState<W extends HasController>
     if (widget.controller is FormFieldController) {
       final formController = (widget.controller as FormFieldController);
       return textStyle.copyWith(
-        fontSize: formController.fontSize?.toDouble(),
+        fontSize: formController.labelStyle?.fontSize,
```
kmahmood74, line 295: what if labelStyle is not provided or doesn't have fontSize in it? We should resort to the original in that case

kmahmood74 (changes requested):

modules/ensemble/lib/widget/helpers/form_helper.dart
TheNoumanDev: @kmahmood74 if labelStyle is not specified, then its properties would be null, and as we are using `copyWith` operator with textStyle constructor, if it would be null, it will retain the default fontsize automatically which is 14 pixel.

kmahmood74: you could have the fontSize specified directly on the widget.controller as well. How will that apply?

TheNoumanDev: The fontSize property in FormFieldController seemed redundant to me since labelStyle can handle font size adjustments. 
To simplify, we can consider removing fontSize and prioritizing labelStyle.fontSize in the formFieldTextStyle method, while ensuring the default size is applied if not specified.
if not, I can update to `formController.labelStyle?.fontSize ?? formController.fontSize?.toDouble()`

TheNoumanDev: I’ve reviewed the code, and it appears that all widgets extending FormFieldController do not utilize the fontSize property independently; they pass all values through labelStyle for label as this `FormFieldController` doesn't define textstyles for actual field, so the widget extending it is defining the textStyle separately e.g if you see `TextInput` code, it defines `textStyle` separately so `fontSize` for actual field is not related as far I know.  
The only exception is the Time widget, which has a hard-coded fontSize that isn’t customizable by the user that's why didn't removed the property. Additionally, for the `formFieldTextStyle` getter, only the switch and checkbox widgets are relevant, and they don't contain text beyond the label itself. This reinforces the idea that the fontSize property might be unnecessary in most cases.

if am missing something do let me know.

kmahmood74: Time widget has this code -

```dart
  Text prettyValue(BuildContext context) {
    if (value != null) {
      return Text(
        MaterialLocalizations.of(context)
            .formatTimeOfDay(value!, alwaysUse24HourFormat: use24hFormat),
        style: TextStyle(fontSize: fontSize?.toDouble()),
```

so it does use fontSize?

DateRange defines its own for some reason. But yes it's only the time widget, where is the hardcoded fontSize for time widget?

TheNoumanDev: by hard coded, i meant we are not customizing it based on user interaction, so it is just used bcz it was declared, we can remove it replacing with same function `formFieldTextStyle` to get the default style as well.
And it is using to built the below text in Time widget field and right now we are not offering styles to customize the text in field, if we offer and add functionality for it, we would provide whole `textStyle` which will definitely have `fontSize` along with all other properties. Alone `fontSize` doesn't make sense.
image

kmahmood74: got it, feel free to fix the time widget's fontsize as well as part of this pr

kmahmood74 (approved):

Issue #1629, opened 2024-09-12
Require a Collapsable Widget 


mehsaandev:
Require Collapsable widget which is called Accodion in Flutter

Following are my requirements Screens:
![image](https://github.com/user-attachments/assets/2be067f1-42ca-4bdb-a14d-05c24aff26f0)
![image](https://github.com/user-attachments/assets/4c4effd9-f52d-479f-9925-f81079290c4f)

Here's the Flutter Package for Accordion
 https://pub.dev/packages/accordion

mehsaandev:
here's the EDL for Collapsable Widget of Ensemble React:

`Collapsible:
            headerStyle:
              headerBg: "#09090f"
              borderColor: "#09090f"
              textColor: "rgb(201, 205, 210)"
              fontSize: 17
            contentStyle:
              contentBg: "#09090f"
            expandIconPosition: "end"
            expandIcon: 
              name: Remove
              color: white
            collapseIconPosition: "end"
            collapseIcon: 
              name: Add
              color: white
            items:
              - key: "1"
                label: Testing Question
                children:
                  Text:
                    text: Testing Answer`

TheNoumanDev:
Example EDL: 
```yaml
- Collapsible:
    isAccordion: true
    headerStyle:
      backgroundColor: red
      borderColor: green
      borderColorOpened: red
      backgroundColorOpened: grey
      borderWidth: 5
      borderRadius: 5
    contentStyle:
      backgroundColor: white
      borderColor: green
      borderWidth: 3
      horizontalPadding: 20
    flipRightIconIfOpen: true
    openAndCloseAnimation: true
    paddingBetweenOpenSections: 10
    paddingBetweenClosedSections: 10
    items:
      - header: 
          Text: 
            text: "Section 1 Heading"
        content: 
          Text: 
            text: "Section 1 Content"
        isOpen: true
        leftIcon:
          Icon: 
            name: code
      - header: 
          Text: 
            text: "Section 2 Heading"
        content: 
          Text: 
            text: "Section 2 Content"
        headerBackgroundColor: red
        contentHorizontalPadding: 40
        contentVerticalPadding: 20
        onOpenSection: 
          executeCode:
            body: |
              console.log('section 2 open');
        onCloseSection: |
          console.log("Section 2 closed");
```

amin-nas:
The proposed EDL is not consistent with how we do other widgets. I suggest following changes:

[...]

PR #1645, opened 2024-09-25
added colapsible widget

TheNoumanDev:
requested ticket: https://github.com/EnsembleUI/ensemble/issues/1629

TODO:
- right now  it just supports `items` not `itemTemplate`, will create ticket for that as I tried to implement it for now but it was not happening, maybe I was missing something and it was taking a lot of time. Will work on it in free time.

Example EDL: 
```yaml
Collapsible:
  isAccordion: true
  headerStyle:
    backgroundColor: red
    borderColor: green
    borderColorOpened: red
    backgroundColorOpened: grey
    borderWidth: 5
    borderRadius: 5
  bodyStyle:
    backgroundColor: white
    borderColor: green
    borderWidth: 3
    horizontalPadding: 20
  flipRightIconIfOpen: true
  openAndCloseAnimation: true
  paddingBetweenOpenSections: 10
  paddingBetweenClosedSections: 10
  items:
    - header: 
        Text: 
          text: "Section 1 Heading"
      body: 
        Text: 
          text: "Section 1 body"
      isOpen: true
      leftIcon:
        Icon: 
          name: code
    - header: 
        Text: 
          text: "Section 2 Heading"
      body: 
        Text: 
          text: "Section 2 body"
      headerStyle:
         backgroundColor: green
      bodyStyle:
        backgroundColor: white
        borderColor: green
        borderWidth: 3
        horizontalPadding: 40
        verticalPadding: 80
      onOpenSection: 
        executeCode:
          body: |
            console.log('section 2 open');
      onCloseSection: |
        console.log("Section 2 closed"); 
```

### DIFF
--- a/modules/ensemble/lib/widget/collapsible.dart
+++ b/modules/ensemble/lib/widget/collapsible.dart
@@ -65,9 +65,9 @@ class HeaderStyleComposite extends WidgetCompositeProperty {
   Map<String, Function> methods() => {};
 }
 
-// ContentStyleComposite class to handle content styling
-class ContentStyleComposite extends WidgetCompositeProperty {
-  ContentStyleComposite(super.widgetController) {
+// bodyStyleComposite class to handle body styling
+class bodyStyleComposite extends WidgetCompositeProperty {
+  bodyStyleComposite(super.widgetController) {
     backgroundColor = this.backgroundColor;
     borderColor = this.borderColor;
     borderWidth = this.borderWidth;
@@ -83,8 +83,8 @@ class ContentStyleComposite extends WidgetCompositeProperty {
   double? horizontalPadding;
   double? verticalPadding;
 
-  factory ContentStyleComposite.from(WidgetController controller, dynamic payload) {
-    ContentStyleComposite composite = ContentStyleComposite(controller);
+  factory bodyStyleComposite.from(WidgetController controller, dynamic payload) {
+    bodyStyleComposite composite = bodyStyleComposite(controller);
     if (payload is Map) {
       composite.backgroundColor = Utils.getColor(payload['backgroundColor']);
       composite.borderColor = Utils.getColor(payload['borderColor']);
@@ -134,19 +134,19 @@ class Collapsible extends StatefulWidget
     return {
       'openSections': () => _controller.openSections,
       'headerStyle': () => controller.headerStyle,
-      'contentStyle': () => _controller.contentStyle,
+      'bodyStyle': () => _controller.bodyStyle,
     };
   }
 
   @override
   Map<String, Function> setters() {
     return {
-      'items': (value) => _controller.items = List<YamlMap>.from(value),
+      'items': (value) => _controller.items = List<Map>.from(value),
       'isAccordion': (value) => _controller.isAccordion = Utils.getBool(value, fallback: true),
       'initialOpeningSequenceDelay': (value) =>
           _controller.initialOpeningSequenceDelay = Utils.optionalInt(value),
       'headerStyle': (value) => _controller.headerStyle = HeaderStyleComposite.from(controller, value),
-      'contentStyle': (value) => _controller.contentStyle = ContentStyleComposite.from(controller, value),
+      'bodyStyle': (value) => _controller.bodyStyle = bodyStyleComposite.from(controller, value),
       'leftIcon': (value) => _controller.leftIcon = value,
       'rightIcon': (value) => _controller.rightIcon = value,
       'flipLeftIconIfOpen': (value) =>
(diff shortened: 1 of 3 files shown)


modules/ensemble/lib/widget/widget_registry.dart
```diff
@@ -135,6 +136,7 @@ class WidgetRegistry {
         PopupMenu.type: () => PopupMenu(),
         EnsembleCalendar.type: () => EnsembleCalendar(),
         Countdown.type: () => Countdown(),
+        Collapsible.type: () => Collapsible(),
```
vusters, line 139: why not call it "Accordion"?

modules/ensemble/lib/widget/collapsible.dart
```diff
@@ -0,0 +1,345 @@
...
+  }
+
+  @override
+  Map<String, Function> setters() {
+    return {
+      'items': (value) => _controller.items = List<YamlMap>.from(value),
+      'isAccordion': (value) => _controller.isAccordion = Utils.getBool(value, fallback: true),
+      'initialOpeningSequenceDelay': (value) =>
+          _controller.initialOpeningSequenceDelay = Utils.optionalInt(value),
+      'headerStyle': (value) => _controller.headerStyle = HeaderStyle(
```
vusters, line 84: This won't work. If you set this dynamically it will not trigger binding because it is nested inside. You will need to use something similar to BoxAnimationComposite. That ensure any nested property set will trigger bindings.

But an easier way is to not use headerStyle nor contentStyle. Simply flatten them at the root and use something like headerBackGroundColor (for header) and backgroundColor (for content).

modules/ensemble/lib/widget/collapsible.dart
```diff
@@ -0,0 +1,345 @@
...
+  Map<String, Function> getters() {
+    return {
+      'openSections': () => _controller.openSections,
+    };
+  }
+
+  @override
+  Map<String, Function> setters() {
+    return {
+      'items': (value) => _controller.items = List<YamlMap>.from(value),
```
vusters, line 80: Why don't you just use "children" property? See the other containers. This will ensure the children won't get evaluated until it is ready to be used. This may not work properly.

modules/ensemble/lib/widget/collapsible.dart
```diff
@@ -0,0 +1,345 @@
...
+  }
+
+  void closeSection(int index) {
+    final newOpenSections = List<int>.from(_openSections.value);
+    newOpenSections.remove(index);
+    _openSections.value = newOpenSections;
+  }
+
+  @override
+  void dispose() {
```
vusters, line 181: This is wrong because you don't call super.dispose(). Remove the function if you don't use it.

modules/ensemble/lib/widget/widget_registry.dart
TheNoumanDev: We already have this widget in **Ensemble React** as `Collapsible` widget with a flag name `isAccordion`,  then @kmahmood74 mentioned to keep it consistent with **Ensemble React**.

modules/ensemble/lib/widget/collapsible.dart
TheNoumanDev: @vusters yeah i understood, and fixed that using composite classes as wanted to keep the same structure as react and separate styles for both header and content, tested it, bindings are working now.

modules/ensemble/lib/widget/collapsible.dart
TheNoumanDev: For now, i have tested its working fine but I have seen dropdown example, how it is using children and itemtemplate, I will create one more ticket and PR against it, enhancing it to support itemtemplate as well.

modules/ensemble/lib/widget/collapsible.dart
TheNoumanDev: Fixed @vusters

modules/ensemble/lib/widget/collapsible.dart
vusters: no you want to address this now. Once the schema is out harder to adjust. Also the real reason is children and itemTemplate are special such taht their content won't be evaluated immeidately. You want to only evaluate the content when you are ready to render it

vusters, line 144: use List<Map> in case we support JS later. No reason to use YamlMap

modules/ensemble/lib/widget/collapsible.dart
```diff
@@ -0,0 +1,345 @@
...
+              _buildIcon(widget.controller.leftIcon),
+          rightIcon: _buildIcon(item['rightIcon']) ??
+              _buildIcon(widget.controller.rightIcon),
+          paddingBetweenOpenSections:
+              Utils.optionalDouble(item['paddingBetweenOpenSections']) ?? widget.controller.paddingBetweenOpenSections,
+          paddingBetweenClosedSections:
+              Utils.optionalDouble(item['paddingBetweenClosedSections']) ?? widget.controller.paddingBetweenClosedSections,
+          header: _buildChildWidget(context, item['header']) ??
+              Text('Section $index'),
+          content: _buildChildWidget(context, item['content']) ??
```
vusters, line 343: use item["body"] to be consistent with our syntax (we use header/body often)

vusters (approved):
You can merge it in once you fixed some small misc asks.

Issue #1652, opened 2024-09-27
need support for property on `Carousel` to disable the swipe action


TheNoumanDev:
Usecase: when `Carousel` is being used while creating a onBoarding flow, developer doesn't want the use to just swipe through and pass the whole flow rather developer want to make user go through whole flow and let the Carousel swipe items by pressing buttons.

TheNoumanDev:
ticket: https://github.com/EnsembleUI/ensemble/issues/1652

there could be four possible values for `scrollType` which is `bouncing || clamping || neverScrollable || alwaysScrollable` and default for iOS is `bouncing ` and for other is `clamping`.

Example EDL: 
```yaml
Carousel:
  id: carousel
  styles:
    height: 100
    gap: 20
    layout: auto
    borderColor: blue
    padding: 10
    currentIndex: 0
    indicatorType: rectangle
    indicatorPosition: topCenter
    enlargeCenterPage: true
    indicatorWidth: 10
    indicatorHeight: 2
    indicatorColor: green
    autoplay: true
    autoplayInterval: 3
    scrollType: neverScrollable    # it could be bouncing || clamping || neverScrollable || alwaysScrollable
```

cc: @amin-nas

PR #1653, opened 2024-09-27
added support for property on Carousel to disable the swipe action

TheNoumanDev:
ticket: https://github.com/EnsembleUI/ensemble/issues/1652

there could be four possible values for `scrollType` which is `bouncing || clamping || neverScrollable || alwaysScrollable` and default for iOS is `bouncing ` and for other is `clamping`.

Example EDL: 
```yaml
Carousel:
  id: carousel
  styles:
    height: 100
    gap: 20
    layout: auto
    borderColor: blue
    padding: 10
    currentIndex: 0
    indicatorType: rectangle
    indicatorPosition: topCenter
    enlargeCenterPage: true
    indicatorWidth: 10
    indicatorHeight: 2
    indicatorColor: green
    autoplay: true
    autoplayInterval: 3
    scrollType: neverScrollable    # it could be bouncing || clamping || neverScrollable || alwaysScrollable
```

### DIFF
--- a/modules/ensemble/lib/widget/carousel.dart
+++ b/modules/ensemble/lib/widget/carousel.dart
@@ -101,8 +101,8 @@ class Carousel extends StatefulWidget
       'direction': (value) =>
           _controller.direction = Utils.optionalString(value),
       'cacheKey': (value) => _controller.cacheKey = Utils.optionalString(value),
-      'scrollPhysics': (value) =>
-          _controller.scrollPhysics = CarouselScrollPhysics.values.from(value),
+      'scrollType': (value) =>
+          _controller.scrollType = CarouselScrollPhysics.values.from(value),
     };
   }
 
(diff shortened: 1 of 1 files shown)


modules/ensemble/lib/widget/carousel.dart
```diff
@@ -100,6 +101,8 @@ class Carousel extends StatefulWidget
       'direction': (value) =>
           _controller.direction = Utils.optionalString(value),
       'cacheKey': (value) => _controller.cacheKey = Utils.optionalString(value),
+      'scrollType': (value) =>
+          _controller.scrollType = CarouselScrollPhysics.values.from(value),
```
kmahmood74, line 105: check (CarouselScrollPhysics.values.contains(value))  before setting it. We don't want to throw an error. If the value is not valid, set it to null.

kmahmood74 (approved):
approved with changes required. Please make the change before merging. Also remember to update schema, documentation and kitchen sink example

PR #1656, opened 2024-09-29
enable modules with commands

sharjeelyunus:
commands to enable modules

**ensemble_camera**: `npm run hasCamera`

**ensemble_file_manager** : `npm run hasFiles`

**ensemble_contacts**: `npm run hasContacts`

**ensemble_connect**: `npm run hasConnect`

**ensemble_location**: `npm run hasLocation`

**ensemble_deeplink**: `npm run hasDeeplink`

**notifications**: `npm run hasNotifications`

**ensemble_firebase_analytics** `npm run hasFirebaseAnalytics`

**enable multiple modules with a single command**
```js
npm run enable
```

**Generate Keystore**
```js
npm run generate_keystore storePassword=<password> keyPassword=<password> keyAlias=<alias>
```


starter/scripts/constants.dart
```diff
@@ -0,0 +1,36 @@
...
+# Keep Firebase-related classes
+-keep class com.google.firebase.** { *; }
+-dontwarn com.google.firebase.**
+
+# Keep other necessary classes for Flutter
+-keep class io.flutter.** { *; }
+-dontwarn io.flutter.**
+''';
+
+const connectProguardRules = '''
```
snehmehta, line 16: I believe we can remove this until we update proper rules

starter/android/build.gradle
```diff
@@ -20,10 +20,33 @@ allprojects {
 
 rootProject.buildDir = '../build'
 subprojects {
+    afterEvaluate { project ->
```
snehmehta, line 23: Just to log, can you share which modules still breaks with out this manual override?

This was specifically add for high_chart which we removed, might want to update/remove other modules instead of force update on all package

snehmehta (commented):
I have done a single pass on code based, I'll do another pass with actual running the command.

So, main advantage I see right now is it based on starter so always in sync but uses a lot of string and regex manipulation which we need to find better way.

snehmehta:
Also, this is part does project setup and rest app signing and deployment to respected store will be handle by?

starter/scripts/constants.dart
sharjeelyunus: done

starter/android/build.gradle
sharjeelyunus: I think these modules causes the problems
- Auth
- Camera (QR Code)
- Connect

Not sure if others too, have to create builds 1 by 1 for each to be sure

sharjeelyunus:
> Also, this is part does project setup and rest app signing and deployment to respected store will be handle by?

yes, starter commands are only responsible for enabling the modules, and generating keystore (if needed)

The rest of the process is managed in a separate repository: https://github.com/EnsembleUI/build-system. This repository handles additional setup tasks and executes the necessary commands for the project.

snehmehta (approved):

Issue #1646, opened 2024-09-26
Require Swipe Functionality to ListView Item


mehsaandev:
Requirements:
 - Require a swipe functionality in the ListView in which each item of the list is swipeable and on Swipe left, there should be an option of adding Action Button/Icon by which we can perform any action(e.g. Delete or Edit) on that item of the List
- e.g. images: 
1) Image Before Swipe:
![image](https://github.com/user-attachments/assets/13e560de-58f8-4582-9e89-36668cb9f0ff)

2) Image After Swipe:
![image](https://github.com/user-attachments/assets/43ea5c81-1ec5-4ccc-a125-de6cb654531d)

- After Swipe the Delete Icon is the action Button which will be able to delete that entry on API call.

kmahmood74:
cc @vusters

TheNoumanDev:
PR: https://github.com/EnsembleUI/ensemble/pull/1659

Example EDL:
```yaml
- ListView:
    id: listView
    styles: 
      expanded: true 
      enableSwipe: true # default is false to maintain backward compatability
      swipeAnimationDuration: 200 # miliseconds
      swipeThreshold: 0.2  # (means 20 percent of screen and 1 means 100%)
    swipeWidget:
      FlexRow:
        styles:
          height: ${device.height}
          backgroundColor: black
        children:
          - Icon:
              name: delete
              color: red
              onTap: |
                console.log("hello");
    children:
      - MyRow:
          inputs:
            p: ${users}
      - MyRow:
          inputs:
            p: ${users}
      - MyRow:
          inputs:
            p: ${users}
```

kmahmood74:
EDL looks good, thanks @TheNoumanDev

PR #1659, opened 2024-09-30
added support for swipeWidget on swipe in listView Item

TheNoumanDev:
ticket: https://github.com/EnsembleUI/ensemble/issues/1646

Example EDL:
```yaml
View:
  header:
    title: "ListView: Slidable Example"
  onLoad:
    invokeAPI:
      name: getPeople

  body:
    Column:
      children:
        - ListView:
            id: listView
            styles: { expanded: true }
            item-template:
              data: ${getPeople.body.users}
              name: users
              template:
                Slidable:
                  styles:
                    direction: horizontal
                    closeOnScroll: true
                    DragStartBehavior: start
                    useTextDirection: true
                  startContent:
                    options:
                      extentRatio: 0.5
                      openThreshold: 0.25
                      motion: behind
                    children:
                      - icon: star
                        backgroundColor: red
                        foregroundColor: green
                        borderRadius: 20
                      - label: Cancel
                        onPressed:
                          showToast:
                            message: User deleted
                  endContent:
                    children:
                      - icon: star
                        label: Favorite
                        backgroundColor: yellow
                        onPressed:
                          showToast:
                            message: Added to favorites
                  child:
                    MyRow:
                      inputs:
                        p: ${users}

MyRow:
  inputs:
    - p
  body:
    Column:
      children:
        - Row:
            styles:
              gap: 12
              crossAxis: center
              padding: 8 0
            children:
              - Image:
                  styles: { width: 50, height: 50, borderRadius: 50 }
                  source: ${p.image}
              - Column:
                  styles: { gap: 2 }
                  children:
                    - Text:
                        styles: { fontWeight: bold }
                        text: ${p.firstName} ${p.lastName}
                    - Text:
                        text: ${p.email}
              - Spacer
              - Icon:
                  icon: chevron_right
                  styles: { size: 20, color: black }

API:
  getPeople:
    uri: https://dummyjson.com/users
    method: GET
```

will add schema PR, docs and kitchen SInk once PR gets approved before merging it,


vusters:
1. Your swipeWidget will not work. The reason it seems to work is because you simply print out to console. Each ListView's item (i.e. each children and each item template) has to be wrapped inside a Swipeable, otherwise how would you render say a Delete button for each and can tell which item you actually want to delete on? A simpler way to think about this is how would I display a unique piece of text differently for each item's swipe out section? You can't.

2. Never write code if you don't have to. It's a liability. Use this package that has 5.5K stars and 100% - mean well tested https://pub.dev/packages/flutter_slidable

3. What you want to implement is probably a generic **Card** widget that can swipe left/right/up/down to expose more content. To keep it simple for now let's just call it "Slidable" and follow the package structure (it's a decent structure)
```
ListView:
  itemTemplate:
    data: ...
    name: item
    template:
      Slidable:
        body: <your_widget_to_display_as_default_state>
        startContent: <widget_to_display_on_swipe_right>
        endContent: <widget_to_display_on_swipe_left>
```

Use start/end vs left/right to account for RTL languages.

kmahmood74:
@TheNoumanDev please make the changes that Vu suggested

TheNoumanDev:
> @TheNoumanDev please make the changes that Vu suggested

On it.

kmahmood74:
@TheNoumanDev any update?

TheNoumanDev:
@kmahmood74 Sorry for delay, was stuck with KPN task.
@vusters PTAL. Thanks

modules/ensemble/lib/layout/list_view.dart
```diff
@@ -133,22 +124,16 @@ class ListViewController extends BoxLayoutController {
...
-  EnsembleAction? onSwipe;
-
   void _bind(ListViewState state) {
     widgetState = state;
   }
 }
 
 class ListViewState extends EWidgetState<ListView>
-    with TickerProviderStateMixin, TemplatedWidgetState, HasChildren<ListView> {
+    with TemplatedWidgetState, HasChildren<ListView> {
```
vusters: Just to confirm, there is no code changes in list_view right? This is just a revert to the original version?

modules/ensemble/lib/widget/slidable.dart
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  @override
+  Map<String, Function> methods() {
+    return { };
+  }
+}
+
+class ActionPaneOptionsComposite {
+  ActionPaneOptionsComposite() {
```
vusters, line 64: dont' repeat the code in here.

modules/ensemble/lib/widget/slidable.dart
```diff
@@ -0,0 +1,225 @@
...
+  @override
+  EnsembleSlidableController get controller => _controller;
+
+  @override
+  State<StatefulWidget> createState() => EnsembleSlidableState();
+
+  @override
+  Map<String, Function> getters() {
+    return {
+      'startContent': () => _controller.startContent,
```
vusters, line 37: prob should mention this before - we need more descriptive name: startDrawer/endDrawer? @amin-nas

modules/ensemble/lib/widget/slidable.dart
```diff
@@ -0,0 +1,225 @@
...
+    };
+  }
+
+  @override
+  Map<String, Function> methods() {
+    return { };
+  }
+}
+
+class ActionPaneOptionsComposite {
```
vusters, line 63: This has to extend a base composite for binding to work.

modules/ensemble/lib/widget/slidable.dart
```diff
@@ -0,0 +1,225 @@
...
+        color: iconModel.color,
+      ).icon;
+    }
+
+    return iconframework.Icon(value).icon;
+  }
+
+  Widget _buildChildWidget(BuildContext context, dynamic widgetDefinition) {
+    if (widgetDefinition == null) {
+      return Container();
```
vusters, line 216: use ShrinkBox.empty()

modules/ensemble/lib/layout/list_view.dart
TheNoumanDev: Yes that is outdated, there is no changes in listView.

modules/ensemble/lib/widget/slidable.dart
TheNoumanDev: Updated!

modules/ensemble/lib/widget/slidable.dart
TheNoumanDev: Updated it to `startDrawer/endDrawer` for now, do let me if something better came up😅

modules/ensemble/lib/widget/slidable.dart
```diff
@@ -136,61 +97,72 @@ class EnsembleSlidableState extends EWidgetState<EnsembleSlidable> {
...
-        return const SizedBox.shrink(); // Return an empty widget if both are missing
+        return const SizedBox
+            .shrink(); // Return an empty widget if both are missing
       }
 
       return SlidableAction(
         onPressed: (context) {
           if (action['onPressed'] != null) {
-            final onPressedAction = EnsembleAction.from(action['onPressed'], initiator: widget);
+            final onPressedAction =
```
vusters, line 143: This certainly can return null which will throw an except below. Check for it and never never ever use force exclamation like "onPressAction!". If you have to do that, something is wrong.

vusters (approved):
You are good to check if AFTER addressing the stated.

modules/ensemble/lib/widget/slidable.dart
TheNoumanDev: yup got the point, thanks, and will take care in future as well, but the same has been implemented at every action, maybe we need to keep updating such mistake while fixing other stuff.

Issue #1663, opened 2024-10-01
Require multiple Images Upload Support in uploadFiles action


mehsaandev:
Requirements:
- Require Multiple Upload Image Support in uploadFiles action
- Each file should have it's property name
- Upload File should support multipart/form-data 

Sample Postman Request and Response Config:
![image](https://github.com/user-attachments/assets/286ed73d-c8c7-49bd-8177-040f1060306c)

Sample API for testing:
**https://api-proloan.prospertech.io/account/document**
Request Type: **Post**

snehmehta:
Sample EDL

```yaml
View:

  styles:
    scrollableView: true


  header:
    title: Home
  onLoad: |
    ensemble.storage.files = {};

  body:
    Column:
      styles:
        padding: 24
        gap: 8
      children:

        - Button:
            label: pick front
            onTap:
              pickFiles:
                id: filePicker
                onComplete: |-
                  ensemble.storage.files.document_front = filePicker.files[0];

        - Button:
            label: pick back
            onTap:
              pickFiles:
                id: filePickerback
                onComplete: |-
                  ensemble.storage.files.document_back = filePickerback.files[0];

        - Button:
            label: pick selfie
            onTap:
              pickFiles:
                id: filePickerselfie
                onComplete: |-
                  ensemble.storage.files.selfie = filePickerselfie.files[0];
                  
        - Button:
            label: Upload files
            onTap:
              uploadFiles:
                files: ${ensemble.storage.files}
                uploadApi: fileUploadApi


API:
  fileUploadApi:
    url: https://api-proloan.prospertech.io/account/document
    method: POST
    

```

PR #1666, opened 2024-10-02
file upload handles map and multiple fieldName

snehmehta:
Ticket: https://github.com/EnsembleUI/ensemble/issues/1663

### DIFF
--- a/modules/ensemble/lib/action/upload_files_action.dart
+++ b/modules/ensemble/lib/action/upload_files_action.dart
@@ -145,6 +145,7 @@ Future<void> uploadFiles({
           : (error) =>
               ScreenController().executeAction(context, action.onError!),
       progressCallback: (progress) {
+        if (action.id == null) return;
         fileResponse?.setProgress(taskId, progress);
         scopeManager?.dispatch(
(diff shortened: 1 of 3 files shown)


modules/ensemble/lib/action/upload_files_action.dart
```diff
@@ -145,6 +145,7 @@ Future<void> uploadFiles({
           : (error) =>
               ScreenController().executeAction(context, action.onError!),
       progressCallback: (progress) {
+        if (action.id == null) return;
```
kmahmood74, line 148: FileUploadAction.id seems to be required but is defined as optional in FileUploadAction class. Why not just make it required so you don't have to make these checks?

kmahmood74 (approved):
approved with one comment

kmahmood74:
please remember to update documentation, schema and add a kitchen sink example

modules/ensemble/lib/action/upload_files_action.dart
snehmehta: Id is not required. It is used when user want to bind something. Like here if user want to listen to progress of uploading.

I have put check already, in rest of file upload. I saw this was missing throwing exceptions in logs saw just fixed it.

Issue #1686, opened 2024-10-10
Require support for List of images for multiple files on uploadFiles action


mehsaandev:
**Currently we support the following format to upload files using uploadFiles action:**
- An Object having Multiple Objects of Files in a key value pair format
e.g. 
```{
firstFile: { fileData },
secondFile: {fileData},
thirdFile: {fileData}
}
```
- Each key of file can only have file, in the example above firstFile key can only have 1 file.

**New Requirements:**
- Now I require some change in the format in which I should be able to add List of files with each key.
- e.g.
```
{
firstFile: [{ fileData },{ fileData },{ fileData }]
secondFile: [{ fileData },{ fileData },{ fileData }]
thirdFile: [{ fileData },{ fileData },{ fileData }]
```
- Each key should should store the list of Files

PR #1687, opened 2024-10-11
camera blob default value, uploadFiles support multiple file

snehmehta:
Ticket: https://github.com/EnsembleUI/ensemble/issues/1686


kmahmood74 (approved):

kmahmood74 (changes requested):
I think at the least, we should specify the extension if possible? if the server is storing as files (most likely), it's quite critical to have an extension so that when the file is served, it automatically gets its content type. We cannot hardcode 'data' as the filename for every image captured on flutter web. Best is to get the filename from the dev in EDL and then add extension to it if you have it.

kmahmood74 (approved):
please just make the filename unique as I commented above and merge, approving

Issue #1637, opened 2024-09-18
Need native support for Bluetooth/BLE communication with external devices


reubensjohn:
It would be great if you can provide native support for Bluetooth/BLE communication within the Ensemble framework. The app I am developing has to get data from an external BLE device in real-time

kmahmood74:
@reubensjohn do you need BLE or Bluetooth? the two are quite different

kmahmood74:
both BLE and Bluetooth are needed. Assigning to @snehmehta

kmahmood74:
@snehmehta please add kitchen sink example, documentation and schema for this.

kmahmood74:
what's the yaml for it?

PR #1693, opened 2024-10-15
Bluetooth module: Support for BLE devices

snehmehta:
issue: https://github.com/EnsembleUI/ensemble/issues/1637

### DIFF
--- a/modules/ensemble/lib/action/bluetooth_action.dart
+++ b/modules/ensemble/lib/action/bluetooth_action.dart
@@ -46,12 +46,50 @@ class InitializeBluetoothAction extends EnsembleAction {
           ));
     }
 
-    // TODO: ON page pop cancel stream
-    // bluetoothManager.init(
-    //   context: context,
-    //   initiator: initiator,
-    //   onDataStream: onDataStream,
-    // );
+    bluetoothManager.init(
+      context: context,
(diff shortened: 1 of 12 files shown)


modules/ensemble/lib/action/bluetooth_action.dart
```diff
@@ -0,0 +1,189 @@
...
+
+  factory ConnectBluetoothAction.fromYaml(
+      {Invokable? initiator, Map? payload}) {
+    return ConnectBluetoothAction(
+      initiator: initiator,
+      inputs: Utils.getMap(payload?['inputs']),
+      onError: EnsembleAction.from(payload?['onError'], initiator: initiator),
+      onDataStream:
+          EnsembleAction.from(payload?['onDataStream'], initiator: initiator),
+      deviceId: Utils.getString(payload?['deviceId'], fallback: 'fallback'),
```
vusters, line 134: should throw error if deviceId is not specified instead.

vusters (approved):

modules/ensemble/lib/action/bluetooth_action.dart
snehmehta: yes, updating the negative flow right now.

PR #1716, opened 2024-11-05
Device orientation

TheNoumanDev:


modules/ensemble/lib/framework/device.dart
```diff
@@ -13,22 +14,95 @@ import 'package:ensemble_ts_interpreter/invokables/invokable.dart';
...
 import 'package:intl/intl.dart';
 
-/// get device information as well as requesting device permissions
-class Device
-    with
-        Invokable,
-        MediaQueryCapability,
-        LocationCapability,
-        DeviceInfoCapability {
+class Device with Invokable, MediaQueryCapability, LocationCapability, DeviceInfoCapability, WidgetsBindingObserver {
```
vusters, line 21: Setup your formatter to property format code so it is consistent with everyone else. Your review reformats a lot of the code and it shows more changes than needed

modules/ensemble/lib/framework/data_context.dart
```diff
@@ -98,6 +98,20 @@ class DataContext implements Context {
     if (_contextMap['device'] == null) {
       _contextMap['device'] = Device();
     }
+
+    // Always set up the listener
+    print('👂 Setting up device update listener...');
```
vusters, line 103: This doesn't make sense. What is it for? DataContext already points to Device class which handle listening for changes.

modules/ensemble/lib/framework/device.dart
```diff
@@ -13,22 +14,95 @@ import 'package:ensemble_ts_interpreter/invokables/invokable.dart';
...
+  static bool _isInitialized = false;
+  bool _isHandlingChange = false;
+  MediaQueryData? _lastReportedData;
+
+  static final StreamController<MediaQueryData> _deviceUpdateController = 
+      StreamController<MediaQueryData>.broadcast();
+  static Stream<MediaQueryData> get onDeviceUpdate => _deviceUpdateController.stream;
+  
+  Device._internal() {
+    if (!_isInitialized) {
```
vusters, line 32: Device is a singleton, meaning there's only 1 instance of it. You don't need to check if it's initialized or not.

modules/ensemble/lib/framework/device.dart
```diff
@@ -13,22 +14,95 @@ import 'package:ensemble_ts_interpreter/invokables/invokable.dart';
...
+    }
+  }
+
+  factory Device() => _instance;
+
+  @override
+  void didChangeMetrics() {
+    print('📏 Metrics changed - handling update...');
+    if (!_isHandlingChange) {
+      WidgetsBinding.instance.addPostFrameCallback((_) => _handleMediaQueryChange());
```
vusters, line 49: Why do addPostFrameCallback() here for? Just call handleMediaqueryChange() directly.

modules/ensemble/lib/framework/device.dart
```diff
@@ -13,22 +14,95 @@ import 'package:ensemble_ts_interpreter/invokables/invokable.dart';
...
+      _isInitialized = true;
+    }
+  }
+
+  factory Device() => _instance;
+
+  @override
+  void didChangeMetrics() {
+    print('📏 Metrics changed - handling update...');
+    if (!_isHandlingChange) {
```
vusters, line 48: what is this flag for, preventing multiple concurrent call? There is no async code here so that's not going to happen.

modules/ensemble/lib/framework/device.dart
```diff
@@ -13,22 +14,95 @@ import 'package:ensemble_ts_interpreter/invokables/invokable.dart';
...
+      
+      // Check for real changes
+      if (_hasSignificantChanges(_lastReportedData, newData)) {
+        
+        _lastReportedData = newData;
+        MediaQueryCapability.data = newData;
+        
+        // Broadcast the change
+        print('📢 Broadcasting device update...');
+        _deviceUpdateController.add(newData);
```
vusters, line 74: Why? Why don't you set the orientation/width/height directly?

modules/ensemble/lib/framework/device.dart
```diff
@@ -13,22 +14,95 @@ import 'package:ensemble_ts_interpreter/invokables/invokable.dart';
...
+      }
+    } catch (e) {
+      print('💥 Error handling MediaQuery change: $e');
+    } finally {
+      _isHandlingChange = false;
+    }
+  }
+  
+  bool _hasSignificantChanges(MediaQueryData? oldData, MediaQueryData newData) {
```
vusters, line 83: Maybe you use this for testing but why not just remove this function and change it to "orienttation != newData.orientation || size != newData.size". Simpler and less code to look through

vusters (changes requested):

kmahmood74:
@snehmehta please review this

vusters (approved):

sharjeelyunus (approved):

Issue #1580, opened 2024-08-19
Slider That Allows You to Select a Range


bigbunny2:
I need to use a slider for one of my screens but there's no option to have more than one value for the slider. For example if I wanted to select a range between 10 and 30, there's no way to do that.

For a solution, I just want a way to be able to have two values!

I could definitely use JavaScript but I think that's doing a lot more than is necessary.

For context, the screen I'm making needs a slider because the user needs to select a distance they're comfortable with, and a price range they're comfortable with.

PR #1723, opened 2024-11-14
Range slider

rmahmadkhan:
Ticket #1580 

### Example code

```
Slider:
  id: sliderOne
  label: Range Slider
  min: 0
  max: 10
  divisions: 10
  initialValue: 1 4
  isRange: true
  onChange: |
    sliderOne.value.text = "Slider is set to " + this.value;
```


rmahmadkhan:
There is an issue of theme with the slider labels which needs to be resolved before merging.

Screenshot 2024-11-14 at 5 11 28 PM

rmahmadkhan:
I also see a PR #1725 related to more customizations in simple Silder widget. Since we would want the maximum similar customizations, should i wait for the that to merge and then use those here as well?

cc @kmahmood74 @TheNoumanDev

modules/ensemble/lib/util/utils.dart
```diff
@@ -1,25 +1,22 @@
...
+import 'package:ensemble_ts_interpreter/invokables/UserLocale.dart';
 import 'package:ensemble_ts_interpreter/invokables/invokableprimitives.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_i18n/flutter_i18n.dart';
+import 'package:flutter_localizations/flutter_localizations.dart';
 import 'package:google_fonts/google_fonts.dart';
+import 'package:path/path.dart' as p;
 import 'package:yaml/yaml.dart';
-import 'package:flutter_localizations/flutter_localizations.dart';
```
TheNoumanDev, line 20: Please keep the changes minimal in PR. This shows removed and re-added imports that are just reordered rather than actually changed.

modules/ensemble/lib/widget/input/slider.dart
```diff
@@ -36,8 +35,8 @@ class EnsembleSlider extends StatefulWidget
   @override
   Map<String, Function> setters() {
     return {
-      'initialValue': (value) =>
-          _controller.value = Utils.optionalDouble(value) ?? 0,
+      'initialValue': (value) => _controller.value =
+          Utils.getRangeValues(value) ?? const RangeValues(0, 0),
```
TheNoumanDev, line 39: The getter implementation for `RangeValues` needs to expose the individual `start` and `end` values since returning only the RangeValues object doesn't allow direct access to these values.

modules/ensemble/lib/widget/input/slider.dart
```diff
@@ -124,27 +128,56 @@ class SliderState extends FormFieldWidgetState<EnsembleSlider> {
...
+                            if (widget.controller.onChange != null) {
+                              ScreenController().executeAction(
+                                context,
+                                widget.controller.onChange!,
+                                event: EnsembleEvent(widget),
+                              );
+                            }
+                          }
+                        : null,
+                  ),
```
TheNoumanDev, line 180: Please remove the duplicate code which can be converted to method and use them.

modules/ensemble/lib/widget/widget_registry.dart
```diff
@@ -51,25 +48,28 @@ import 'package:ensemble/widget/markdown.dart';
...
-import 'package:ensemble/widget/slidable.dart';
-import 'package:ensemble/widget/accordion.dart';
 import 'package:ensemble/widget/visualization/barchart.dart';
 import 'package:ensemble/widget/visualization/chart_js/chart_js.dart';
 import 'package:ensemble/widget/visualization/line_area_chart.dart';
 import 'package:ensemble/widget/visualization/topology_chart.dart';
 import 'package:ensemble/widget/webview/webview.dart';
-import 'package:ensemble/widget/youtube/youtube.dart';
 import 'package:ensemble/widget/weeklyscheduler.dart';
+import 'package:ensemble/widget/youtube/youtube.dart';
```
TheNoumanDev, line 72: also i dont see any required change in this file, please update the changes if not required.

TheNoumanDev (changes requested):

modules/ensemble/lib/util/utils.dart
rmahmadkhan: that was done by IDE import rearrangment. I will make it like previous

modules/ensemble/lib/widget/input/slider.dart
rmahmadkhan: I intentionally did not do that. Since we are using RangeValue for for both, a user can get start or end value through `value`. 
A thing we can do instead is make value as single `value` getter (in case start == end), and also add `start` and `end` getters. wdys?

modules/ensemble/lib/widget/input/slider.dart
rmahmadkhan: Move whole onChanged function to a separate function

PR #1731, opened 2024-11-21
refactor webview with new package to support header override

TheNoumanDev:
replaced 'https://pub.dev/packages/webview_flutter' with 'https://pub.dev/packages/flutter_inappwebview' to support headers override.
Added following new properties:
```yaml
  WebView:
    styles:
      height: ${device.height - device.safeAreaBottom - device.safeAreaTop - 10 }
    headers:
      "Authorization": "Bearer abc123456789"
    headerOverrideRules:
      - urlPattern: 'https://example.com/api'
        matchType: CONTAINS  # CONTAINS, EXACT, REGEX
        headers:
          Custom-Header: CustomValue123
        mergeExisting: true
```

Test with following android and web, things worked fine by my end, iOS still needs some testing.
- Kitchen sink Example worked fine (Android, web)
- Kpn login (the top error is coming, bcz sometimes It have options API calls which don't have header but login works as it has that error):

https://github.com/user-attachments/assets/af7aaefa-a09a-4264-a5ea-80553222e06e

- Trip App worked fine on android with cookies being passed, but on web I wasn't able to login , it was given error

https://github.com/user-attachments/assets/07cf0916-d0df-4215-a665-5d89385d9e78


modules/ensemble/lib/widget/webview/webview.dart
```diff
@@ -50,13 +50,33 @@ class EnsembleWebView extends StatefulWidget
...
       'allowedLaunchSchemes': (value) => _controller.schemes =
           Utils.getList<String>(value) ??
               EnsembleWebViewController.defaultSchemes,
     };
   }
+
+  List<HeaderOverrideRule> parseHeaderRules(YamlList yamlList) {
+    return yamlList.map((rule) {
+      Map<String, String> headers = {};
+      (rule['headers'] as YamlMap).forEach((key, value) {
```
kmahmood74, line 66: by assuming it's a YamlMap, we make it impossible for someone to set this value in js. Please just use Map

modules/ensemble/lib/widget/webview/web/webviewstate.dart
```diff
@@ -17,6 +17,7 @@ import 'package:flutter/material.dart';
 class WebViewState extends EWidgetState<EnsembleWebView> {
   final IFrameElement _iframeElement = IFrameElement();
   HtmlElementView? htmlView;
+  final String viewId = 'iframeElement-${DateTime.now().millisecondsSinceEpoch}';
```
kmahmood74, line 20: why make it a unique viewId, is it causing memory leaks?

modules/ensemble/lib/widget/webview/web/webviewstate.dart
```diff
@@ -29,18 +30,18 @@ class WebViewState extends EWidgetState<EnsembleWebView> {
...
-      'iframeElement',
+      viewId,
       (int viewId) => _iframeElement,
     );
 
     return HtmlElementView(
       key: UniqueKey(),
-      viewType: 'iframeElement',
+      viewType: viewId,
     );
```
kmahmood74, line 45: we should add a dispose method here and dispose off the htmlView and iframeElement. That should get rid of the memory leak

modules/ensemble/lib/widget/webview/native/webviewstate.dart
```diff
@@ -183,40 +54,247 @@ class WebViewState extends EWidgetState<EnsembleWebView> with CookieMethods {
...
+    if (widget.controller.url == null) {
+      return;
+    }
+    final mainUrl = Uri.parse(widget.controller.url!);
+    if (cookie != null) {
+      final cookieDomain = cookie.domain ?? mainUrl.host;
+
+      // Set cookie for main domain
+      await widget.controller.cookieManager.setCookie(
+        url: Uri.parse('https://$cookieDomain'),
```
kmahmood74, line 104: we can't assume that it's always https. Please get the protocol from the url as well and don't hardcode it

modules/ensemble/lib/widget/webview/native/webviewstate.dart
```diff
@@ -183,40 +54,247 @@ class WebViewState extends EWidgetState<EnsembleWebView> with CookieMethods {
...
+      );
+    }
+
+    // Handle additional cookies
+    if (widget.controller.cookies.isNotEmpty) {
+      for (var cookieData in widget.controller.cookies) {
+        final cookieDomain = cookieData['domain'] ?? mainUrl.host;
+
+        await widget.controller.cookieManager.setCookie(
+          url: Uri.parse('https://$cookieDomain'),
```
kmahmood74, line 122: same here

kmahmood74 (changes requested):
pretty big change, let's have a zoom call to go over it

amin-nas:
@TheNoumanDev different error now
```
 ✝  Ensemble/ensemble/starter   webview0x45±  flutter run
Launching lib/main.dart on iPhone 16 Pro in debug mode...
Running pod install...                                              3.4s
Running Xcode build...                                                  
Xcode build done.                                           51.2s
Failed to build iOS app
Error (Xcode): ../modules/ensemble/lib/widget/webview/webview.dart:125:64: Error: The argument type 'Uri' can't be assigned to the parameter type 'WebUri?'.


Could not build the application for the simulator.
Error launching application on iPhone 16 Pro.
```

PR #1740, opened 2024-11-26
[1741] updated api to support urlencoded content type

TheNoumanDev:
before we were just using `application/json` unless some other `conten-Type` is not specified but `application/x-www-form-urlencoded` requires special type of formating like `key1=value1&key2=value2` to make it work.

Ticket: https://github.com/EnsembleUI/ensemble/issues/1741

### DIFF
--- a/modules/ensemble/lib/framework/apiproviders/http_api_provider.dart
+++ b/modules/ensemble/lib/framework/apiproviders/http_api_provider.dart
@@ -77,9 +77,9 @@ class HTTPAPIProvider extends APIProvider {
       
       if (contentType == 'application/x-www-form-urlencoded') {
         // For form-urlencoded, convert body to query string format
-        if (api['body'] is YamlMap) {
+        if (api['body'] is Map) {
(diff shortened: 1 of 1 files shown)


modules/ensemble/lib/framework/apiproviders/http_api_provider.dart
```diff
@@ -73,15 +73,33 @@ class HTTPAPIProvider extends APIProvider {
...
-        bodyPayload = json.encode(eContext.eval(api['body']));
-
-        // set Content-Type as json but don't override user's value if exists
-        if (headers['Content-Type'] == null) {
-          headers['Content-Type'] = 'application/json';
+      final contentType = headers['Content-Type']?.toLowerCase() ?? '';
+      
+      if (contentType == 'application/x-www-form-urlencoded') {
+        // For form-urlencoded, convert body to query string format
+        if (api['body'] is YamlMap) {
```
vusters, line 80: use Map

vusters (approved):

PR #1762, opened 2024-12-10
Added dismissible property to showDialog action

mehsaandev:


modules/ensemble/lib/action/dialog_actions.dart
```diff
@@ -18,11 +18,13 @@ class ShowDialogAction extends EnsembleAction {
   ShowDialogAction({
     super.initiator,
     required this.body,
+    this.dismissible = true,
```
kmahmood74, line 21: don't hardcode the default here, you already have a default when reading the payload. Bad design to hardcode in multiple places

kmahmood74 (approved):
approved with a minor change

kmahmood74:
remember to - 

1. update schema
2. add this as an example to kitchen sink's existing dialog example
3. update documentation

modules/ensemble/lib/action/dialog_actions.dart
mehsaandev: removed hard coded default value

PR #1777, opened 2024-12-18
Updated Share Action: added file sharing support

mehsaandev:


mehsaandev:
@kmahmood74 can you please take a look at this?

modules/ensemble/lib/action/misc_action.dart
```diff
@@ -57,29 +57,77 @@ class CopyToClipboardAction extends EnsembleAction {
...
+    final sharePositionOrigin = box!.localToGlobal(Offset.zero) & box.size;
+    final evaluatedText = scopeManager.dataContext.eval(_text);
+    final evaluatedTitle =
+        Utils.optionalString(scopeManager.dataContext.eval(_title));
+    List<XFile>? xFiles;
+    if (_files != null) {
+      final evaluatedFiles = scopeManager.dataContext.eval(_files);
+      if (evaluatedFiles != null) {
+        final filesList =
+            evaluatedFiles is List ? evaluatedFiles : [evaluatedFiles];
```
kmahmood74, line 95: why would it not be a list?

modules/ensemble/lib/action/misc_action.dart
```diff
@@ -57,29 +57,77 @@ class CopyToClipboardAction extends EnsembleAction {
...
 
-    Share.share(
-      scopeManager.dataContext.eval(_text),
-      subject: Utils.optionalString(scopeManager.dataContext.eval(_title)),
-      sharePositionOrigin: box!.localToGlobal(Offset.zero) & box.size,
-    );
-    return Future.value(null);
+        xFiles = filesList.map((file) {
+          // Handle case where file is a Map containing file info
+          if (file is Map) {
```
kmahmood74, line 99: when is it a Map vs a String? how does it work across web, android and ios?

kmahmood74:
@mehsaandev please don't just ignore my comments. Respond and let's get this done

modules/ensemble/lib/action/misc_action.dart
mehsaandev: Added a separate function to handle the case where the file is a URL.

modules/ensemble/lib/action/misc_action.dart
mehsaandev: That was just an edge case like if the user added one file instead of a file array.

Like if user add `filerPicker.files[0]` instead of `filerPicker.files`

kmahmood74:
@snehmehta you know this area well, could you please review this.

modules/ensemble/lib/action/misc_action.dart
```diff
@@ -57,29 +58,169 @@ class CopyToClipboardAction extends EnsembleAction {
...
+
+    return ShareAction(
+      payload['text'],
+      title: payload['title']?.toString(),
+      files: payload['files'],
+    );
+  }
+
+  // Helper method to determine MIME type
+  String getMimeType(String? extension) {
```
snehmehta, line 85: https://github.com/EnsembleUI/ensemble/blob/959406bdff99df95a51dd924e5957a823ee1db5e/modules/ensemble/lib/util/upload_utils.dart#L60

you can lookup mimeType like this instead of manual mapping.

snehmehta (changes requested):

modules/ensemble/lib/action/misc_action.dart
mehsaandev: Updated to this.

snehmehta (approved):
Approach feel free to merge it once testing

PR #1790, opened 2024-12-25
Update ensemble_schema.json for existing screens to validate against it

v-krupsky:
Changes to make existing screen definitions validate against the schema.

Main changes:
- Widget can now be a string/null (there is usage of widgets in a string/null fashion)
- Widget can be just an object to support user-defined widgets

- Reworked Conditional widget to support the syntax of existing screens
- Added Firestore API definition
- Fixed some properties to be values or formulas - not the full list but just the things that were breaking in the validation
- Added definitions for the following Action types: Log Event, Get Device Token, Execute Action Group, Execute Conditional Action

Had to update some screens - attaching the full list of screens and script I'm using to validate screens against schema.
[attachment for EnsembleUI pull request.zip](https://github.com/user-attachments/files/18243485/attachment.for.EnsembleUI.pull.request.zip)

### DIFF
--- a/modules/ensemble/assets/schema/ensemble_schema.json
+++ b/modules/ensemble/assets/schema/ensemble_schema.json
@@ -18,7 +18,7 @@
           "description": "Configure the application header",
           "properties": {
             "title": {
-              "oneOf": [
+              "anyOf": [
                 {
                   "type": "string"
                 },
@@ -112,9 +112,9 @@
                     "type": "number"
                   }
                 },
-                "fixedContent": {                  
-                    "$ref": "#/$defs/Widgets",
-                    "description": "Provide content to be fixed in the footer"                  
+                "fixedContent": {
+                  "$ref": "#/$defs/Widget",
+                  "description": "Provide content to be fixed in the footer"
                 },
                 "onMinSize": {
                   "$ref": "#/$defs/Action-payload",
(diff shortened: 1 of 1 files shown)


sharjeelyunus:
Hi @v-krupsky,

The `ensemble_schema.json` is no longer in use. We now have TypeScript interfaces that automatically generate the schema in a different repo.

@kmahmood74 shouldn't we remove the ensemble_schema.json file from here to avoid confusing people?

kmahmood74:
> Hi @v-krupsky,
> 
> The `ensemble_schema.json` is no longer in use. We now have TypeScript interfaces in a different repo that automatically generate the schema for studio.
> 
> @kmahmood74 shouldn't we remove the ensemble_schema.json file from here to avoid confusing people?

As @sharjeelyunus noted, the `ensemble_schema.json` is an output of the typescript schema. Even that is not the one here so we should remove from this repo.

v-krupsky:
Thank you @sharjeelyunus @kmahmood74 - could you please advise where and how can I get the current schema?

sharjeelyunus:
schemas can be generated from here: https://github.com/EnsembleUI/ensemble-web-studio/tree/main/src/schema
currently they are in a private repo, please let us know what do need so we can help you accordingly

Closing this for now

PR #1799, opened 2024-12-31
added `saveFile` action

mehsaandev:


kmahmood74:
@mehsaandev no reviewer? is this draft?

mehsaandev:
Yes this is draft, finishing it.

modules/ensemble/lib/action/save_file.dart
```diff
@@ -0,0 +1,190 @@
...
+          fileBytes = base64Decode(blobData); // Decode base64
+        } else if (blobData is List<int>) {
+          fileBytes = Uint8List.fromList(blobData);
+        } else {
+          throw Exception(
+              'Invalid blob data format. Must be base64 or List<int>.');
+        }
+      } else if (source != null) {
+        // If blobData is not available, check for source (network URL)
+        Dio dio = Dio();
```
snehmehta, line 64: why use dio when we have http package for network handling?

modules/ensemble/lib/action/save_file.dart
```diff
@@ -0,0 +1,190 @@
...
+        await _notifyFileSystem(filePath);
+      }
+    } catch (e) {
+      throw Exception('Failed to save document: $e');
+    }
+  }
+
+  /// Notify the Android media store about the new file
+  Future<void> _notifyFileSystem(String filePath) async {
+    try {
```
snehmehta, line 144: Running a process is not recommended if possible please find another way

modules/ensemble/lib/action/save_file.dart
```diff
@@ -0,0 +1,190 @@
...
+        await _saveDocumentToDocumentsFolder(fileName!, fileBytes);
+      }
+    } catch (e) {
+      throw Exception('Failed to save file: $e');
+    }
+  }
+
+  Future<void> _saveImageToDCIM(String fileName, Uint8List fileBytes) async {
+    try {
+      final result = await ImageGallerySaver.saveImage(
```
snehmehta, line 86: what is reason for using ImageGallerySaver, cause adding in android would be placing file on that location. is it ios?

snehmehta (changes requested):

modules/ensemble/lib/action/save_file.dart
mehsaandev: If I just place the file in that location it doesn't register the image in the gallery, using ImageGallerySaver it automatically registers the image to the gallery in the respective platforms I.e. Android and IOS

kmahmood74:
Due to the urgency of this fix, I would suggest having a meeting to go
through the PR. We need these done today as the customer is waiting

mehsaandev:
I've made the required changes. 
- For `ImageGallerySaver` It registers the file to gallery so that's I added this, only saving the file to Pictures path won't register it to gallery
- Removed Document registry, saving the document is sufficient.
- Replaced Dio with HTTP

@snehmehta PTAL

PR #1809, opened 2025-01-03
Added controlDeviceBackNavigation action 

mehsaandev:


mehsaandev:
@snehmehta  

Demo video:

https://github.com/user-attachments/assets/1f35b930-c6d0-4e4e-8f59-787e440313f4

modules/ensemble/lib/action/disable_hardware_navigation.dart
```diff
@@ -0,0 +1,89 @@
...
+          event: EnsembleEvent(initiator, data: {'error': e.toString()}),
+        );
+      }
+      rethrow;
+    }
+  }
+
+  void _disableBackButton(BuildContext context) {
+    print(disable);
+    ModalRoute? currentRoute = ModalRoute.of(context);
```
snehmehta, line 68: since finding route is depending on context, I recommend it to test it with different scenerio like custom widget, global widget. 

fyi: we have some static content through the app which isn't necessarily upto date with all screen navigation.

modules/ensemble/lib/action/disable_hardware_navigation.dart
```diff
@@ -0,0 +1,89 @@
...
+          onFailure!,
+          event: EnsembleEvent(initiator, data: {'error': e.toString()}),
+        );
+      }
+      rethrow;
+    }
+  }
+
+  void _disableBackButton(BuildContext context) {
+    print(disable);
```
snehmehta, line 67: use debugPrint or remove it if not necessary

snehmehta (approved):

modules/ensemble/lib/action/disable_hardware_navigation.dart
mehsaandev: Tested with Custom widget and Globel widget, working fine on both

modules/ensemble/lib/action/disable_hardware_navigation.dart
mehsaandev: Removed print().

@snehmehta PTAL

Issue #1813, opened 2025-01-03
require closeApp action to close the app


mehsaandev:

mehsaandev:
Example Usage:

```
Button
  label: Close App
  onTap:
      closeApp:
```

PR #1814, opened 2025-01-03
added closeApp action

mehsaandev:
https://github.com/EnsembleUI/ensemble/issues/1813
@sharjeelyunus PTAL

### DIFF
--- a/modules/ensemble/lib/action/close_app.dart
+++ b/modules/ensemble/lib/action/close_app.dart
@@ -1,0 +1,27 @@
+import 'package:ensemble/framework/action.dart';
+import 'package:ensemble/framework/scope.dart';
+import 'package:ensemble/util/utils.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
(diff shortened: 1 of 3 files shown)


mehsaandev:
@sharjeelyunus I created a separate file for close_app action, is it fine or should I make this action in `actions.dart` because it doesn't require much implementation?

kmahmood74 (approved):
please remember to add documentation, kitchen sink example and update schema. Also note my comment about iOS release mode

modules/ensemble/lib/action/close_app.dart
```diff
@@ -0,0 +1,24 @@
...
+
+  factory CloseAppAction.from({Invokable? initiator, Map? payload}) {
+    return CloseAppAction(
+      initiator: initiator,
+    );
+  }
+
+  @override
+  Future execute(BuildContext context, ScopeManager scopeManager) {
+    SystemNavigator.pop();
```
kmahmood74, line 21: this won't work for iOS release mode. Please note that in the documentation and kitchen sink example of it

kmahmood74 (approved):

PR #1787, opened 2024-12-23
added moenage stuff

TheNoumanDev:
ticket: https://github.com/EnsembleUI/ensemble/issues/1773

Right now creating script for Build system for all configurations related to updating gradle, manifest and kotlin files needed for moenage. Also for iOS.

Wot is implemented for Moenagage is:
- Push Notifications
- Events
- InApp Notifications

PR #1816, opened 2025-01-05
removed moengage dependency from ensemble module

TheNoumanDev:
was getting error while creating a build where even moengage was not enabled bcz we were using moengage package in ensemble module to get access to some Enums from it, so removed that package.


snehmehta (approved):
I missed to follow up but here it happens most of time when we add package so added a comment in initial PR, this pretty common issue in our system.
https://github.com/EnsembleUI/ensemble/pull/1787#discussion_r1897136407

Also, just to make sure it was minified issue when building just confirm in build system we don't have that issue when enabling moengage.

Issue #1812, opened 2025-01-03
Firebase Auth to be accessible by Firestore calls


amin-nas:
I want to set Firestore rules to limit read/write to authenticated users. Currently, when I sign in using Firebase Auth in an Ensemble app, the subsequent Firestore calls with auth-required rule cannot be made successfully.

PR #1821, opened 2025-01-08
initialise same app for firestore and auth

snehmehta:
Ticket: https://github.com/EnsembleUI/ensemble/issues/1812


modules/ensemble/lib/framework/apiproviders/firestore/firestore_api_provider.dart
```diff
@@ -58,7 +58,7 @@ class FirestoreAPIProvider extends APIProvider with LiveAPIProvider {
     } else {
       // Initialize the new Firebase app with the options
       FirebaseApp app = await Firebase.initializeApp(
-        name: appId,
+        name: 'customFirebase',
```
kmahmood74, line 61: huh?

snehmehta: firebase auth initialisation:

https://github.com/EnsembleUI/ensemble/blob/3b31f77b6791bfe761699a8b6408cc32967553c6/modules/auth/lib/signin/auth_manager.dart#L178-L179

kmahmood74: I understand, my question was why hardcode the name of the app? is it supposed to match something somewhere else? where?

snehmehta: we just need to name firebase app, since we have ensemble firebase app initialised. 
Only the firebase auth and firestore need to match the name of firebase app. 

Either way we'll need to have single initialisation of firebase service to not worry about matching everywhere.

kmahmood74: the reason I did it this way was to allow multiple apps to run in Ensemble Go and the Studio. That use case will fail if we hardcode the app name right?

snehmehta: In that case, we can keep firestore initialisation as it is and update the auth side.

kmahmood74: yes - on the auth side we can just use the appId as well and then the two will be the same right? I am not sure if that'll be an issue or not?

amin-nas:
@TheNoumanDev - please help close this PR. If there are changes required based on the comments, make the change and reply here for @kmahmood74 to review.

sharjeelyunus (approved):

PR #1822, opened 2025-01-08
conditional import on web implementation

snehmehta:
Issue:
```
Error: Dart library 'dart:html' is not available on this platform.
import 'dart:html' as html;
       ^
Context: The unavailable library 'dart:html' is imported through these packages:
```

Fix:
Conditional Import of html package


snehmehta:
Hi @mehsaandev closing this as @kmahmood74 already merge his version fix into main. 
so fix the web error in different PR.

Issue #1696, opened 2024-10-22
Refactor Tooltip Implementation to Utilize Native Properties instead.


TheNoumanDev:
**Current Issue:**
- **Implementation:** Tooltip is currently implemented as a separate widget.
- **Problem:** This approach contradicts Ensemble’s design philosophy, where widgets should not need to be wrapped by another widget solely to provide additional properties.

**Proposed Solution:**
- **Integration:** Introduce native Tooltip properties directly within each Ensemble widget.
- **Properties to Add:**
```
Column:
  styles:
    height: 100
    backgroundColor: red
    tooltip:
      message: "Custom styled tooltip"
      onTriggered: |-
          console.log('Tooltip was triggered');
      styles:
        backgroundColor: "#2196F3"
        triggerMode: longPress
        textStyle:
          fontSize: 16
          color: "#FFFFFF"
          fontWeight: bold
  children:
    - Text:
        text: Styled Tooltip
```
  
CC: @kmahmood74

TheNoumanDev:
Example EDL:
```yaml
Column:
  styles:
    height: 100
    backgroundColor: red
    tooltipMessage: "Custom styled tooltip"
    tooltipStyle:
      backgroundColor: "#2196F3"
      triggerMode: tap
      onTriggered: |-
        console.log('Tooltip was triggered');
      textStyle:
        fontSize: 16
        color: "#FFFFFF"
        fontWeight: bold
  children:
    - Text:
        text: Styled Tooltip
```

PR #1834, opened 2025-01-14
[1696] Refactored Tooltip implementation

TheNoumanDev:
ticket: https://github.com/EnsembleUI/ensemble/issues/1696

Once Approved will add schema and docs.
Example EDL is added in issue.

Docs: https://github.com/EnsembleUI/ensemble_docs/pull/66
Schema: https://github.com/EnsembleUI/ensemble-web-studio/pull/1541
Recording:
https://github.com/user-attachments/assets/ba8eb279-424d-4c7e-9c00-d53a495a0c24


modules/ensemble/lib/widget/tooltip.dart
```diff
@@ -1,163 +0,0 @@
-import 'package:ensemble/framework/event.dart';
-import 'package:ensemble/framework/extensions.dart';
-import 'package:ensemble/framework/theme/theme_manager.dart';
-import 'package:ensemble/widget/helpers/controllers.dart';
-import 'package:flutter/foundation.dart';
-import 'package:flutter/material.dart';
```
sharjeelyunus, line 6: why are we removing this, if aren't supporting this anymore, should we just add deprecated message.
PS: Ensemble  react also has the Tooltip widget and is widely used in Atlas. Best case is we should create utils and support both

CC: @kmahmood74

modules/ensemble/lib/widget/helpers/tooltip_composite.dart
```diff
@@ -0,0 +1,77 @@
...
+    verticalOffset = Utils.optionalDouble(inputs['verticalOffset']);
+    preferBelow = Utils.optionalBool(inputs['preferBelow']);
+    waitDuration = Utils.getDuration(inputs['waitDuration']);
+    showDuration = Utils.getDuration(inputs['showDuration']);
+    triggerMode = TooltipTriggerMode.values.from(inputs['triggerMode']);
+    backgroundColor = Utils.getColor(inputs['backgroundColor']);
+    borderRadius = Utils.getBorderRadius(inputs['borderRadius'])?.getValue();
+    padding = Utils.optionalInsets(inputs['padding']);
+    margin = Utils.optionalInsets(inputs['margin']);
+    onTriggered = inputs['onTriggered'] != null ?
```
sharjeelyunus, line 21: why `onTriggered` is part of tooltip styles?
isn't is a action, shouldn't it be a property, instead of a style?
`onTooltipTriggered`?

sharjeelyunus: actually instead of 3 new properties of tooltip, we should combine them in an map of 1 i.e tooltip

```yaml
tooltip:
   message:
   onTriggered:
   styles:
```

TheNoumanDev: updated

sharjeelyunus (approved):

PR #1836, opened 2025-01-15
added error and complete block for saveFile

mehsaandev:


modules/ensemble/lib/action/saveFile/save_file.dart
```diff
@@ -33,6 +39,12 @@ class SaveToFileSystemAction extends EnsembleAction {
       blobData: payload['blobData'],
       source: payload['source'],
       type: payload['type'],
+      onComplete: payload['onComplete'] != null
+          ? EnsembleAction.from(payload['onComplete'])
+          : null,
+      onError: payload['onError'] != null
+          ? EnsembleAction.from(payload['onComplete'])
+          : null,
```
sharjeelyunus, line 47: why are we sending `onComplete` on `onError`?
and this null check seems weird? we are checking if its not null? and passing null if its null? WHY?

mehsaandev: fixed

sharjeelyunus (approved):

Issue #1815, opened 2025-01-03
Data Rendering issues in `ScrollableTabBar`


sharjeelyunus:
The ScrollableTabBar tends to freeze a lot, especially on Android. On iOS, the freeze only lasts a few seconds (mostly not noticeable), but on Android, it can sometimes freeze for over a minute, especially when there's a lot of data. This problem gets worse when you have more content in the tab bar. The whole point of using the ScrollableTabBar is to handle large amounts of data easily.

REF: https://discord.com/channels/1031982848485359626/1306327142895718401

kmahmood74:
`ScrollableTabBar` is correctly implemented to load a tab only when it becomes visible. Please create an example screen that showcases the issue otherwise it is hard to debug.

kmahmood74:
It is possible that the tabs in a scrollabletabbar are all being rendered at the same time. @sharjeelyunus please provide an example

sharjeelyunus:
@kmahmood74 try here on Android, its actually noticeable on web as well: https://studio.ensembleui.com/app/JV3Jtf6e2yCfiYZD8R5t/screen/AI3Kl4m9vzlO4gx1mbF2

Email: khizer.1@codilitysolutions.com
Password: 123456

kmahmood74:
yes I can see it on the web as well but there isn't a lot of content

sharjeelyunus:
I added the firebase crashlytics, and these are the errors I'm getting

```
Fatal Exception: io.flutter.plugins.firebase.crashlytics.FlutterError: Bad state: No element. Error thrown .
at List.single(dart:core)
at ScrollController.position(scroll_controller.dart:159)
at ScrollController.offset(scroll_controller.dart:165)
at ScrollableTabBarState._updateActiveTab(scrollable_tab_bar.dart:238)
at ScrollableTabBarState.onVisibilityChanged(scrollable_tab_bar.dart:233)
at RenderVisibilityDetectorBase._fireCallback(render_visibility_detector.dart:87)
at RenderVisibilityDetectorBase._scheduleUpdate.<fn>(render_visibility_detector.dart:150)
at RenderVisibilityDetectorBase._processCallbacks(render_visibility_detector.dart:62)
at _TaskEntry.run(binding.dart:99)
at SchedulerBinding.handleEventLoopCallback(binding.dart:535)
at SchedulerBinding._runTasks(binding.dart:513)
at TickerFuture.whenCompleteOrCancel.thunk(ticker.dart:438)
[...]

PR #1842, opened 2025-01-17
fix: bad state no element on scrollabletabbar

hborham:
Noticed an issue when trying out ScrollableTabBar and console showing "bad state" errors.   It can be replicated in the [kitchen sync](https://studio.ensembleui.com/app/e24402cb-75e2-404c-866c-29e6c3dd7992/screen/CDK0lJ96cFPCWy1nMU7O?line=29&searchTerm=scrollableTab)

The ios simulator had a more specific message about `ScrollController not attached to any scroll views` https://vscode.dev/github/EnsembleUI/ensemble/blob/bugfix/scrollable-tab-barlutter/packages/flutter/lib/src/widgets/scroll_controller.dart#L171

It looks like the scroll state is being access to early in some cases and again when it's no longer visible


hborham:
@kmahmood74 I stumbled upon this when trying out this widget.  It might relate to #1815

kmahmood74 (approved):

sharjeelyunus (approved):
LGTM. This should fix the crashing and error issues, but the freezing problem may still be there if the scrollable tabbar contains a large amount of data

snehmehta (approved):

PR #1843, opened 2025-01-21
Updated Artifacts Folder Structure

mehsaandev:
New folder structure for Local ensemble Apps:-

inside `ensemble/apps/yourAppName`, you should have the following structure for your app data:
```
asset
config
internal_script
internal_widget
screen
theme
.manifest.json
```

Currently Working Items:
- Screens
- widgets
- scripts
- themes
- Assets
- Env and Secret

**Please change the following files in order to run this new structure:**
### Update `pubspec.yaml` of starter App:
   Add all your apps folder in the `assets` section at the very bottom. For example if the name of your app is `helloApp` then add the following under `assets`
   ```
    - ensemble/apps/helloApp/
    - ensemble/apps/helloApp/.manifest.json
    - ensemble/apps/helloApp/asset/
    - ensemble/apps/helloApp/screen/
    - ensemble/apps/helloApp/internal_widget/
    - ensemble/apps/helloApp/internal_script/
    - ensemble/apps/helloApp/theme/
    - ensemble/apps/helloApp/config/appConfig.json
    - ensemble/apps/helloApp/secret/secrets.json

   ``` 
   
   CC: @amin-nas


sharjeelyunus:
Why do we still have these internal_script and internal_widget? We've been criticizing the Firebase structure for having these collections, and now we're expanding it to new features?

Shouldn't the folder structure for both Electron and the starter be something like this?

- assets
- fonts
- scripts
- widgets
- screens
- theme
- i8n

CC: @kmahmood74 @evshi

mehsaandev:
> > Why do we still have these internal_script and internal_widget? We've been criticizing the Firebase structure for having these collections, and now we're expanding it to new features?
> > Shouldn't the folder structure for both Electron and the starter be something like this?
> > 
> > * assets
> > * fonts
> > * scripts
> > * widgets
> > * screens
> > * i8n
> > * theme.yaml
> > 
> > CC: @kmahmood74 @evshi
> 
> Just a had a convo with @evshi we will be renaming these to proper structure @mehsaandev please discuss the latest structure with evan for consistency

Sure

modules/ensemble/lib/framework/definition_providers/local_provider.dart
```diff
@@ -66,6 +77,72 @@ class LocalDefinitionProvider extends FileDefinitionProvider {
...
+  Future<Map?> getCombinedAppBundle() async {
+    Map code = {};
+    Map output = {};
+    Map widgets = {};
+
+    try {
+      // Get the manifest content
+      final manifestContent =
+          await rootBundle.loadString(path + '.manifest.json');
+      final Map<String, dynamic> manifestMap = json.decode(manifestContent);
```
sharjeelyunus, line 89: Hi @evshi,

This function is taking a lot amount of time to load and decode. Its whole job is to extract widget/script names, which are needed to fetch the content. But, the manifest file, for apps with around 100 screens and 100 widgets, can go up to 4-5 MB in size and takes 3-5 seconds to load and decode.

Why are we storing the YAML directly in the manifest file. When we remove the content keys/values from it, it works fine. The large content data is really slowing things down.

From the Electron app, could we either:
- Remove the content keys/values from the manifest file, or
- Provide a separate file containing only the widget and scripts names, which we can use here?

evshi: Yeah, we should not include the YAML in the manifest. @anserwaseem we discussed this previously can you address this now since it is causing this performance issue?

amin-nas:
@mehsaandev @sharjeelyunus - is this ready to be merged?

sharjeelyunus:
> @mehsaandev @sharjeelyunus - is this ready to be merged?

this needs a lil bit of cleanup

modules/ensemble/lib/util/utils.dart
```diff
@@ -921,10 +922,11 @@ class Utils {
     return strings[0];
   }
 
-  /// prefix the asset with the root directory (i.e. ensemble/assets/), plus
+  /// prefix the asset with the app root directory (i.e. ensemble/apps/<app-name>/assets/), plus
   /// stripping any unnecessary query params (e.g. anything after the first ?)
   static String getLocalAssetFullPath(String asset) {
-    return 'ensemble/assets/${stripQueryParamsFromAsset(asset)}';
+    String path = EnsembleConfigService.config["definitions"]?['local']?["path"];
+    return '${path}/asset/${stripQueryParamsFromAsset(asset)}';
```
sharjeelyunus, line 929: shouldn't this be assets? instead of asset?

also the assets are stored locally in `ensemble/assets` folder in case of ensemble provider. wouldn't this break the assets for ensemble provider?

modules/ensemble/lib/framework/secrets.dart
```diff
@@ -24,8 +28,16 @@ class SecretsStore with Invokable {
...
       // add local overrides
       try {
-        await dotenv.load();
-        secrets.addAll(dotenv.env);
+        String path =
+            EnsembleConfigService.config["definitions"]?['local']?["path"];
+        final secretsString =
+            await rootBundle.loadString('${path}/config/secrets.json');
+        final Map<String, dynamic> appSecretsMap = json.decode(secretsString);
+        // await dotenv.load();
```
sharjeelyunus, line 36: remove this

sharjeelyunus (commented):
please make sure this is not breaking the `ensemble` or `server` provider. Also update the example `helloApp` structure with the updated structure.

modules/ensemble/lib/util/utils.dart
mehsaandev: yes it should be assets

and for Ensemble provider, No it won't because for ensemble provider source of each assets is cloud-based. we are not storing assets locally in case of ensemble provider, so in case of ensemble provider this function will not be called to create a local path

sharjeelyunus: we are storing local assets in ensemble provider as well, for studio yes we provide the cloud-based urls, but when user creates a build, we store the assets in the old assets folder for ensemble provider

modules/ensemble/lib/framework/secrets.dart
```diff
@@ -24,8 +28,14 @@ class SecretsStore with Invokable {
...
-        await dotenv.load();
-        secrets.addAll(dotenv.env);
+        String path =
+            EnsembleConfigService.config["definitions"]?['local']?["path"];
+        final secretsString =
+            await rootBundle.loadString('${path}/config/secrets.json');
+        final Map<String, dynamic> appSecretsMap = json.decode(secretsString);
+        appSecretsMap["secrets"].forEach((key, value) {
+          secrets![key] = value;
+        });
```
sharjeelyunus, line 38: how this will work for ensemble provider? why is this static to local only?

starter/ensemble/apps/helloApp/scripts/common.yaml
```diff
@@ -0,0 +1,14 @@
...
+function saveName(first,last) {
+  ensemble.storage.helloApp = {name: {first: first, last: last}};
+}
+
+function getName() {
+  if ( ensemble.storage.helloApp != null ) {
+    return ensemble.storage.helloApp.name;
+  }
+  return null;
+}
```
sharjeelyunus, line 14: shouldn't this be a JS file??? why scripts files are `.yaml` files?

mehsaandev:
@sharjeelyunus  Updated scripts extension to js, PTAL

sharjeelyunus (approved):

PR #1847, opened 2025-01-27
[1846] Updated http req to bypass SSL certificate based on env parameter

TheNoumanDev:
- If `ssl_pinning_enabled` is set to `true`: requires SSL certificate in secrets.
- If `ssl_pinning_enabled` is set to `false`: will bypass SSL certificate.
- If `ssl_pinning_enabled` is not set in env variables: would make regular `http.Client` req.


modules/ensemble/lib/framework/apiproviders/http_api_provider.dart
```diff
@@ -223,24 +224,29 @@ class HTTPAPIProvider extends APIProvider {
...
-
-      HttpClient httpClient = HttpClient(context: context);
-      return IOClient(httpClient);
-    } else {
-      return http.Client();
+      return IOClient(HttpClient(context: context));
+    } else if (sslPinningEnabled == false) {
+      // Bypass SSL verification
+      return IOClient(
+          HttpClient()..badCertificateCallback = (cert, host, port) => true);
```
sharjeelyunus, line 244: if ssl pinning is not enabled why are we using IOClient?
shouldn't we just use the `http.Client()`?

TheNoumanDev: The three states serve different purposes:

- true: Strict certificate pinning with provided certificate
- false: Explicitly bypass certificate validation
- null: Default certificate validation behavior using `http.Client()` which will be used if we dont specify `ssl_pinning_enabled` variable.

Using http.Client() for false wouldn't achieve the goal of bypassing certificate requirements - it would still fail on invalid certificates. That's why we use IOClient with badCertificateCallback for the false case.

sharjeelyunus: using `badCertificateCallback = (cert, host, port) => true` can expose api's to security risk, can you make sure if we are handling it properly and if possible add test cases

sharjeelyunus: If we really need this, at least we should use a separate env variable for bypassing, don't mix it with current functionality. SSL pinning should work as it was before. And if user specifically wanna bypass the SSL Pinning they can use `bypass_ssl_pinning` (true/false)

amin-nas:
@TheNoumanDev please document this and add Sharjeel as a reviewer of the doc page.

TheNoumanDev:
Updated with new env variable:
if `bypass_ssl_pinning` is true: we will bypass the SSL certificate check.
@sharjeelyunus PTAL

sharjeelyunus (approved):

modules/ensemble/lib/framework/apiproviders/http_api_provider.dart
kmahmood74: I agree, this has nothing to do with certificate pinning, perhaps you are confusing certificate pinning with SSL. You can have SSL (https) without certificate pinning, that is the most common way anyway. Just call the environment variable `allowUnsecureHttpRequests` or something like that. When true, allow http and not just https. Even the, you are simply using what the user specified as the API url, you are never changing it from https to http yourself.

Issue #1851, opened 2025-01-30
Input label does not appear when toggling visibility


amin-nas:
When an input has the following conditions:

- it is child of a Form widget
- the input has a label
- the input has `visible: false`

and then I switch the visible to true, the input box appears but the label does not.

Test screen:
https://studio.ensembleui.com/app/9YfZUVl3SC82UZ5Mnf2e/screen/D1jXo9ejJb7917M9O0fX

PR #1854, opened 2025-01-30
[1851] fixed the label of textfield in form when its label position to start

TheNoumanDev:
Ticket: https://github.com/EnsembleUI/ensemble/issues/1851

Fixed an issue where form labels weren't updating their visibility state when the associated input field's visibility was toggled. While input fields correctly handled visibility changes, it was because the labels added separately in the form layout weren't being rebuilt again.

Changes made:
- Added `ValueListenableBuilder` in form's `buildGrid` method to rebuild the grid layout when properties change
- Added a `formStateNotifier` in `FormController` to track form state changes
- Implemented property change listener in `TextInputState` to notify form of any controller property updates
- Added mounted check before accessing context to prevent unmounted widget errors

Previously, changing an input field's visibility only affected the input itself, leaving its label in an inconsistent state. Now, the entire form grid rebuilds whenever any property changes, ensuring labels and inputs stay in sync.

Test EDL:
```yaml
View:
  styles:
    useSafeArea: true
    backgroundColor: white

  header:
    titleText: Home
  body:
    Column:
      styles:
        padding: 24
        gap: 32
      children:
        - Button:
            label: Test
            onTap: |
              testInputA.visible = true;
              testInputB.visible = true;
        - Button:
            label: Test
            onTap: |
              testInputA.visible = false;
              testInputB.visible = false;
        - TextInput:
            label: I was invisible
            visible: false
            id: testInputA
        - Spacer:
            styles:
              size: 40
        - Form:
            styles:
              labelPosition: start
            children:
              - TextInput:
                  label: I was invisible
                  visible: false
                  id: testInputB

```

Bug:
Bug image
Fixed:
Fixed image


modules/ensemble/lib/layout/form.dart
```diff
@@ -190,62 +197,72 @@ class FormState extends EWidgetState<EnsembleForm>
...
-      if (widget._controller.gap > 0 && i != formItems.length - 1) {
-        rows.add(flutter.SizedBox(
-            width: widget._controller.gap.toDouble(),
-            height: widget._controller.gap.toDouble()));
-      }
-    }
-
-    return hasAtLeastOneLabel ? Column(children: rows) : buildColumn(formItems);
+    return ValueListenableBuilder(
+        valueListenable: widget._controller.formStateNotifier,
```
kmahmood74, line 201: so the only change here is the ValueListenableBuilder(
        valueListenable: widget._controller.formStateNotifier,

rest is the same right?

TheNoumanDev: Yes, rest is same, its just change bcz of indent.

PR #1875, opened 2025-02-18
flutter 3.29 build issues solved

TheNoumanDev:
Flutter 3.29 comes with deprecation of many plugins

### Workmanager Package Changes:
- Changed from pub.dev package (0.5.2) to direct git reference because:
  1. The pub version uses deprecated Flutter plugin registration system (`Registrar` and `PluginRegistrantCallback`)
  2. Flutter 3.29.0 removed these deprecated APIs in favor of new embedding system
  3. The git main branch has updated code using new Flutter plugin APIs but new version has not been published to pub.dev with these updates yet, for that issue is there (https://github.com/fluttercommunity/flutter_workmanager/issues/588)

### Changes in OTP Pin Field Package:
- Removed `registerWith` method and `Registrar` constructor.
- Added null safety check for `signingInfo.apkContentsSigners` to fix type mismatch errors
- No changes to core feature implementation or business logic

### `image_gallery_saver` Package:
- Original package (2.0.3) not maintained and has Flutter 3.29.0 compatibility issues
- Created temporary fork and branch `flutter_2_29_registrar_fix` with required updates
- Removed deprecated plugin registration system and fixed for new Flutter embedding
- Note: will put a PR for the original package, but that package is no more maintained as it already have a lot of pending PRs.

### DIFF
--- a/modules/ensemble/pubspec.yaml
+++ b/modules/ensemble/pubspec.yaml
@@ -91,9 +91,16 @@ dependencies:
   walletconnect_dart: ^0.0.11
   http_parser: ^4.0.2
   shared_preferences: ^2.1.1
-  workmanager: ^0.5.1
+  workmanager:
+    git:
+      url: https://github.com/fluttercommunity/flutter_workmanager.git
+      ref: main 
+      path: workmanager
   flutter_local_notifications: ^17.2.3
-  image_gallery_saver: ^2.0.3
+  image_gallery_saver:
+    git:
+      url: https://github.com/TheNoumanDev/image_gallery_saver.git
+      ref: flutter_2_29_registrar_fix
   flutter_i18n: ^0.35.1
   pointer_interceptor: ^0.9.3+4
   flutter_secure_storage: ^9.2.2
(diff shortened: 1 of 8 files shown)


modules/ensemble_bluetooth/pubspec.yaml
```diff
@@ -24,7 +24,11 @@ dependencies:
 
   collection: ^1.17.1
   flutter_blue_plus: ^1.33.4
-  workmanager: ^0.5.2
+  workmanager:
```
kmahmood74, line 27: do they have a package? it's dangerous getting the latest from main like this?

modules/ensemble/pubspec.yaml
```diff
@@ -61,7 +61,7 @@ dependencies:
   otp_pin_field:
     git:
       url: https://github.com/EnsembleUI/ensemble.git
-      ref: main
+      ref: flutter_3_29_upgrade_fix
```
sharjeelyunus, line 64: shouldn't this be main?

TheNoumanDev: yesh updated.

Issue #1703, opened 2024-10-28
Support for Firebase signInWithCustomToken


amin-nas:
I want to use a custom auth mechanism that checks username/password and returns a token. Then I want to use this token to authenticate user against Firebase so that user can access Firebase resources such as Firestore. 

Firebase docs:
https://firebase.google.com/docs/auth/flutter/custom-auth

Example use case:

1. User opens Ensemble app and authenticate against a custom backend
2. Backend returns a token
3. App logs user into Firebase using this token
4. User can now access collections that require auth

PR #1891, opened 2025-02-26
Firebase custom token auth

Umair-Manzoor-47:
Github Ticket : https://github.com/EnsembleUI/ensemble/issues/1703

1. Send credentials to backend. 
2. onComplete auth from backend send token to this `signInWithCustomToken` Action.
3. it Signs in Existing User if already exists or Creates a new one if User doesn't exists already.
4. User is now Authenticated and Signed in on firebase.

#### Attributes:
1. **token** :  Custom `jwtToken` that will be sent from backend minted with firebase admin SDK
2. **onComplete** : `Action` on signed in successfully. 
3. **onError** : `Action` on error on sign in.

#### Returns under event.data:

- Authenticated User as `user`
- firebase return token as `idToken`  

**Note**: firebase user `uid` is accessible under `event.data.user.id`

```
    signInWithCustomToken:
      token: {custom jwtToken} 
      onAuthenticated:
        executeCode:
          body: |
            ensemble.storage.userAuthId = event.data.idToken;
            console.log("ensemble.storage.userAuthId");
          onComplete:
            executeActionGroup:
              actions:
                - showToast:
                    message: ${ensemble.storage.userAuthId}


```

### DIFF
--- a/modules/auth/lib/signin/auth_manager.dart
+++ b/modules/auth/lib/signin/auth_manager.dart
@@ -146,7 +146,7 @@ class AuthManager with UserAuthentication {
         print('Sign in with custom firebase auth token failed');
         return null;
       }
-      String? token = await user.getIdToken();
+      String? firebaseToken = await user.getIdToken();
       Future<void> updateCurrentUser(BuildContext context, User newUser) async {
         await StorageManager()
             .writeToSystemStorage(UserAuthentication._idKey, newUser.uid);
@@ -155,7 +155,7 @@ class AuthManager with UserAuthentication {
       }
 
       updateCurrentUser(context, user);
-      return token;
+      return firebaseToken;
     } catch (e) {
       print(e.toString());
       return null;
--- a/modules/auth/lib/signin/auth_manager.dart
+++ b/modules/auth/lib/signin/auth_manager.dart
@@ -132,7 +132,7 @@ class AuthManager with UserAuthentication {
     }
   }
 
-  Future<bool?> signInWithCustomAuth(
+  Future<bool?> signInWithCustomToken(
       BuildContext context,
       {required String idToken}
   ) async {
--- a/modules/auth/lib/signin/auth_manager.dart
+++ b/modules/auth/lib/signin/auth_manager.dart
@@ -132,7 +132,7 @@ class AuthManager with UserAuthentication {
     }
   }
 
-  Future<bool?> signInWithCustomToken(
+  Future<String?> signInWithCustomToken(
       BuildContext context,
       {required String idToken}
   ) async {
@@ -146,6 +146,7 @@ class AuthManager with UserAuthentication {
         print('Sign in with custom firebase auth token failed');
         return null;
       }
+      String? token = await user.getIdToken();
       Future<void> updateCurrentUser(BuildContext context, User newUser) async {
         await StorageManager()
             .writeToSystemStorage(UserAuthentication._idKey, newUser.uid);
@@ -154,7 +155,7 @@ class AuthManager with UserAuthentication {
       }
 
       updateCurrentUser(context, user);
-      return user.isAnonymous;
+      return token;
     } catch (e) {
       print(e.toString());
       return null;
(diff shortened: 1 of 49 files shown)


amin-nas:
@Umair-Manzoor-47 let's rename this to `signInWithCustomToken`
Also, we need to have the id token available in `event.data.user`. Since we will be calling secured cloud functions, we need to pass this token in the header. This token should be available for all methods of firebase auth supported by Ensemble.

Umair-Manzoor-47:
@amin-nas 
For clarification 
`event.data.user` must have `tokenId` sent from backend? 
OR `UUID` that is sent from **firebase** after user is **authenticated** and signed in. 
Now `event.data.user` already has **UUID**.

modules/auth/lib/signin/sign_in_with_custom_token.dart
```diff
@@ -0,0 +1,26 @@
...
+class SignInWithCustomTokenImpl implements SignInWithCustomToken {
+  @override
+  Future<void> signInWithCustomToken(BuildContext context,
+      {required SignInWithCustomTokenAction action}) async {
+    final token = await AuthManager().signInWithCustomToken(context,idToken: action.idToken!);
+    if (token != null) {
+      if (action.onAuthenticated != null) {
+        AuthenticatedUser? currentUser = AuthManager().getCurrentUser();
+        ScreenController().executeAction(context, action.onAuthenticated!,
+            event: EnsembleEvent(null, data: {'user': currentUser, 'token': token}));
```
sharjeelyunus, line 18: please expose `token` as `idToken` to be consistent with other auth actions

modules/auth/lib/signin/auth_manager.dart
```diff
@@ -132,6 +132,36 @@ class AuthManager with UserAuthentication {
...
 
+  Future<String?> signInWithCustomToken(
+      BuildContext context,
+      {required String idToken}
+  ) async {
+    try {
+      customFirebaseApp ??= await _initializeFirebaseSignIn();
+      final _auth = FirebaseAuth.instanceFor(app: customFirebaseApp!);
+
+      UserCredential userCredential = await _auth.signInWithCustomToken(idToken);
```
TheNoumanDev, line 143: in `UserCredential ` we are getting `AuthCredential` and `AdditionalUserInfo` as well along `user` info, so can you please also get them if they are not null and store it in [`data`](https://github.com/EnsembleUI/ensemble/blob/3375794044d3e1f7951d353f6594ce6767ce99ca/modules/ensemble/lib/framework/stub/auth_context_manager.dart#L126) property of `AuthenticatedUser`.

TheNoumanDev (changes requested):

modules/auth/lib/signin/auth_manager.dart
```diff
@@ -132,6 +132,36 @@ class AuthManager with UserAuthentication {
...
+      User? user = userCredential.user;
+      if (user == null) {
+        print('Sign in with custom firebase auth token failed');
+        return null;
+      }
+      String? firebaseToken = await user.getIdToken();
+      Future<void> updateCurrentUser(BuildContext context, User newUser) async {
```
TheNoumanDev, line 150: Before merging, please create a common method for `updateCurrentUser` as the same method is being used in `signInAnonymously`
Plus, if possible, add a line description of some complex code so that it would be easy for someone to understand in the future

TheNoumanDev (approved):

sharjeelyunus:
@Umair-Manzoor-47 can you confirm what type of token user has to provide to use this action? Will it be a identity token? Or a JWT token?

Umair-Manzoor-47:
@sharjeelyunus 
It will be `JWT` token but it must be minted with some `firebase attributes` required for auth. 
as mentioned here: [Firebase Custom JWT token](https://firebase.google.com/docs/auth/admin/create-custom-tokens#java_1)

sharjeelyunus:
then why are we asking user to provide `idToken` in the edl?

TheNoumanDev (approved):

Issue #1867, opened 2025-02-11
Need a property to keep keyboard open and TextInput focused after performing an action


ridsashabbir:
Currently, after a user clicks the `Add Comment` button, the keyboard closes, and the text input loses focus. This behavior should be updated to reflect the commenting experience on platforms like Instagram. The keyboard should remain open, and the text input should stay focused even after the user adds a comment.

Here is an example usage that we need to recreate

https://github.com/user-attachments/assets/60d43208-65c6-4907-9e8b-335f66047985

And that's how it looks in ensemble using workarounds

https://github.com/user-attachments/assets/c4ac0be8-3f20-45c6-abea-07ab1f075354

sharjeelyunus:
I think if we are adding a property then it should be inside `Form`

amin-nas:
> I think if we are adding a property then it should be inside `Form` instead of `TextInput`

Let's do both as you are not forced to use Form widget. 

@ridsashabbir in the example you shared, is it possible to dismiss the keyboard? if so, how?

ridsashabbir:
> > I think if we are adding a property then it should be inside `Form` instead of `TextInput`
> 
> Let's do both as you are not forced to use Form widget.
> 
> [@ridsashabbir](https://github.com/ridsashabbir) in the example you shared, is it possible to dismiss the keyboard? if so, how?

@amin-nas it is possible only if user clicks the back button from his device's hardware, other than this keyboard should remain open

ridsashabbir:
> I think if we are adding a property then it should be inside `Form` instead of `TextInput`

@sharjeelyunus it is already in `Form` and we are performing add comment action `onSubmit` of form

ridsashabbir:
reopening the issue as it is working for adding comment, but showing the weird behavior if you try to delete a comment

CC: @mehsaandev 

https://github.com/user-attachments/assets/a702019d-1bb8-4fba-b960-d9b14f13f5a3

mehsaandev:
> reopening the issue as it is working for adding comment, but showing the weird behavior if you try to delete a comment
> 
> CC: [@mehsaandev](https://github.com/mehsaandev)
> 
>  Record_2025-03-04-16-38-32.mp4

Added Fix:  https://github.com/EnsembleUI/ensemble/pull/1905

PR #1890, opened 2025-02-26
new properties to form and TextInput widget for keyboard dissmiss control

mehsaandev:
Ticket: https://github.com/EnsembleUI/ensemble/issues/1867
Demo Video:
https://github.com/user-attachments/assets/b6977565-88c6-4d38-8ac5-6e15265f526c

CC: @ridsashabbir @sharjeelyunus

PR #1905, opened 2025-03-04
Fix: added `endingWidget` in TextInput widget to avoid keyboard dismiss on performing action on `endingWidget`

mehsaandev:
Ticket Reference: https://github.com/EnsembleUI/ensemble/issues/1867#issuecomment-2698498984
Demo 
https://github.com/user-attachments/assets/e2181776-06f0-4978-9faf-3ae984b6ed61


sharjeelyunus:
hey @mehsaandev, this is not right, the problem is when opening the model, or clicking outside widget, keyboard acts weird, it should actually close the keyboard in that case. The keyboard should remain open when submitting the form, but clicking outside, it should close. shared the you video

mehsaandev:
> hey @mehsaandev, this is not right, the problem is when opening the model, or clicking outside widget, keyboard acts weird, it should actually close the keyboard in that case. The keyboard should remain open when submitting the form, but clicking outside, it should close. shared the you video

I am updating the TextField focus as we discussed in the meeting

mehsaandev:
Added `endingWidget` property in TextInput widget as we discussed.
I didn't remove the `dismissbleKeyboard` property from TextInput in case we may need it and `dismissibleKeyboard` for Form widget is also there because we need that.
@sharjeelyunus

sharjeelyunus:
@mehsaandev is this solving the case we discussed last time?

```
the problem is when opening the model, or clicking outside widget, keyboard acts weird, it should actually close the keyboard in that case. The keyboard should remain open when submitting the form, but clicking outside, it should close. shared the you video
```

this PR still has changes in dialog, do we still need that?

PS: even with the ending widget, it should still only keep the keyboard open if dismissbleKeyboard is false.

please try to understand the problem, based on that provide a solution, please share the example by recreation the solution for the problem

mehsaandev:
> @mehsaandev is this solving the case we discussed last time?
> 
> ```
> the problem is when opening the model, or clicking outside widget, keyboard acts weird, it should actually close the keyboard in that case. The keyboard should remain open when submitting the form, but clicking outside, it should close. shared the you video
> ```
> 
> this PR still has changes in dialog, do we still need that?
> 
> PS: even with the ending widget, it should still only keep the keyboard open if dismissbleKeyboard is false.
> 
> please try to understand the problem, based on that provide a solution, please share the example by recreation the solution for the problem

Yes this solution is working, since `endingWidget` is a part of TextInput so by clicking on endingWidget keyboard does not dismiss, Hence we don't need `dismissibleKeyboard` property at all neither in TextInput nor in Form. So Now I am removing `dismissibleKeyboard` properties as we discussed on call.

sharjeelyunus:
@mehsaandev as we have reverted dismissible keyboard, is this change in button from dismissible keyboard PR, it this still necessary?

https://github.com/EnsembleUI/ensemble/pull/1890/files#diff-8de04ac6337e8273d16a5639c72762a63662e922b8210d0cfd21f313e2417e20R282-R286

mehsaandev:
Added `endingWidget` in TextInput as the solution to this problem. 
The problem is that when a user writes any message and sends it using the external button TextInput looses it's focus and keyboard dismisses but we want to keep the keyboard open when we send the message. 
By adding `endingWidget` we can achieve this as `endingWidget` is a part of TextInput and by tapping endingWidget keyboard will not loose focus to TextInput hence it will not dismiss the keyboard. and if we click anywhere else on the screen, keyboard will dismiss which is the expected behavior.

The Following Video is using icon as `endingWidget` to send message:
https://github.com/user-attachments/assets/0fc53e62-9f98-4e8d-a48b-aa5568cd8a09

CC: @sharjeelyunus @ridsashabbir

mehsaandev:
> @mehsaandev as we have reverted dismissible keyboard, is this change in button from dismissible keyboard PR, it this still necessary?
> 
> https://github.com/EnsembleUI/ensemble/pull/1890/files#diff-8de04ac6337e8273d16a5639c72762a63662e922b8210d0cfd21f313e2417e20R282-R286

No we don't need that. I've reverted it

sharjeelyunus (approved):

PR #1909, opened 2025-03-07
Ensuring local asset availability before using firebase asset url

mehsaandev:
@sharjeelyunus PTAL

Demo Video:
- If local assets are available on ensemble provider, it will fetch assets from local asset file

https://github.com/user-attachments/assets/f8480685-fbde-4a9e-a7aa-5966b1820791

- If local assets are not available on ensemble provider, it will use firebase asset URL source

https://github.com/user-attachments/assets/9de4af1d-552b-468f-833d-e6fc97c2906b


TheNoumanDev (approved):
LGTM.
Also, please format the code where necessary, like the `getAssetName` function, so the next person doesn't have to check it every time to avoid messing up the git history.

sharjeelyunus (approved):

Issue #1913, opened 2025-03-11
Camera loading State


Umair-Manzoor-47:
#### Issue
When camera is processing image or cropping according to `overlayWidget`. There's **delay** and no indication for it. 
It seams that application is stuck but it isn't. so a `circularLoadingBar` is needed as camera is loading image. 
#### Preview:
https://github.com/user-attachments/assets/b3ac0fd4-e236-4fd1-8612-07e0e123e34a

PR #1918, opened 2025-03-12
Loading State for camera overlay capture

Umair-Manzoor-47:
Github Ticket: https://github.com/EnsembleUI/ensemble/issues/1913
### Issue
While processing image to crop for overlay Widget `Camera` must goes into `loading state`. 
So application doesn't seems like stuck.

### EDL
```
              openCamera:
                options:
                  captureOverlay: // bool: if image should be cropped according to overlay
                loadingWidget:  // Widget: Custom widget to show on loading Screens
                overlayWidget:  // Widget: Overlay widget to which image will be cropped 

```
### Demo
https://github.com/user-attachments/assets/a748cce3-6d8e-46ac-a4ab-e07b5e5aaed8


modules/camera/lib/camera.dart
```diff
@@ -510,6 +511,17 @@ class CameraState extends EWidgetState<Camera> with WidgetsBindingObserver {
...
             ),
           ),
+        if (isCropping)
+          Container(
+            color: Colors.black.withOpacity(0.7), // Black background
+            width: MediaQuery.of(context).size.width, // Full screen width
+            height: MediaQuery.of(context).size.height, // Full screen height
+            child: Center(
+              child: CircularProgressIndicator(
+                color: Colors.white, // Optional: Change the color of the progress indicator
```
sharjeelyunus, line 521: shouldn't we allow user to build his own loading indicator? if loading widget is not defined we can use this as default, but user should have the option to define his own widget for customization

modules/camera/lib/camera.dart
```diff
@@ -1027,10 +1039,17 @@ class CameraState extends EWidgetState<Camera> with WidgetsBindingObserver {
...
+          isCropping = true;
+        });
         try {
           file = await takeOverlayCapture();
         } on Exception catch (e) {
           print(e);
+        } finally {
+          setState(() {
+            isCropping = false;
+          });
```
sharjeelyunus, line 1052: based on the video in PR, this is taking too much time, is there a way to optimize this? maybe doing a something in background, please checkout why its taking that much time, we can push this for now, but there should be way to optimize take

Umair-Manzoor-47: Actually there's image encoding decoding going on there and also file reading too

modules/camera/lib/camera.dart
Umair-Manzoor-47: Yeah we can have custom widget.

TheNoumanDev:
Please update the Docs and schema, add links in the description, and add an new example EDL to the issue.

sharjeelyunus (approved):

PR #1934, opened 2025-03-25
Semantics Map for Screen Readers

Umair-Manzoor-47:
### Semantics Map

This PR introduces a **semantics mapping** for the `Widget`, ensuring better accessibility support.  

### Added Semantics:
**Attributes**
- `label`: Provides a descriptive name for the `widget`.
- `hint`: Helps screen reader users understand the widget’s function.
- `focusable`: Ensures the `widget` is focusable, improving accessibility and keyboard navigation (`default: true`).
 
EDL:  
### Web:
```yaml
  button:
    semantics: 
      label: Icon Button 
      hint: It navigates to nowhere. 
      focusable: true
```

https://github.com/user-attachments/assets/4e451953-c4e2-4b4b-a719-fa75a12a8e9d

### Android:

https://github.com/user-attachments/assets/4e48f6fa-bcf3-42c3-b0e1-0b6e5df8feb5


modules/ensemble/lib/widget/helpers/controllers.dart
```diff
@@ -580,3 +584,21 @@ class EnsembleBoxController extends EnsembleWidgetController
       borderRadius != null ||
       boxShadow != null;
 }
+
+class EnsembleSemantics {
```
TheNoumanDev, line 588: can u please refactor it in similar way we implemented `BoxShadowComposite`.
Right now, if we update some properties, it won't trigger bindings, e.g if we do `widgetID.semantics.focusable = true` in some code block, it wont be updating the value to it.

modules/ensemble/lib/widget/helpers/input_wrapper.dart
```diff
@@ -78,12 +78,7 @@ class InputWrapper extends StatelessWidget {
             ),
           ),
         // semantics for whatever text input comes through
-        MergeSemantics(
-          child: Semantics(
-              label: controller.label,
-              child: widget,
-            ),
-        ),
+        widget,
```
TheNoumanDev, line 81: why we are removing `MergeSemantics`, this would get rid of that default behavior created for input widgets to have there `label` as part of semantics, when explicit `semantics` are not declared.

TheNoumanDev (commented):
have you tested it on native devices as well? how semantics were behaving before by default configurations and with `semantic` map, please also mention those findings, Thanks

modules/ensemble/lib/widget/helpers/controllers.dart
Umair-Manzoor-47: It is just a `property` class it doesn't have any part in `rendering` so no need for binding.

modules/ensemble/lib/framework/ensemble_widget.dart
```diff
@@ -143,6 +143,16 @@ abstract class EnsembleWidgetState<W extends EnsembleWidget> extends State<W> {
         );
       }
 
+      if(widgetController.semantics != null && widgetController.semantics!.focusable){
```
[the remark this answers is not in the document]
Umair-Manzoor-47: `focusable` has default value `true`. I should remove it from check.

modules/ensemble/lib/framework/ensemble_widget.dart
```diff
@@ -143,6 +143,17 @@ abstract class EnsembleWidgetState<W extends EnsembleWidget> extends State<W> {
         );
       }
 
+      if(widgetController.semantics != null){
+        rtn =  FocusableActionDetector(
```
TheNoumanDev, line 147: wot happens if developer set `widgetController.semantics.focusable = false`?

Umair-Manzoor-47: it skips that widget from readers

TheNoumanDev (approved):

PR #1942, opened 2025-03-28
feat: add JS code validation

sharjeelyunus:
Validates the JavaScript with context-aware rules.

This validator performs static analysis on the provided [Node] or the entire program to catch potential runtime issues before execution. Specifically, it checks for:

1. **Syntax errors** – Any malformed JavaScript structures.
2. **Undefined variables or objects** – Ensures all identifiers used are declared in the current context.
3. **Invalid property or method access** – Verifies that accessed properties/methods exist on the target object.
4. **Invalid method calls** – Ensures functions being called exist and are callable.
5. **Incorrect object structure** – Validates property names of object literals passed to methods or APIs.
6. **Typo suggestions** – Uses Levenshtein distance to suggest corrections for potentially misspelled properties or methods.

### Example
**Code**

```js
//@code
ensemble.storage.set('test', 'test')
if (ensemble.storage.get('test') == 'test') {
  ensemble.showToas({message: "hello"})
}
```

**Error**

```
Exception occurred while running JavaScript code: Line: 4 Message: Property "showToas" does not exist on object "ensemble". Detailed Error: Code: ensemble.showToas
Available properties: storage, user, formatter, utils, device, version
Available methods: callExternalMethod, callNativeMethod, share, rateApp, copyToClipboard, getDeviceToken, getPhoneContacts, getPhoneContactPhoto, showBottomModal, dismissBottomModal, showBottomSheet, dismissBottomSheet, showDialog, navigateViewGroup, showToast, setLocale, clearLocale, openAppSettings, deviceSecurity, requestNotificationAccess, showLocalNotification, dismissDialog, closeAllDialogs, executeActionGroup, takeScreenshot, saveFile, controlDeviceBackNavigation, closeApp, navigateScreen, navigateModalScreen, invokeAPI, openUrl, stopTimer, openCamera, navigateBack, startTimer, uploadFiles, invokeHaptic, playAudio, pauseAudio, stopAudio, resumeAudio, seekAudio, debug, initNotification, updateSystemAuthorizationToken, saveKeychain, clearKeychain, connectSocket, disconnectSocket, messageSocket, dispatchEvent, executeConditionalAction, logEvent, authenticateByBiometric. Recovery: Check if you have used the correct property name (case sensitive)
Did you mean one of these? showToast .
````

Related PR: https://github.com/EnsembleUI/ensemble_live/pull/88

### CODE REVIEW

File: modules/ensemble/lib/framework/devmode.dart [hunk 1]
```diff
@@ -53,4 +55,18 @@ class DevMode {
 
     return context;
   }
+
+  static Map<String, dynamic> validateJsCode(String jsCode, DataContext ctx) {
```
kmahmood74, line 59: where is the ctx coming from? If you are re-using the context, calling evaluate will actually run the code which will change values on the screen.

sharjeelyunus: we are using the `_pageDataContext`
flow is, we get the code from LLM, pass the code to `validateJS` function in `ensemble_live`, this `validateJS` function calls the `validateJsCode` function in runtime

Ensemble Live PR: https://github.com/EnsembleUI/ensemble_live/pull/88

sharjeelyunus:
Yes, you're right @kmahmood74, sorry about this. I've fixed the issue here: [PR Link](https://github.com/EnsembleUI/ensemble_live/pull/88). Now, we're creating a new data context, so when if we run the code validation, it won't cause any changes to the app.

For clarity, here's the difference:

**Before (using pageDataContext):**
When we execute the code, notice that as soon as I make any change in the code, the validator runs. Even though I haven't saved or executed the code, the button label changes unexpectedly.

[Video before fix](https://github.com/user-attachments/assets/54b7093d-dac4-42e7-a856-34b6dc2cfed8)

**After (using sandbox context):**
Now, when the validator runs, it doesn't affect anything on the screen. The button label only changes once the code is saved and executed during the screen load.

[Video after fix](https://github.com/user-attachments/assets/68a6342c-bc3b-49c3-a5e8-7811e5cc49f1)

sharjeelyunus:
sorry my bad, this is still not right, its causing some other issues with the validation, looking for fix

kmahmood74 (changes requested):
This is not the right way to implement validations. I suggest using a clean visitor pattern. Look at the Bindings and JSInterpreter - they are both RecursiveVisitors. Implement a Validator that extends RecursiveVisitor and run the code through it. It is how validations should be implemented. Give chatgpt the existing code and ask it to implement a Validator as a RecursiveVisitor.

kmahmood74 (approved):

PR #1949, opened 2025-04-04
Making rendering of the screen optional in ensembleApp

TheNoumanDev:
A solution to the duplicate GlobalKey error when using 2 EnsembleApp in a native flutter app.
Discord discussion: https://discord.com/channels/1031982848485359626/1351798223781625926/1351798223781625926

The approach we are now using is:
- making the rendering of the screen in `EnsembleApp` optional by passing a child widget from the host app, so that we do all the initialization for ensemble and return that widget back to the host app, so that it works as it is.
- And then just use `Screen` widget at multiple places as many times we want for specific Ensemble screens without the duplicate globalKey error.

Below is a recording with multiple EnsembleApp.
- From home tab going to media details screen that shows match stats in ensemble `Screen`
- From sports tab, we are opening Ensemble `Screen` and then going to the media screen as well that has another ensemble `Screen`.

Recording: https://drive.google.com/file/d/1WQLx8G9EYJyDA-CMvaIjJx4dLPj0uQcH/view?usp=sharing

### DIFF
--- a/modules/ensemble/lib/ensemble_app.dart
+++ b/modules/ensemble/lib/ensemble_app.dart
@@ -379,6 +379,7 @@ class EnsembleAppState extends State<EnsembleApp> with WidgetsBindingObserver {
         appProvider: AppProvider(definitionProvider: config.definitionProvider),
         screenPayload: widget.screenPayload,
         apiProviders: APIProviders.clone(config.apiProviders ?? {}),
+        placeholderBackgroundColor: widget.placeholderBackgroundColor,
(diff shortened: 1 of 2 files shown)


kmahmood74:
@vusters can you please review, this has been pending and urgent

vusters:
Chatted wit Nouman. Ship it

vusters (approved):

kmahmood74 (approved):
good for Vu, good for me

PR #1961, opened 2025-04-16
refactor: replace root_jailbreak_sniffer with safe_device for enhance…

sharjeelyunus:
…d device security checks

- Refactored DeviceSecurity to utilize safe_device for checking if the device is rooted, debugged, or an emulator.
- Added package name and signature validation to the device security checks.


TheNoumanDev (approved):
LGTM

PR #1975, opened 2025-05-05
fix late Initialization

Umair-Manzoor-47:
`TaggerController` in TagInput wasn't being initialized. 
Fixed it.

### DIFF
--- a/modules/ensemble/lib/widget/input/taginput/taginput.dart
+++ b/modules/ensemble/lib/widget/input/taginput/taginput.dart
@@ -247,38 +247,34 @@ class TagInputController extends BaseInputController {
     _initialTagApplied = false;
     _applyingInitialTag = true;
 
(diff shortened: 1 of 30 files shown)


modules/ensemble/lib/widget/input/taginput/taginput.dart
```diff
@@ -423,6 +421,7 @@ class TagInputState extends FormFieldWidgetState<BaseTextInput>
   void didUpdateWidget(covariant BaseTextInput oldWidget) {
     super.didUpdateWidget(oldWidget);
     widget.controller.inputFieldAction = this;
+    widget._controller._taggerController = oldWidget._controller._taggerController;
```
TheNoumanDev, line 424: can you please add the comment above it, why we reassign the controller, so it is easier to debug for the next time.

TheNoumanDev (approved):

PR #1980, opened 2025-05-06
Range slider

M-Taha-Dev:
**Overview:**

This PR adds range slider functionality to the existing EnsembleSlider widget, allowing users to select both a lower and upper value within a specified range. This implementation preserves all existing functionality while adding new capabilities with minimal API changes.

**Features**

- Toggle between single-value and range-slider modes using the enableRange property
- Support for setting start and end values individually or via a single initialRangeValues property
- Compatibility with existing styling composites (track, thumb, tick marks, overlay, value indicators)
- Uses the existing onChange event handler for both slider types for API consistency

Sample EDL & Emulator Output:
image
image

### DIFF
--- a/modules/ensemble/lib/widget/input/slider/slider_state.dart
+++ b/modules/ensemble/lib/widget/input/slider/slider_state.dart
@@ -18,8 +18,6 @@ class SliderState extends FormFieldWidgetState<EnsembleSlider> {
       widget: FormField<double>(
         key: validatorKey,
         validator: (value) {
-          print('slider required');
-          print(widget.controller.required);
           if (widget.controller.required) {
             if (widget.controller.enableRange) {
               if (widget.controller.startValue == widget.controller.minValue &&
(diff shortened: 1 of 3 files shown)


modules/ensemble/lib/widget/input/slider/slider_state.dart
```diff
@@ -121,4 +128,31 @@ class SliderState extends FormFieldWidgetState<EnsembleSlider> {
...
+          ? (RangeValues values) {
+              setState(() {
+                widget.controller.startValue = values.start;
+                widget.controller.endValue = values.end;
+              });
+              if (widget.controller.onChange != null) {
+                ScreenController().executeAction(
+                  context,
+                  widget.controller.onChange!,
+                  event: EnsembleEvent(widget),
```
TheNoumanDev, line 151: we are not returning the current value in the `onChange` event, `startValue `and `endValue` should be returned in the payload of the event, same for the simple slider.

modules/ensemble/lib/widget/input/slider/slider_state.dart
```diff
@@ -83,7 +89,8 @@ class SliderState extends FormFieldWidgetState<EnsembleSlider> {
 
           return SliderTheme(
             data: themeData,
-            child: Slider(
+            child: widget.controller.enableRange ? _buildRangeSlider(context, decimalPlaces):
```
TheNoumanDev, line 92: can you please convert the simple `Slider` to function as well, to keep the consistancy.

TheNoumanDev (changes requested):

PR #1981, opened 2025-05-07
Added Textstyle property for date widget

M-Taha-Dev:
This PR introduces a new textStyle property to style the selected date value in the date widget. To maintain backwards compatibility, the implementation includes logic to fall back to using labelStyle for styling date values when textStyle is not specified. This ensures existing applications continue to work as expected without modifications.

Sample EDL and output
image
image

### DIFF
--- a/modules/ensemble/lib/widget/input/form_date.dart
+++ b/modules/ensemble/lib/widget/input/form_date.dart
@@ -85,77 +85,78 @@ class DateState extends FormFieldWidgetState<Date> {
           .format(widget._controller.value!)
       : '';
 
-@override
-Widget buildWidget(BuildContext context) {
-  // Create a local TextStyle variable that we can modify
+  @override
+  Widget buildWidget(BuildContext context) {
+      // Create a local TextStyle variable that we can modify
   TextStyle dateTextStyle = formFieldTextStyle;
   
   // If textStyle is provided in DateController, apply it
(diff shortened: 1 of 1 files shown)


modules/ensemble/lib/widget/input/form_date.dart
```diff
@@ -123,14 +125,14 @@ class DateState extends FormFieldWidgetState<Date> {
...
-                              textStyle: formFieldTextStyle,
+                              textStyle: widget._controller.textStyle != null ? dateTextStyle : formFieldTextStyle,
                               enabled: widget._controller.enabled ?? true,
                               onCleared: () {
                                 setState(() {
                                   widget._controller.value = null;
                                 });
                               })
-                          : Text(selectedValue, style: formFieldTextStyle)));
+                          : Text(selectedValue, style: widget._controller.textStyle != null ? dateTextStyle : formFieldTextStyle,)));
```
TheNoumanDev, line 135: why are we introducing a new variable `dateTextStyle`, please update the `formFieldTextStyle` with the new properties provided using the `copyWith` method like so that the old styling properties stay preserved adding new ones
```
formFieldTextStyle.copyWith(
  # here update properties we get in `textStyle`
)

TheNoumanDev (changes requested):

TheNoumanDev (approved):

Issue #1982, opened 2025-05-08
Animation Stops on Header When Bottom Sheet is Dismissed


M-Taha-Dev:
The header has an animation that switches between two titles when scrolling. This works fine at first, but if you open and close the bottom sheet, both titles show up together and the transition animation stops working.
Please find the attached video.

https://github.com/user-attachments/assets/103e14f0-6134-4196-b0f4-3cd5486766a0

PR #1986, opened 2025-05-13
Fix Animation stuck

Umair-Manzoor-47:
Fixes: https://github.com/EnsembleUI/ensemble/issues/1982
Added `didUpdateWidget` to sustain `scrollController` state:

https://github.com/user-attachments/assets/b85c0d2a-4476-47d4-bfde-c668520e408d

### DIFF
--- a/modules/ensemble/lib/framework/view/page.dart
+++ b/modules/ensemble/lib/framework/view/page.dart
@@ -919,7 +919,7 @@ class AnimatedAppBar extends StatefulWidget {
 
 class _AnimatedAppBarState extends State<AnimatedAppBar> with WidgetsBindingObserver{
   bool isCollapsed = false;
-  bool hasInitialized = false;
+  
   @override
   void initState() {
     super.initState();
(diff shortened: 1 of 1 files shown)


modules/ensemble/lib/framework/view/page.dart
```diff
@@ -943,6 +943,33 @@ class _AnimatedAppBarState extends State<AnimatedAppBar> {
     }
   }
 
+  @override
+  void didChangeMetrics() {
+    super.didChangeMetrics();
+
+    if (hasInitialized) {
+      _updateCollapseState();
+      hasInitialized = true;
```
TheNoumanDev, line 952: `hasInitialized ` flag is set to true only when it is already true, it makes no sense, and it will never come into the if block as its initilaze value is set to false at line number 922.

modules/ensemble/lib/framework/view/page.dart
```diff
@@ -943,6 +943,33 @@ class _AnimatedAppBarState extends State<AnimatedAppBar> {
     }
   }
 
+  @override
+  void didChangeMetrics() {
```
TheNoumanDev, line 947: this is called when device metrics change (orientation, keyboard, etc.), not when widgets are rebuilt, idk if that is required or not, please confirm.

TheNoumanDev (changes requested):

Issue #1984, opened 2025-05-12
Ability to store data locally in secure storage


amin-nas:
I want to store data locally using the platform’s secure storage facilities for security related data:

- for iOS, use the Keychain
- for Android, use the KeyStore

Use cases include storing password and jwt token.

sharjeelyunus:
we already have keychain manager here with 2 actions `saveToKeychain` & `clearKeychain` for secure storage: https://github.com/EnsembleUI/ensemble/blob/main/modules/ensemble/lib/framework/keychain_manager.dart
but not sure if they actually solves the use case

amin-nas:
thanks @sharjeelyunus - perhaps we can have a generic action that applies to iOS and Android:

- setSecureStorage - I should be able to pass an object
- clearSecureStorage

I would also need to reference those items. E.g. if I am storing a key of `token`, I should be able to reference it as `ensemble.secureStorage.token`

PR #1987, opened 2025-05-14
[Feat] Add secureStorage actions

TheNoumanDev:
Ticket: https://github.com/EnsembleUI/ensemble/issues/1984
docs: https://github.com/EnsembleUI/ensemble_docs/pull/82
Schema: https://github.com/EnsembleUI/ensemble-web-studio/pull/1639

When we were using `ensemble.storage.variable`, it was storing user information in plain text in the device, which is not ideal way to do it as it may caontain sensitive information.

So we can use `keyChain` & `keyStore` for secure storage, for which we already have action `saveKeyChain` to store the values, it can be used in the YAML and JS, both ways.

But we didn't had any way to read those values back, so i introduced a new action `readKeyChain`. It is only available in YAML, as this calls a async function whose return type is Future and we use callbacks to handle the result. Our JS is sync and we cannot use async/await in JS. 
```yaml
readKeyChain:
  key: adminPassword
  onComplete:
    invokeAPI:
      name: apiName
      inputs:
        adminPass: ${event.data.value} # it would be the value of the above key.
```

But for JS compatibility, i added 2 new actions `SetSecureStorage ` and `GetSecureStorage `. It encrypts the text using the technique `AES (Advanced Encryption Standard) in CBC mode` which gets the encryptionKey form the secrets. Wot it does is
- its gets the Plain text from action, encrypt it and store it in base64 in device storage.
- When we need it, it decrypt the encrypted_text and then pass it back to the UI to show the plain text.

Usage in Yaml
- With simple string as value
```yaml
setSecureStorage:
  key: adminPassword
  value: 'simpleString'     // stored as something asldakljd790asd9#asd@==
```
- with object as value
```yaml
setSecureStorage:
  key: adminPassword
  value:
     userName: 'username123'
     password: 'password123'
```
- Getting value back
```yaml
getSecureStorage:
  key: adminPassword
  onComplete:
    invokeAPI:
      name: apiName
      inputs:
        adminPass: ${event.data.value} # it would be the value of the above key.
```

Usage in JS:
```JS
// To save the value
ensemble.setSecureStorage({
  'key': 'name',
  'value': 'EnsembleUI'
});

// To get the value
ensemble.getSecureStorage({'key': 'name'});
// can also get value as
ensemble.getSecureStorage('name');
```

### DIFF
--- a/modules/ensemble/lib/framework/data_context.dart
+++ b/modules/ensemble/lib/framework/data_context.dart
@@ -19,6 +19,7 @@ import 'package:ensemble/framework/apiproviders/http_api_provider.dart';
 import 'package:ensemble/framework/config.dart';
 import 'package:ensemble/framework/data_utils.dart';
 import 'package:ensemble/framework/device.dart';
+import 'package:ensemble/framework/encrypted_storage_manager.dart';
 import 'package:ensemble/framework/error_handling.dart';
 import 'package:ensemble/framework/keychain_manager.dart';
 import 'package:ensemble/framework/app_info.dart';
@@ -91,12 +92,11 @@ class DataContext implements Context {
     _addLegacyDataContext();
   }
 
-
-    // device is a common name. If user already uses that, don't override it
-    // This is now in ensemble.device.*
+  // device is a common name. If user already uses that, don't override it
+  // This is now in ensemble.device.*
   void _addLegacyDataContext() {
     if (_contextMap['device'] == null) {
-        _contextMap['device'] = Device();
+      _contextMap['device'] = Device();
     }
   }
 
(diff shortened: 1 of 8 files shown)


amin-nas:
@sharjeelyunus please review

modules/ensemble/lib/framework/data_context.dart
```diff
@@ -551,6 +551,12 @@ class NativeInvokable extends ActionInvokable {
         final evalMessage = scope?.dataContext.eval(message);
         messageSocket(socketName, evalMessage);
       },
+      ActionType.setSecureStorage.name: (dynamic inputs) =>
+          EncryptedStorageManager.setSecureStorage(inputs),
+      ActionType.getSecureStorage.name: (dynamic inputs) =>
+          EncryptedStorageManager.getSecureStorage(inputs),
+      ActionType.clearSecureStorage.name: (dynamic inputs) =>
+          EncryptedStorageManager.clearSecureStorage(inputs),
```
sharjeelyunus, line 559: please don't add new actions here, this is the old way to add the data context, to make the actions invokables we now add the actions in `action_invokable.dart`

sharjeelyunus (approved):
LGTM

modules/ensemble/lib/framework/encrypted_storage_manager.dart
```diff
@@ -0,0 +1,163 @@
...
+
+    if (val == null) {
+      StorageManager().remove(PREFIX + key);
+      return;
+    }
+
+    try {
+      final encryptedValue = _encryptValue(val);
+      print('original value: $val');
+      print('Encrypted value: $encryptedValue');
```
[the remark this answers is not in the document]
TheNoumanDev: done

modules/ensemble/lib/framework/data_context.dart
TheNoumanDev: moved `setSecureStorage` and `clearSecureStorage`, but had to keep `getSecureStorage` storage here as we need synchronous value to be return, but `action_invokable.dart` is running execute function which is configured to return future value.

PR #2000, opened 2025-05-20
Overlay fixes

Umair-Manzoor-47:
Fixed overlay hidden when no data.
Provided properties: 
`overlayWidth` [double]
`overlayPadding` [double]

```yaml
TagInput:
   styles:
        overlayWidth: 460
        overlayPadding: 0 10 0 10
```

https://github.com/user-attachments/assets/cccb2b48-f314-4ed8-8d6d-eb432d1db7d3


modules/ensemble/lib/widget/input/taginput/taginput.dart
```diff
@@ -602,6 +611,8 @@ class TagInputState extends FormFieldWidgetState<BaseTextInput>
               ? TriggerStrategy.deferred
               : TriggerStrategy.eager,
           overlayHeight: _calculatedOverlayHeight,
+          overlayWidth: widget._controller.overlayWidth ?? 380,
```
sharjeelyunus, line 614: shouldn't default value be the device widget instead of a random constant?

PR #2003, opened 2025-05-22
Fix text pasting on selectable

Umair-Manzoor-47:
on Web we could still copy and paste text in `inputfield` even when `selectable` is set to `false`.
This PR addresses that issue.


modules/ensemble/lib/widget/input/form_textfield.dart
```diff
@@ -457,11 +457,24 @@ class TextInputState extends FormFieldWidgetState<BaseTextInput>
         onTapOutside: (_) => removeOverlayAndUnfocus(),
         onFieldSubmitted: (value) => widget.controller.submitForm(context),
         onChanged: (String txt) {
+          // If text suddenly increased by more than one character,
+          // it could indicate a paste operation
+          if (txt != previousText &&
```
kmahmood74, line 462: This seems like a major hack. You couldn't find a proper solution? I thought it was not working only in mobile web right, this statement will affect every textinput on every platform?

also what about copy? How do you not allow copy? did you ignore that entirely?

@TheNoumanDev will this break the accessibility when running with a screen reader that is entering values?

kmahmood74: I know it only applies to selectable: false but those are everywhere. Please answer all my questions as follows - 
1. on mobile web, how do you do keyboard paste? I mean there is no Ctrl + v on mobile web right, isn't context menu the only option?
2. on native mobile and desktop web, setting `selectable: false` should disable all copy/paste operations. Have you tried it?
3. what about copy operation? would the user still be able to copy on mobile web even after your change?

Umair-Manzoor-47: 1. Video attached. (contextMenuBuilder is kept null here).
2. on native applications I can't interact with text. Also keeping `contextMenuBuilder = null` works fine.
3. Yeah, user can copy even on my change on mobile web. I couldn't avoid copy. 
I build and deployed web app k on [vercel](https://textfield-kappa.vercel.app/) with `contextMenuBuilder = null` but still able to bring up contextMenu `onLongPress`. 

https://github.com/user-attachments/assets/cc34c49c-2326-4b6d-b876-0bacc23ba92d

PR #2005, opened 2025-05-22
Firebase functions & App check

M-Taha-Dev:
## Add Firebase Functions API provider with App Check support

**Feature:** 
Added support for calling Firebase Cloud Functions through EDL & integrated App Check.

**Key Changes:**
- Created `FirebaseFunctionsAPIProvider` for callable functions
- Integrated Firebase App Check with debug token management
- Added YAML support for function calls with data parameters

**Usage:**
```yaml
  createCustomToken:
    type: firebaseFunction
    name: generateCustomToken
    useAppcheck: false
    data: 
      email: testing123@example.com
      uid: 123456789
    onResponse:
      executeCode:
        body: |-
          console.log('printing body');
          console.log(response.body);
          apiResponse2.text = response.body.token
    onError:
      executeCode:
        body: |-
          console.log('getting error')
          console.log(response)
```


starter/lib/generated/ensemble_modules.dart
```diff
@@ -202,5 +209,14 @@ class EnsembleModules {
     } else {
       GetIt.I.registerSingleton<BluetoothManager>(BluetoothManagerStub());
     }
+    if (useAppCheck) {
```
sharjeelyunus, line 212: I think we don't need this here, and this can be done with env variables, we should do this all this in `ensemble.dart` or `ensemble_app.dart`

starter/lib/generated/ensemble_modules.dart
```diff
@@ -86,10 +88,15 @@ class EnsembleModules {
...
+  // enable appcheck (if true)
+  static const useAppCheck = false;
+  static const useFirebaseFunctions = false;
+
+
   Future<void> init() async {
     // Note that notifications is not a module
 
-    if (useMoEngage || useNotifications || useFirebaseAnalytics || useAuth) {
+    if (useMoEngage || useNotifications || useFirebaseAnalytics || useAuth || useAppCheck || useFirebaseFunctions) {
```
sharjeelyunus, line 99: see the firebase api provider, we initialize it via env, shouldn't se do the same for firebase functions as well?

sharjeelyunus (approved):
Firebase AppCheck or Firebase Functions API Provider aren't the modules, we shouldn't need to change anything in starter for these, I believe we can do all this via env configs. Please try it out and let me know if you get any issues

PS: Also attach some videos/screenshots
Use cases
1. What happens when I only enable Firebase AppCheck
2. What happens if I only enable useFirebaseFunctions
3. What happens when I enable both

modules/ensemble/lib/framework/apiproviders/firebase_functions/firebase_functions_api_provider.dart
```diff
@@ -0,0 +1,204 @@
...
+import 'package:flutter/foundation.dart';
+import 'package:flutter/widgets.dart';
+import 'package:yaml/yaml.dart';
+
+import 'package:firebase_app_check/firebase_app_check.dart';
+
+class FirebaseFunctionsAPIProvider extends APIProvider {
+  // final FirebaseFunctions _functions = FirebaseFunctions.instance;
+  final FirebaseFunctions _functions = FirebaseFunctions.instanceFor(
+    region: 'us-central1', // Use your region
```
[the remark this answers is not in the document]
M-Taha-Dev: added dynamic support for region while calling firebase function

starter/lib/generated/ensemble_modules.dart
M-Taha-Dev: I've moved this part to ensemble.dart

modules/ensemble/lib/framework/apiproviders/firebase_functions/firebase_functions_api_provider.dart
```diff
@@ -0,0 +1,485 @@
...
+            messagingSenderId: webConfig['messagingSenderId'],
+            projectId: webConfig['projectId'],
+            authDomain: webConfig['authDomain'],
+            storageBucket: webConfig['storageBucket'],
+            measurementId: webConfig['measurementId'],
+          );
+        }
+      }
+
+      // Fallback to hardcoded values if no config found
```
sharjeelyunus, line 186: But why?? Shouldn't it throw error instead of adding hardcoded values of a random Firebase account?

modules/ensemble/lib/framework/apiproviders/firebase_functions/firebase_functions_api_provider.dart
```diff
@@ -0,0 +1,485 @@
...
+        return false;
+      }
+      
+      String? appCheckEnabled = appConfig.envVariables?['firebase_app_check'];
+      if (appCheckEnabled == null) {
+        log('firebase_app_check environment variable is not set, defaulting to false');
+        return false;
+      }
+      
+      // Check if appcheck is enabled (true/false or yes/no)
```
sharjeelyunus, line 89: This should be configured only by true/false. Why do we need yes/no, 0/1?

sharjeelyunus (commented):
Please clean up your PR

M-Taha-Dev:
> Firebase AppCheck or Firebase Functions API Provider aren't the modules, we shouldn't need to change anything in starter for these, I believe we can do all this via env configs. Please try it out and let me know if you get any issues
> 
> PS: Also attach some videos/screenshots Use cases
> 
> 1. What happens when I only enable Firebase AppCheck
> 2. What happens if I only enable useFirebaseFunctions
> 3. What happens when I enable both

If only appcheck is enabled ("firebase_app_check": "true" in env) -> this wont be of much use right now as the logic for closing the app or showing appropriate message on appcheck failure is still pending

If we only enable useFirebaseFunctions (now it is "cloud_function_provider": "firebase" in env file), user will be able to call firebase functions that don't enforce appcheck on them.

If we enable both, user will be able to call firebase functions that enforce appcheck as well.

Please find sample attached config:
image

TheNoumanDev (approved):
LGTM, @sharjeelyunus please give it a look as well.

Also please dont just log error msgs, about config not found, or invalid configuration, please propagate error msgs properly similarly we are doing in other files so that user in studio also get to know why firebase functions are not working. 
Please look into file at this path, see wot appropriate error we are getting and use that: `ensemble/lib/framework/error_handling.dart`

M-Taha-Dev:
hey @sharjeelyunus, please find the attached video containing demo for PR (firefunctions & appcheck) under different senarios

https://github.com/user-attachments/assets/3dc84ebe-271c-4f34-9a2a-9787d0875831

PR #2014, opened 2025-06-01
added pseudo apiSecureStorage keyword to get async keychain values

M-Taha-Dev:
Added logic to handle pseudo apiSecureStorage keyword to fetch async values from Keychain manager. The scope of this keyword is in API only.
Sample EDL:
image


vusters (approved):
this looks good.

sharjeelyunus (approved):

PR #2026, opened 2025-06-05
fix jwtToken evaluation 

Umair-Manzoor-47:
`token` wasn't being evaluated in `signInWithCustomToken` .
It is evaluated with dataContext and proper error handling is done.


TheNoumanDev (approved):
LGTM

PR #2039, opened 2025-06-18
added support for showMenu which allows user to show bottom menu outs…

M-Taha-Dev:
Hide/Show Bottom Navigation Bar
Added ability to hide bottom navigation on specific pages using showMenu in View styles.

Solution:
Store bottom navbar during creation, then conditionally render it based on showMenu keyword - show if true, hide if false.

Why this approach:
Initially tried shell Navigator pattern but it broke the onResume functionality when navigating back from screens that were added in the navigator, so switched to this simpler solution that preserves existing navigation behavior.

image


M-Taha-Dev:
@sharjeelyunus, Please find the attached Demo video

https://github.com/user-attachments/assets/cca9fae1-96a3-4d47-a4af-b93f07c6804f

PR #2041, opened 2025-06-24
added support for hiding header onListViewScroll

M-Taha-Dev:
Added support for hiding header when a user scrolls in listview. 

- added a new function buildScrollablePageContentWithCollapsableHeader that will be invoked only if we have set collapsableHeader property true in header styles

- when header collaps, body exceeds the defined SafeArea, to prevent this, I've added safe area for this scenario as well. we will need to set collapseSafeArea to true in header styles
- also added floating property to AnimationHeader

### DIFF
--- a/modules/ensemble/lib/framework/view/page.dart
+++ b/modules/ensemble/lib/framework/view/page.dart
@@ -93,7 +93,7 @@ class PageState extends State<Page>
 
   @override
   bool get wantKeepAlive => true;
-
+  
   @override
   void didUpdateWidget(covariant Page oldWidget) {
     super.didUpdateWidget(oldWidget);
@@ -196,6 +196,7 @@ class PageState extends State<Page>
   @override
   void didPopNext() {
     super.didPopNext();
+      currentPageKey = widget._pageModel.hashCode.toString();
     if (widget._pageModel.viewBehavior.onResume != null) {
       ScreenController().executeActionWithScope(
           context, _scopeManager, widget._pageModel.viewBehavior.onResume!,
@@ -253,6 +254,7 @@ class PageState extends State<Page>
         // has completely rendered. This will be sufficient for most use case
         widget.onRendered();
       });
+
     }
 
     // build the root widget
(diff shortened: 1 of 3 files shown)


M-Taha-Dev:
https://github.com/user-attachments/assets/da761316-9bb3-4bc2-a453-24712f233fd7

modules/ensemble/lib/framework/view/page.dart
```diff
@@ -544,12 +549,20 @@ class PageState extends State<Page>
 
             // appBar is inside CustomScrollView if defined
             appBar: fixedAppBar,
-            body: FooterLayout(
-              body: isScrollableView
-                  ? buildScrollablePageContent(hasDrawer)
-                  : buildFixedPageContent(fixedAppBar != null),
-              footer: footerWidget,
-            ),
+body: FooterLayout(
```
TheNoumanDev, line 552: Can you please format the code.

modules/ensemble/lib/framework/view/page.dart
```diff
@@ -631,6 +644,37 @@ class PageState extends State<Page>
...
+        bool isScrollableView = widget._pageModel.runtimeStyles?['scrollableView'] == true;
+if (isScrollableView) {
+  currentPageKey = widget._pageModel.hashCode.toString();
+  if (persistentControllers.containsKey(currentPageKey!)) {
+    externalScrollController = ScrollController();
+    persistentControllers[currentPageKey!] = externalScrollController!;
+  } else {
+    externalScrollController = ScrollController();
+    persistentControllers[currentPageKey!] = externalScrollController!;
+  }
```
TheNoumanDev, line 657: why this `if-and-else` is having the same thing happening?

modules/ensemble/lib/layout/helpers/list_view_core.dart
```diff
@@ -161,16 +162,24 @@ class _ListViewCoreState extends State<ListViewCore> {
...
         widget.shrinkWrap) {
-      final currentOffset = _scrollController.position.pixels;
+      _scrollController.addListener(() {
+        
+      final currentOffset = _scrollController.offset;
+      final scrollingUp = currentOffset > _previousOffset;
+      final scrollingDown = currentOffset < _previousOffset;
+
+      if (scrollingUp) {
+        currentExternalScrollController!.animateTo(100, duration: const Duration(milliseconds: 16), curve: Curves.linear);
```
TheNoumanDev, line 176: Please use the values from `animation` map and then have these as default values if not given.

PR #2055, opened 2025-07-15
feat: implement intelligent error handling

sharjeelyunus:
- Added a custom error handler to distinguish between fatal and non-fatal errors based on error type, source library, and context.
- Introduced methods to determine if errors should be treated as fatal for both Flutter and async errors.
- Updated error handling setup to ensure comprehensive coverage for both synchronous and asynchronous errors.

### DIFF
--- a/modules/firebase_analytics/README.md
+++ b/modules/firebase_analytics/README.md
@@ -1,16 +1,90 @@
-# firebase_analytics
+# Firebase Analytics Module for Ensemble
 
-A new Flutter project.
+This module provides a robust integration of Firebase Analytics and Crashlytics for Ensemble-based Flutter applications. It offers intelligent error handling, event logging, and user tracking, with support for platform-specific Firebase configurations.
+
+## Features
+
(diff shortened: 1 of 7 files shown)


TheNoumanDev:
@sharjeelyunus seems good, but there's a concern. Before this change, apps using Firebase Analytics would report every error as fatal. With these changes, we're now differentiating between fatal and non-fatal errors, which might break existing application release structures. 

Wouldn't it be better to control this smart behavior with a configuration property? That way, whoever enables it gets the smart error differentiation, while existing apps keep working as before (only applies if there are old customers using it).

Also, we're doing string matching which might break in the future if Crashlytics changes the keywords for errors - then non-fatal errors might be recorded as fatal or vice versa.

sharjeelyunus:
@TheNoumanDev 
1. lil confused on how it will break the existing apps structure? Before that we were reporting all errors as fatal, and now we are categorising them!!
2. We aren't string matching on un common, unheard firebase errors, we are string matching on flutter's most common standard errors

TheNoumanDev (approved):

PR #2103, opened 2025-10-02
added always on display action and uprgraded gradle version

TheNoumanDev:
Ticket: https://pub.dev/packages/wakelock_plus

Example EDL of action:
```yaml
wakelock:
  enable: true # could be true and false
  onComplete: |
    console.log('Wakelock enabled - screen will stay awake');
  onError: |
    console.log('Failed to enable wakelock');
```
And we can get the current status from using `ensemble.device.wakelockEnabled`

Its latest version required `package_info_plus ` to be upgraded, and `package_info_plus ` required the latest Kotlin and gradle version.


sharjeelyunus (approved):
`enable` should not be required, it should be true by default

PS: remove unnecessary print statements

PR #2185, opened 2026-04-06
feat(image): add support for custom HTTP headers in image requests

### LINKED ISSUE #2184: Support Custom HTTP Headers for Network Image Requests
Currently, when loading images from a network source, only a URL can be provided. However, many image endpoints require custom HTTP headers (e.g., `Authorization`, `Cookie`, or API keys) to successfully fetch the resource.

This limitation prevents loading images from secured or authenticated sources.

#### **Problem**

* Only URL-based image loading is supported.
* No way to attach headers to the request.
* Fails for protected resources (e.g., private CDN, auth-based image services).

#### **Proposed Solution**

Introduce support for passing custom HTTP headers in network image requests.

Example:

```yaml
Image:
  source: https://example.com/image.png
  headers:
    Authorization: Bearer <token>
    X-Custom-Header: value
```

#### **Use Cases**

* Loading images from private CDN (e.g., Cloudflare R2 with auth)
* Signed or session-based image endpoints
* Enterprise/internal image services

usmanvrtx:
### Summary
Add support for custom HTTP headers in network images.

### Key Changes

**Image Request Handling**  
- Enable passing custom HTTP headers (e.g., `Authorization`) with network image requests.  
- Extend existing image loading logic to include header configuration.  
- Ensure headers are correctly attached to all outgoing image fetch requests.  

**API Consistency**  
- Align network image behavior with standard HTTP request patterns.  
- Avoid ad-hoc workarounds for authenticated or protected image resources.  

**Code Improvements**  
- Refactor request pipeline to support header injection cleanly.  
- Maintain backward compatibility for existing image usage.  

**Video Demonstration**  
[Watch video](https://github.com/user-attachments/assets/efc8e950-866f-4000-b4e5-a6a2f4643871)

### Usage Example

```yaml
- Image:
    source: https://example.com/image.png
    headers:
      Authorization: "Bearer ${token}"
      X-Custom-Header: "my-custom-value"

### DIFF
--- a/modules/ensemble/lib/widget/image.dart
+++ b/modules/ensemble/lib/widget/image.dart
@@ -1,8 +1,9 @@
-import 'dart:io';
+import 'dart:io' as io;
 
 import 'package:cached_network_image/cached_network_image.dart';
 import 'package:ensemble/action/haptic_action.dart';
 import 'package:ensemble/framework/action.dart';
+import 'package:ensemble/framework/data_context.dart';
 import 'package:ensemble/framework/event.dart';
 import 'package:ensemble/framework/widget/colored_box_placeholder.dart';
 import 'package:ensemble/framework/widget/widget.dart';
@@ -74,10 +75,8 @@ class EnsembleImage extends StatefulWidget
       'pinchToZoom': (value) =>
           _controller.pinchToZoom = Utils.optionalBool(value),
       'colorFilter': (value) => _controller.colorFilter = ColorFilterComposite.from(value),
-      'headers': (value) =>
-          _controller.headers = Utils.getMap(value)
+      'headers': (value) => _controller.headers = Utils.getMap(value)
           ?.map((key, value) => MapEntry(key.toString(), value.toString())),
-
     };
   }
 }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: modules/ensemble/lib/widget/image.dart
Comment (TheNoumanDev) [inline modules/ensemble/lib/widget/image.dart:289]: Headers are passed to `CachedNetworkImage` and `SvgPicture.network`, but not here for web, can you verify if we can add it here as well?

File: modules/ensemble/lib/widget/image.dart [hunk 1]
```diff
@@ -74,6 +74,8 @@ class EnsembleImage extends StatefulWidget
       'pinchToZoom': (value) =>
           _controller.pinchToZoom = Utils.optionalBool(value),
       'colorFilter': (value) => _controller.colorFilter = ColorFilterComposite.from(value),
+      'headers': (value) =>
+          _controller.headers = Utils.getMap(value)?.cast<String, String>(),
```
TheNoumanDev: This will throw a runtime error if any header value isn't a String (e.g., timeout: 30). The rest of the codebase uses explicit.toString() conversion, can you please verify this as well?

TheNoumanDev (changes requested):

modules/ensemble/lib/widget/image.dart
usmanvrtx: We check `isMemoryPath` here, so this is only for local/memory sources (e.g., blob URLs on web), not real network requests. Since no HTTP call is made, headers don’t apply.

modules/ensemble/lib/widget/image.dart
usmanvrtx: updated this to this to use explicit.toString() for header to prevent type errors.

TheNoumanDev (approved):
LGTM

PR #2186, opened 2026-04-07
feat(tab): add persistentTabBar option for keep the tab pinned

### LINKED ISSUE #2158: Support sticky/pinned TabBar with scrollable tab content
Currently, when a screen has a TabBar with scrollable content inside each tab, the entire view scrolls — including the TabBar itself, which disappears off-screen. We need the TabBar to remain pinned/sticky at the top while only the tab content scrolls beneath it.

usmanvrtx:
### Summary
Add `persistentTabBar` support so the tab bar stays pinned while only the tab content scrolls.

### Key Changes

**TabBar Behavior**  
- Add `persistentTabBar` parameter to `TabBar`.  
- Keep the tab bar fixed at the top.  
- Scroll only the selected tab content.

**Usage**  
```yaml
  body:
    TabBar:
      styles:
        persistentTabBar: true

### DIFF
--- a/modules/ensemble/lib/layout/tab_bar.dart
+++ b/modules/ensemble/lib/layout/tab_bar.dart
@@ -90,6 +90,8 @@ abstract class BaseTabBar extends StatefulWidget
           EnsembleAction.from(action, initiator: this),
       'onTabSelectionHaptic': (value) =>
           _controller.onTabSelectionHaptic = Utils.optionalString(value),
+      'persistentTabBar': (value) =>
+          _controller.persistentTabBar = Utils.getBool(value, fallback: false),
     };
   }
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: modules/ensemble/lib/layout/tab_bar.dart [hunk 1]
```diff
@@ -282,7 +284,12 @@ class TabBarState extends BaseTabBarState {
           // builder gives us dynamic height control vs TabBarView, but
           // is sub-optimal since it recreates the tab content on each pass.
           // This means onLoad API may be called multiple times in debug mode
-          tabContent
+
+          widget._controller.persistentTabBar
+              ? Expanded(
```
TheNoumanDev: We are wrapping it with `Expanded` widget based on `persistentTabBar`, and also wrapping it with `Expanded` when `isExpanded` is also true, so if we set both properties, it gets wrapped twice in `Expanded` widget. Can you check if using both properties breaks it or works fine?

Verify with some edge cases:
- persistentTabBar: true + content has ListView
- persistentTabBar: true + content has Expanded child
- Tab content shorter than the screen, verify that it takes space as widget space or it takes extra space.

usmanvrtx: **Property Behavior:**

**`expanded` property:**
- `expanded: true` → Content stretches to fill remaining screen space, even if content is shorter than available height
- `expanded: false` → Content only takes the space it needs

**`persistentTabBar` property:**
- `persistentTabBar: true` → Enables scrolling while keeping tab bar visible; works with or without the `expanded` property
- `persistentTabBar: false` → Default behavior (TabBarView)

TheNoumanDev (approved):
LGTM

PR #2190, opened 2026-04-27
Add `useIndexedTab` property for smooth tab switching with lazy-loaded indexed building

usmanvrtx:
**Description:**  
Introduces `useIndexedTab`, a boolean property that enables IndexedStack-based tab building in TabBar widgets. 

**Problem it solves:**  
By default, when switching tabs, the previous tab's widget is removed from the widget tree. Switching back to that tab requires rebuilding it, which causes lag or freezing if the tab contains heavy data.

**Solution:**  
With `useIndexedTab: true`, tabs are built on-demand and cached. When you switch to a tab for the first time, it's built and cached. Tab switching becomes instant—no rebuild, no delay, no freeze. Unvisited tabs remain lazy and are not built until accessed.

**How it works:**  
- Tabs are built only when first accessed (lazy loading)
- Built tabs are cached and kept in the widget tree while the screen is active
- Uses `IndexedStack` (expanded) or `Column` with `Offstage` (non-expanded) to maintain widgets in tree
- Only the active tab is visible; others remain cached and ready

**Impact:**  
Eliminates freezed transitions when switching between tabs with heavy content.

https://github.com/user-attachments/assets/d20a713e-76bd-4498-b043-ebdbcbbe97db

### DIFF
--- a/modules/ensemble/lib/layout/tab/tab_bar_controller.dart
+++ b/modules/ensemble/lib/layout/tab/tab_bar_controller.dart
@@ -66,8 +66,6 @@ class TabBarController extends BoxController {
     var setters = super.getBaseSetters();
     setters.addAll({
       'items': (values) => items = values,
-      'useIndexedTab': (value) =>
-      useIndexedTab = Utils.getBool(value, fallback: false),
     });
     return setters;
   }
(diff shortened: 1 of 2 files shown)


TheNoumanDev (approved):
LGTM

TheNoumanDev (approved):
LGTM

PR #2289, opened 2026-06-17
Ignore keyboard-only device metric changes

usmanvrtx:
## Description

Prevents keyboard open/close events from triggering unnecessary device metric updates. The page now tracks the last dispatched screen width, height, and orientation, and only notifies bindings when those values actually change.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Documentation update

## What Has Changed

- Added caching for the last dispatched device metrics in `PageState`.
- Skipped device metric notifications when only keyboard-related metrics change.
- Captured the initial device metrics after the first frame so later updates can be compared correctly.

## How to Test

1. Open a page that listens to `${device.width}` or `${device.orientation}`.
2. Focus and unfocus a text input to show and hide the keyboard.
3. Confirm the bindings do not re-fire unless the screen size or orientation actually changes.

### DIFF
--- a/modules/ensemble/lib/framework/view/page.dart
+++ b/modules/ensemble/lib/framework/view/page.dart
@@ -104,6 +104,12 @@ class PageState extends State<Page>
   Timer? _titleBarHeightPollTimer;
   Timer? _headerVisibilityPollTimer;
 
+  // Last dispatched device metrics. We only want to notify bindings when the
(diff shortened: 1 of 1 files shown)


sharjeelyunus (approved):
LGTM

PR #2345, opened 2026-08-21
fix: use caret constraint for ensemble_test_runner's ensemble dependency

TheNoumanDev:
## Summary
`melos version` corrupts hosted dependency constraints written as a compound range (e.g. `ensemble_test_runner`'s `">=1.2.50 <2.0.0"`) when bumping `ensemble` — it only replaces the leading token, leaving invalid syntax behind. Broke the last beta release; would've broken the next stable release too. Bug confirmed in melos itself.

Fix repairs melos's output instead of changing the dependency declaration:
- `.github/scripts/fix_melos_version_rewrite.dart` — detects and repairs the corruption using real `pub_semver` parsing, preserving the original range style and upper bound; fails loudly on anything unrecognized instead of guessing
- `.github/scripts/repair_and_retag.sh` — shared by both release workflows: runs the repair after `melos version`, amends it into melos's commit, re-points every tag melos created
- `release-beta-version.yml` / `release-melos-version.yml` — pull release scripts and `publish-ensemble-pubdev.yml` from the workflow's own ref rather than the branch/ref being released (which may lack or predate them); stable pipeline additionally guards against overlaying an older ref's files onto main
- `publish-ensemble-pubdev.yml` — added opt-in beta publishing. pub.dev only trusts publishes triggered by a genuine tag push (confirmed — `workflow_call`/`workflow_dispatch` origins are rejected outright), so the opt-in travels as an annotated tag message (`[publish-to-pubdev]`) rather than a workflow input; beta tags without the marker are skipped

## Test plan
- [x] Reproduced and fixed the corruption for both beta and stable bumps
- [x] Repair script: fixes real corruption, preserves range style, no-op when healthy, fails loudly on unrecognized shapes, idempotent
- [x] Stable scenario: all 21 tags melos creates survive the amend/re-tag step; stale-ref overlay guard verified in both directions
- [x] `actionlint`, `shellcheck`, `dart analyze` clean
- [x] Confirmed live: a real beta release (`ensemble 1.2.50-beta.11`) ran the full pipeline and published successfully to pub.dev as a prerelease

### DIFF
--- a/tools/ensemble_test_runner/pubspec.yaml
+++ b/tools/ensemble_test_runner/pubspec.yaml
@@ -19,7 +19,7 @@ environment:
   flutter: ">=3.24.0"
 
 dependencies:
-  ensemble: ">=1.2.50 <2.0.0"
+  ensemble: "^1.2.50"
   device_frame: ^1.3.0
   ensemble_device_preview: ^1.1.3
   flutter:
--- a/tools/ensemble_test_runner/pubspec.yaml
+++ b/tools/ensemble_test_runner/pubspec.yaml
@@ -19,7 +19,7 @@ environment:
   flutter: ">=3.24.0"
 
 dependencies:
-  ensemble: "^1.2.50"
+  ensemble: ">=1.2.50 <2.0.0"
   device_frame: ^1.3.0
   ensemble_device_preview: ^1.1.3
   flutter:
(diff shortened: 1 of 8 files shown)


tools/ensemble_test_runner/pubspec.yaml
```diff
@@ -19,7 +19,7 @@ environment:
   flutter: ">=3.24.0"
 
 dependencies:
-  ensemble: ">=1.2.50 <2.0.0"
+  ensemble: "^1.2.50"
```
sharjeelyunus: this is not right, we don't wanna leave the ensemble versions open ended, this can cause issues

see, how we are handling the real issue in `release-melos-version.yml`

we are using `--scope` and `--manual-version` from melos here

I think the better way would be to have a an input `Beta: true/false" in release-melos-version.yml instead of managing 2 different pipelines

sharjeelyunus (approved):

PR #2362, opened 2026-09-10
Fix/id focus state

usmanvrtx:
## Description

This PR fixes `${id.hasFocus}` resetting to `false` when a focused TV widget is rebuilt after navigating back.

The previous widget controller is available through the widget ID context. The replacement controller now inherits its `hasFocus` value before bindings are evaluated, preventing ID-bound focus styles from rendering unfocused during the rebuild.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Documentation update

### Preserve ID-bound focus state during rebuilds

- Look up the previous widget instance using its YAML `id`.
- When the previous controller has focus, copy `hasFocus` to the replacement controller before bindings run.
- Keep the existing TV focus node, navigation, scrolling, actions, and binding-dispatch behavior unchanged.

### Behavior Compatibility

- No changes to D-pad traversal, focus coordinates, navigation, actions, or scrolling.
- No expected mobile behavior change: `hasFocus` is set by the existing TV focus wrappers.
- The change applies only when an ID-bound widget is rebuilt and the previous controller is focused.

## How to Test

1. From `modules/ensemble`, run:

   ```bash
   flutter analyze lib/framework/widget/view_util.dart

### DIFF
--- a/modules/ensemble/lib/framework/widget/view_util.dart
+++ b/modules/ensemble/lib/framework/widget/view_util.dart
@@ -379,7 +379,12 @@ class ViewUtil {
     }
 
     // Preserve ID-bound TV focus styles while a focused widget is rebuilt.
-    if (previousContext is HasController && w is HasController) {
+    // Only carry focus from a widget registered in THIS scope — an ancestor
+    // scope hit with the same id is a collision, not the widget being rebuilt.
+    if (widgetId != null &&
+        scopeNode.scope.dataContext.contextMap.containsKey(widgetId) &&
+        previousContext is HasController &&
+        w is HasController) {
       final previousController = previousContext.controller;
       final nextController = w.controller;
       if (previousController is WidgetController &&
--- a/modules/ensemble/lib/framework/widget/view_util.dart
+++ b/modules/ensemble/lib/framework/widget/view_util.dart
@@ -348,6 +348,21 @@ class ViewUtil {
         child: customWidget);
   }
 
+  /// Returns the [WidgetController] behind [value] (a controller or a widget
+  /// that owns one via [HasController]).
+  static WidgetController? _focusControllerOf(dynamic value) {
+    if (value is WidgetController) {
+      return value;
+    }
+    if (value is HasController) {
+      final controller = value.controller;
+      if (controller is WidgetController) {
+        return controller;
+      }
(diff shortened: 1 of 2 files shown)


TheNoumanDev:
Code review(medium · 2 findings)
modules/ensemble/lib/framework/widget/view_util.dart
  ● 388 [correctness] The rebuild copy writes hasFocus = true onto the new controller without verifying the rebuilt widget's focus node is actually focused, so when the rebuild produces a new element instead of reusing the old one, ${id.hasFocus} is stuck true permanently.
  ● 382 [correctness] getContextById walks the parent-scope chain, and the new `previousContext is HasController` branch acts on outer-scope hits that the previous EnsembleController-only check ignored for old-style widgets.

Scope: the branch is now fix/id-focus-state (PR #2362, base android_TV_implementation) — 17 added / 6 removed lines in view_util.dart only. Note: 23c57fbf added resync logic in box_wrapper.dart, and a2b0dcd9 reverted all of it, so box_wrapper is net-unchanged in the PR. Analyzes clean apart from the pre-existing unused_local_variable at line 434.

Two findings, both on the new block at view_util.dart:382-390. The core issue is that the copy is one-directional — it only ever writes true, and never re-checks the rebuilt node's actual state. That's correct for the case the comment describes (Flutter reusing the element, FocusNode persists), but wrong when the rebuild produces a new element, and the initState/didUpdateWidget resync that covered exactly that case was removed in the second commit with nothing put in its place. The blast radius is limited to ${id.hasFocus} bindings since the focus ring reads live focus state — but that binding is the point of the PR.

The second is narrower: hoisting the lookup and widening the type check to HasController means getContextById's parent-scope walk can now match an outer-scope widget sharing the same id, which the old EnsembleController-only check silently ignored for legacy widgets.

The rest of the diff is clean — the previousContext EnsembleController refactor is behavior-preserving, hoisting getContextById out of the widgetInstance != null branch has no side effects (pure map walk), CustomWidgetModel still returns early before the new code, and running the copy before _updateWidgetBindings means skipping the ModelChangeEvent dispatch is right.

TheNoumanDev (approved):

PR #2366, opened 2026-09-14
feat(ensemble_agent): initial release of on-device AI agent runtime with provider-agnostic API and native plugin support

sharjeelyunus:


sharjeelyunus:
NOT YET READY

